PR #9, opened 2024-09-18
Stream

vicb:


builder/src/templates/worker.ts
```diff
@@ -23,6 +21,7 @@ export default {
         customServer: false,
         dev: false,
         dir: "",
+        minimalMode: false,
```
petebacondarwin, line 24: why this?

builder/src/templates/worker.ts
```diff
@@ -32,50 +31,97 @@ export default {
...
+
+  // Only allow statusCode to be modified if not sent
+  let { statusCode } = res;
+  Object.defineProperty(res, "statusCode", {
+    get: function () {
+      return statusCode;
+    },
+    set: function (val) {
+      if (this.finished || this.headersSent) {
+        return;
```
petebacondarwin, line 105: Not a throw?

petebacondarwin (approved):

builder/src/templates/worker.ts
vicb: This is to make the default explicit

builder/src/templates/worker.ts
vicb: Adding a console error for now

builder/src/build/build-worker.ts
```diff
@@ -100,12 +100,18 @@ globalThis.fetch = (input, init) => {
...
+import { Readable } from 'node:stream';
 globalThis.fetch.__nextPatched = isPatchedAlready;
 fetch = globalThis.fetch;
 const CustomRequest = class extends globalThis.Request {
   constructor(input, init) {
     console.log("CustomRequest", input);
-    if (init) delete init.cache;
+    if (init) {
+      delete init.cache;
+      if (init.body?.__node_stream__ === true) {
```
mhart, line 111: I think it should be fine to do without the custom property and just check `instance of Readable`:

```
    if (init) {
      init = {
        ...init,
        cache: undefined,
        body: init.body instanceof Readable ? Readable.toWeb(init.body) : init.body
      }
    }
```

(at least, these works in my tests so far)

PR #12, opened 2024-09-20
implement webpack chunks file updating using ast manipulation

dario-piotrowicz:
this allows us not to have to require the use of the `serverMinification: false` option
(as we're no longer relying on known variable names)


examples/api/next.config.mjs
```diff
@@ -1,11 +1,6 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   output: "standalone",
-  experimental: {
-    // IMPORTANT: this option is necessary for the chunks hack since that relies on the webpack-runtime.js file not being minified
-    //            (since we regex-replace relying on specific variable names)
-    serverMinification: false,
-  },
```
dario-piotrowicz, line 8: 💪

vicb: Woohoo, that's so cool!

vicb:
That's amazing @dario-piotrowicz 

Could you please update `TODO.md` and `builder/README.md`? (which reminds me that I should remove the URL pkg from there too).

I need to go for 1h, I'll review after that.

Thanks!

dario-piotrowicz:
> That's amazing @dario-piotrowicz
> 
> Could you please update `TODO.md` and `builder/README.md`? (which reminds me that I should remove the URL pkg from there too).
> 
> I need to go for 1h, I'll review after that.
> 
> Thanks!

Thanks ❤️ 

`TODO.md` and`builder/README.md` updated 🙂 

I've removed the section regarding updating the config file entirely as I think that that is now totally unnecessary and that we should not bother telling users to set the `standalone` output there, I hope that's fine 🙂

builder/package.json
```diff
@@ -32,6 +34,10 @@
     "glob": "^11.0.0",
     "next": "14.2.5",
     "tsup": "^8.2.4",
-    "typescript": "^5.5.4"
+    "typescript": "^5.5.4",
+    "vitest": "^2.1.1"
+  },
+  "dependencies": {
+    "ts-morph": "^23.0.0"
```
petebacondarwin, line 41: dep or devDep?

builder/src/build/patches/investigated/update-webpack-chunks-file/test-fixtures/minified-webpacks-file.js
petebacondarwin, line 1: Could we generate this file on the fly by minifying the unminified version?
That way we have one less file to keep in sync...

examples/create-next-app/next.config.mjs
```diff
@@ -1,9 +1,6 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   output: "standalone",
```
petebacondarwin, line 3: Remove the standalone too?

petebacondarwin (commented):
Initial notes on all but the actual transform files 😄

builder/src/build/patches/investigated/update-webpack-chunks-file/index.ts
```diff
@@ -0,0 +1,32 @@
+import { readdirSync, readFileSync, writeFileSync } from "node:fs";
+
+import { NextjsAppPaths } from "../../../../nextjs-paths";
+import { getUpdatedWebpackChunksFileContent } from "./get-updated-webpack-chunks-file-content";
+
+/**
+ * Fixes the webpack-runtime.js file by removing its webpack dynamic requires.
+ *
+ * This hack is especially bad for two reasons:
+ *  - it requires setting `experimental.serverMinification` to `false` in the app's config file
```
petebacondarwin, line 10: Is this still true?

builder/src/build/patches/investigated/update-webpack-chunks-file/get-chunk-installation-identifiers.ts
```diff
@@ -0,0 +1,77 @@
...
+function getInstalledChunksDeclaration(
+  sourceFile: ts.SourceFile,
+  installChunkDeclaration: ts.VariableDeclaration
+) {
+  const allVariableDeclarations = sourceFile.getDescendantsOfKind(ts.SyntaxKind.VariableDeclaration);
+  const installChunkDeclarationIdx = allVariableDeclarations.findIndex(
+    (declaration) => declaration === installChunkDeclaration
+  );
+
+  // the installedChunks declaration is comes right before the installChunk one
```
petebacondarwin, line 70:
```suggestion
  // the installedChunks declaration comes right before the installChunk one
```

builder/src/build/patches/investigated/update-webpack-chunks-file/get-chunk-installation-identifiers.ts
```diff
@@ -0,0 +1,77 @@
+import * as ts from "ts-morph";
+
+export async function getChunkInstallationIdentifiers(sourceFile: ts.SourceFile): Promise<{
+  installedChunks: string;
+  installChunk: string;
+}> {
+  const installChunkDeclaration = getInstallChunkDeclaration(sourceFile);
```
petebacondarwin, line 7: Is it possible that there might be some false positives for "install chunk" declarations?
If so, could we cover ourselves a bit tighter by returning all of them here and then filtering them by the ones that have a matching "installed chunks" declaration directly in front?
And then if there is still more than one either erroring or at least warning?

builder/src/build/patches/investigated/update-webpack-chunks-file/get-updated-webpack-chunks-file-content.ts
```diff
@@ -0,0 +1,22 @@
+import * as ts from "ts-morph";
+
+import { getChunkInstallationIdentifiers } from "./get-chunk-installation-identifiers";
+import { getFileContentWithUpdatedWebpackFRequireCode } from "./get-file-content-with-updated-webpack-require-f-code";
```
petebacondarwin, line 4: `FRequire` and `require-f` - did you mean for these to be in different orders? Do they mean different things?

builder/src/build/patches/investigated/update-webpack-chunks-file/get-file-content-with-updated-webpack-require-f-code.ts
```diff
@@ -0,0 +1,74 @@
+import * as ts from "ts-morph";
+
+export async function getFileContentWithUpdatedWebpackFRequireCode(
```
petebacondarwin, line 3: Could you add some comments explaining how this all works?

builder/src/build/patches/investigated/update-webpack-chunks-file/get-file-content-with-updated-webpack-require-f-code.ts
```diff
@@ -0,0 +1,74 @@
...
+
+      const binaryExpressionOperator = binaryExpression.getOperatorToken();
+      if (binaryExpressionOperator.getText() !== "=") return;
+
+      const binaryExpressionRight = binaryExpression.getRight();
+      const binaryExpressionRightText = binaryExpressionRight.getText();
+
+      if (
+        !binaryExpressionRightText.includes(installChunk) ||
+        !binaryExpressionRightText.includes(installedChunks)
```
petebacondarwin, line 22: So `binaryExpressionRightText` is the text of the whole fat arrow function that makes use of the `installChunk` and `installedChunks` identifiers? So we need both to exist in the text of the function.

builder/src/build/patches/investigated/update-webpack-chunks-file/get-file-content-with-updated-webpack-require-f-code.ts
```diff
@@ -0,0 +1,74 @@
...
+          return (
+            elementAccess.getExpression().getText() === installedChunks &&
+            elementAccess.getArgumentExpression()?.getText() === functionFirstParameterName
+          );
+        });
+      if (!accessesInstalledChunksUsingItsFirstParameter) return;
+
+      return arrowFunction;
+    })
+    .find(Boolean);
```
petebacondarwin, line 53: Just for my own clarity this is effectively returning the first item in the array that is not falsy?

petebacondarwin (approved):
Fantastic work @dario-piotrowicz

builder/src/build/patches/investigated/update-webpack-chunks-file/index.ts
dario-piotrowicz: nope! 😅 👍

builder/src/build/patches/investigated/update-webpack-chunks-file/get-chunk-installation-identifiers.ts
dario-piotrowicz: yes it's definitely possible to have false positives, but I think that it's very unlikely since we do check for some pretty specific things like the fact that the function has a single parameter and that it is used to access three of its properties in a specific order (`modules`, `ids` and `runtime`).

I feel like it's going to be pretty difficult to match those in non `installChunk` functions.

I can totally add the check you mentioned if you want but I feel that the extra complexity it'd introduce (even if not huge) would outweight the benefit it'd provide (personally I'd go with this and update the various checks if and when there's a need)

builder/package.json
```diff
@@ -32,6 +34,10 @@
     "glob": "^11.0.0",
     "next": "14.2.5",
     "tsup": "^8.2.4",
-    "typescript": "^5.5.4"
+    "typescript": "^5.5.4",
+    "vitest": "^2.1.1"
```
vicb, line 38: Could you please use `catalog:` as the version for the packages in https://github.com/flarelabs-net/poc-next/blob/main/pnpm-workspace.yaml#L5

builder/src/build/patches/investigated/update-webpack-chunks-file/get-chunk-installation-identifiers.test.ts
```diff
@@ -0,0 +1,30 @@
+import { readFile } from "node:fs/promises";
+
+import { expect, test, describe } from "vitest";
+
+import { getChunkInstallationIdentifiers } from "./get-chunk-installation-identifiers";
+import { getWebpackChunksFileTsSource } from "./get-webpack-chunks-file-ts-source";
+
+describe("getChunkInstallationIdentifiers", () => {
+  test("works as expected on unminified code", async () => {
```
vicb, line 9: nit: IMO "works as expected on " can be dropped - "get chunk identifiers on unminified code" would be more informative

builder/src/build/patches/investigated/update-webpack-chunks-file/get-chunk-installation-identifiers.ts
```diff
@@ -0,0 +1,77 @@
+import * as ts from "ts-morph";
+
+export async function getChunkInstallationIdentifiers(sourceFile: ts.SourceFile): Promise<{
```
vicb, line 3: nit: add  comments on exported functions

builder/src/build/patches/investigated/update-webpack-chunks-file/get-chunk-installation-identifiers.ts
```diff
@@ -0,0 +1,77 @@
...
+  const installChunkDeclaration = getInstallChunkDeclaration(sourceFile);
+  const installedChunksDeclaration = getInstalledChunksDeclaration(sourceFile, installChunkDeclaration);
+
+  return {
+    installChunk: installChunkDeclaration.getName(),
+    installedChunks: installedChunksDeclaration.getName(),
+  };
+}
+
+function getInstallChunkDeclaration(sourceFile: ts.SourceFile) {
```
vicb, line 16: nits: 
- add return type
- Add a snippet of the code we're looking for in a function level comment

builder/src/build/patches/investigated/update-webpack-chunks-file/get-chunk-installation-identifiers.ts
```diff
@@ -0,0 +1,77 @@
...
+    });
+
+  if (!installChunkDeclaration) {
+    throw new Error("ERROR: unable to find the installChunk function declaration");
+  }
+
+  return installChunkDeclaration;
+}
+
+function getInstalledChunksDeclaration(
```
vicb, line 61: nits: same as previous fn

builder/src/build/patches/investigated/update-webpack-chunks-file/index.ts
```diff
@@ -0,0 +1,32 @@
+import { readdirSync, readFileSync, writeFileSync } from "node:fs";
```
vicb, line 1: nit: what about moving that file as `builder/src/build/patches/investigated/update-webpack-chunks-file.ts`.

I find it easier to navigate code that way than having multiple `index.ts`

builder/src/build/patches/investigated/update-webpack-chunks-file/index.ts
```diff
@@ -0,0 +1,32 @@
...
+ * This hack is especially bad for two reasons:
+ *  - it requires setting `experimental.serverMinification` to `false` in the app's config file
+ *  - indicates that files inside the output directory still get a hold of files from the outside: `${nextjsAppPaths.standaloneAppServerDir}/webpack-runtime.js`
+ *    so this shows that not everything that's needed to deploy the application is in the output directory...
+ */
+export async function updateWebpackChunksFile(nextjsAppPaths: NextjsAppPaths) {
+  console.log("# updateWebpackChunksFile");
+  const webpackRuntimeFile = `${nextjsAppPaths.standaloneAppServerDir}/webpack-runtime.js`;
+
+  console.log({ webpackRuntimeFile });
```
vicb, line 18: nit: remove?

builder/src/build/patches/investigated/update-webpack-chunks-file/test-fixtures/minified-webpacks-file.js
vicb: +1 
But maybe create an issue and address in a follow up PR - I don't think we should block on this

vicb (approved):
That's awesome.

I added a few minor comments.

builder/package.json
dario-piotrowicz: It does need to be a dependency 🙂 

If we try bundling it in we get this error: https://github.com/flarelabs-net/poc-next/actions/runs/10959287219/job/30431309402

builder/src/build/patches/investigated/update-webpack-chunks-file/get-file-content-with-updated-webpack-require-f-code.ts
dario-piotrowicz: yes exactly 🙂

builder/src/build/patches/investigated/update-webpack-chunks-file/get-file-content-with-updated-webpack-require-f-code.ts
dario-piotrowicz: Yes, if this code looks unclear I can try to improve it

builder/src/build/patches/investigated/update-webpack-chunks-file/get-updated-webpack-chunks-file-content.ts
dario-piotrowicz: yeah.... I switched the two at some point, and I must have missed to update the file name, thanks for noticing it! 🙏

builder/src/build/patches/investigated/update-webpack-chunks-file/index.ts
dario-piotrowicz: this was already there... so I thought you wanted it 🤷 

I can definitely remove it if it's not wanted 🙂

builder/src/build/patches/investigated/update-webpack-chunks-file/test-fixtures/minified-webpacks-file.js
dario-piotrowicz: I'd kinda prefer to have it more layed out and clear, especially since we have it's own counterpart in `test-snapshots` too, so having only one and not the other feels a bit untidy to me (unless you'd also like me to get rid of the snapshot?)

Overall I guess I personally would keep it as is as I don't see the extra file to be a huge deal, but I am definitely happy to remove it and do the minification on the fly if that's what everyone prefers 🙂

dario-piotrowicz: I've created an issue for this: https://github.com/flarelabs-net/poc-next/issues/13 👍

builder/src/build/patches/investigated/update-webpack-chunks-file/index.ts
dario-piotrowicz: I've update the comment, the first point still stands unfortunatenly 🥲 

I've created a [new issue](https://github.com/flarelabs-net/poc-next/issues/14) to track the second point (I had this issue in my original repo)

builder/src/build/patches/investigated/update-webpack-chunks-file/get-file-content-with-updated-webpack-require-f-code.ts
dario-piotrowicz: I've introduced a new variable: `functionUsesChunkInstallationVariables` to hopefully make the code clearer 🙂

builder/src/build/patches/investigated/update-webpack-chunks-file/get-chunk-installation-identifiers.test.ts
dario-piotrowicz: updated 🙂👍 

Please have another look 🙏

builder/src/build/patches/investigated/update-webpack-chunks-file/index.ts
dario-piotrowicz: Sorry I must disagree on this one 😢 

I really prefer barrel files as those help containing related logic and define a restricted "public api" to be used externally 😢 

I am not sure what you're suggesting with:
`builder/src/build/patches/investigated/update-webpack-chunks-file.ts`
what about the rest of the files in the `builder/src/build/patches/investigated/update-webpack-chunks-file` directory? would you want me to move them in `builder/src/build/patches/investigated` or have them still in the same directory but have `builder/src/build/patches/investigated/update-webpack-chunks-file.ts` import them?

(both solution seem quite less clean the the current one to me 😢)

builder/src/build/patches/investigated/update-webpack-chunks-file/index.ts
dario-piotrowicz: removed 🙂👍

builder/package.json
dario-piotrowicz: done! 🙂👍

builder/src/build/patches/investigated/update-webpack-chunks-file/get-file-content-with-updated-webpack-require-f-code.ts
dario-piotrowicz: comment added 🙂

builder/src/build/patches/investigated/update-webpack-chunks-file/get-chunk-installation-identifiers.ts
dario-piotrowicz: done 🙂

builder/src/build/patches/investigated/update-webpack-chunks-file/get-chunk-installation-identifiers.ts
dario-piotrowicz: done 🙂

examples/create-next-app/next.config.mjs
dario-piotrowicz: done 🙂

builder/src/build/patches/investigated/update-webpack-chunks-file/get-file-content-with-updated-webpack-require-f-code.ts
dario-piotrowicz: I've refactored the code to remove the `find(Boolean)` 🙂👍

builder/src/build/patches/investigated/update-webpack-chunks-file/index.ts
IgorMinar: vscode now makes it easy to disambiguate many index.ts files which differ only in parent directory. 

I'm fine with keeping the index.ts and slightly prefer it, but I don't want to be too prescriptive. Since Vic doesn't object. I suggest we side with Dario on this one.

vicb: What I meant is that the following is confusing to me

image

But I can live with it

PR #15, opened 2024-09-20
ci: add prerelease deployments to npm

petebacondarwin:


vicb (approved):
🎉

builder/README.md
```diff
@@ -1,13 +1,12 @@
...
 
-## Build your app
+## Configure your app
 
 - update the `next.config.mjs` as follows
 
   ```typescript
   const nextConfig = {
-    output: "standalone",
```
IgorMinar, line 10: is this change right?

IgorMinar (approved):
yay! thanks! left one q

builder/README.md
vicb: Yep, we use an env flag while building next, no need for that.
(The build output stays the same - the only exception is the addition of the `standalone` folder containing the traced files + a minial server.js)

IgorMinar: ah. ok. thanks

IgorMinar: I love that we don't need to change the config. great work everyone!

PR #17, opened 2024-09-20
minor cleanup

vicb:


vicb:
I removed duplicate instruction from `TODO.md` (The update version is in `builder/README.md`) and move it inside the builder folder

PR #25, opened 2024-09-23
fix prereleases after rename of cloudflare package directory - take 2

petebacondarwin:


dario-piotrowicz (approved):
wouldn't the proper solution be updating this instead?
https://github.com/flarelabs-net/poc-next/blob/4fadf97885bba52c1a0a54e5970aab379f898f1c/.github/workflows/prereleases.yml#L38

🤔 

or if you do want to keep the code as is I'd suggest to rename `packagePath` to `packageName` for clarity

petebacondarwin:
> or if you do want to keep the code as is I'd suggest to rename `packagePath` to `packageName` for clarity

The comment at the top of that file refers to `package-name` so let's go with that.

PR #31, opened 2024-09-24
implement `getCloudflareContext` for production

dario-piotrowicz:
This PR adds a `getCloudflareContext` utility to the `@opennextjs/cloudflare` package so that developers can use it to get the Cloudflare bindings etc...

> [!WARNING]
> The current implementation is only for production/preview, it does not work with `next dev`

___

As part of this change I split the code in the cloudflare package into 2 sub directories, `cli` and `api`, the former contains the code that goes into the binary that gets run to build the worker while the latter exposes the programmatic APIs that the package offers (which is only `getCloudflareContext` right now)


examples/api/app/api/hello/route.ts
```diff
@@ -1,10 +1,18 @@
...
 
+import { getCloudflareContext } from "@opennextjs/cloudflare";
+
 export async function GET() {
-  // Note: we use headers just so that the route is not built as a static one
   const headersList = headers();
-  const sayHi = !!headersList.get("should-say-hi");
-  return new Response(sayHi ? "Hi World!" : "Hello World!");
+
+  const fromCloudflareContext = !!headersList.get("from-cloudflare-context");
```
vicb, line 8:
```suggestion
  const fromCloudflareContext = headersList.has("from-cloudflare-context");
```

examples/api/app/api/hello/route.ts
```diff
@@ -1,10 +1,18 @@
...
+
+  if (!fromCloudflareContext) {
+    return new Response("Hello World!");
+  }
+
+  const { env } = await getCloudflareContext();
```
vicb, line 14: nit: add a link to wrangler.toml?

packages/cloudflare/src/cli/build/patches/investigated/patch-cache.ts
```diff
@@ -1,5 +1,5 @@
 import path from "node:path";
-import { Config } from "../../../config";
+import { Config } from "../../../cli/config";
```
vicb, line 2: I 10 probably needs to be updated with the new path (`/cli`) but I can fix in a follow up

packages/cloudflare/src/cli/build/build-worker.ts
```diff
@@ -1,4 +1,4 @@
-import { Config } from "../config";
```
vicb, line 1: l 50 should be updated with the new path?

vicb (changes requested):
Added a few comments.

packages/cloudflare/src/api/get-cloudflare-context.ts
```diff
@@ -0,0 +1,49 @@
...
+  ctx: Context;
+};
+
+const cloudflareContextSymbol = Symbol.for("__cloudflare-context__");
+
+/**
+ * Utility to get the current Cloudflare context.
+ *
+ * Such context contains three values:
+ *  - `env`: object containing the worker's bindings
```
vicb, line 23: nit: 
- move the doc to l12-14?
- use links `[]()`?

packages/cloudflare/src/cli/build/patches/investigated/patch-cache.ts
vicb: Oh `copyPackage` will copy only the cli part - that's why you do not need to update the package.

I think it's fine for now.

irvinebroque (commented):
👏

examples/api/app/api/hello/route.ts
dario-piotrowicz: what do you mean a link? as an option to specify the toml file?

note that in production there is no toml file 😕

packages/cloudflare/src/api/get-cloudflare-context.ts
dario-piotrowicz: Done 🙂 

Good suggestion regarding `[]()` I didn't know vscode would display md-style links correctly 🙂

packages/cloudflare/src/cli/build/build-worker.ts
dario-piotrowicz: good point!

it currently works as intended but that's because we get a wrong "packageDir" ([here](https://github.com/opennextjs/opennextjs-cloudflare/pull/31/files#diff-3806ec13864197c25b18c92269913af6a5852467e9c476a4d46f1d3430192954L20)):
Screenshot 2024-09-25 at 12 49 08

I'll update this to make it correct

examples/api/app/api/hello/route.ts
vicb: Something like "// Retrieve the env defined in `wrangler.toml`"

packages/cloudflare/src/cli/build/build-worker.ts
vicb: That should help as I will need to ref `getRequestContext()` in the cache handler

packages/cloudflare/src/api/get-cloudflare-context.ts
```diff
@@ -0,0 +1,50 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-empty-interface
+  interface CloudflareEnv {}
+}
+
+export type CloudflareContext<
+  CfProperties extends Record<string, unknown> = IncomingRequestCfProperties,
+  Context = ExecutionContext,
+> = {
+  /**
+   * object containing the worker's bindings ([docs](https://developers.cloudflare.com/workers/runtime-apis/bindings/))
```
vicb, line 13: nit: 

- I would drop `object containing the` (same below) - I think it doesn't add value
- `worker's bindings ([docs](https://developers.cloudflare.com/workers/runtime-apis/bindings/))` -> `[worker's binding](https://developers.cloudflare.com/workers/runtime-apis/bindings/)` ?

packages/cloudflare/src/api/get-cloudflare-context.ts
```diff
@@ -0,0 +1,50 @@
...
+  /**
+   * the current execution context object ([docs](https://developers.cloudflare.com/workers/runtime-apis/context))
+   */
+  ctx: Context;
+};
+
+const cloudflareContextSymbol = Symbol.for("__cloudflare-context__");
+
+/**
+ * Utility to get the current Cloudflare context
```
vicb, line 29: Add a note that this throws when the context cannot be retrieved?

packages/cloudflare/src/api/get-cloudflare-context.ts
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @returns the cloudflare context
+ */
+export async function getCloudflareContext<
+  CfProperties extends Record<string, unknown> = IncomingRequestCfProperties,
+  Context = ExecutionContext,
+>(): Promise<CloudflareContext<CfProperties, Context>> {
+  const cloudflareContext = (
```
vicb, line 37: nit: a more concise version could be `const cloudflareContext = (globalThis as any)[cloudflareContextSymbol] as CloudflareContext<CfProperties, Context> | undefined;`

packages/cloudflare/src/cli/templates/worker.ts
```diff
@@ -1,46 +1,64 @@
...
 import { MockedResponse } from "next/dist/server/lib/mock-request";
 import NextNodeServer, { NodeRequestHandler } from "next/dist/server/next-server";
 import type { IncomingMessage } from "node:http";
+import type { CloudflareContext } from "../../api";
 
 const NON_BODY_RESPONSES = new Set([101, 204, 205, 304]);
 
+const cloudflareContextALS = new AsyncLocalStorage<CloudflareContext>();
+
+(globalThis as Record<symbol, unknown>)[Symbol.for("__cloudflare-context__")] = new Proxy(
```
vicb, line 14: nits:
- `as Record<symbol, unknown>` -> `as any` ?
- `Symbol.for("__cloudflare-context__")` should be defined once (in `get-cloudflare-context.ts`) and used at both places. An alternative could be to add a `setRequestContext()`?

packages/cloudflare/src/api/get-cloudflare-context.ts
dario-piotrowicz: It should not really throw in the final implementation tough...

I don't mind adding the comment for now

packages/cloudflare/src/cli/templates/worker.ts
```diff
@@ -1,46 +1,64 @@
...
+  }
+);
+
 // Injected at build time
 const nextConfig: NextConfig = JSON.parse(process.env.__NEXT_PRIVATE_STANDALONE_CONFIG ?? "{}");
 
 let requestHandler: NodeRequestHandler | null = null;
 
 export default {
   async fetch(request: Request, env: any, ctx: any) {
```
vicb, line 31: would the following work?
```
  async fetch(request: Request && { cf: IncomingRequestCfProperties }, env: any, ctx: any) {
    const cf = request.cf;
```

vicb (approved):
Added a couple commennts but looks good, thanks!

packages/cloudflare/src/api/get-cloudflare-context.ts
dario-piotrowicz: ok thanks, if you're ok with it let's keep it as is and we can always revisit it later on 🙂

packages/cloudflare/src/cli/templates/worker.ts
dario-piotrowicz: updated to `any` 👍

dario-piotrowicz: regarding the symbol, I agree that repeating it is not ideal, but mixing the template files and the api files seems a bit of a messy thing to me that's why I am trying to avoid that 😕 

anyways I did try to import the symbol from `../../api` but that does not work because once we copy those files `../../api` doesn't resolve correctly to the file we want: 
Screenshot 2024-09-25 at 14 02 34

we'd have to deal with different file names and extensions to get it to work:
Screenshot 2024-09-25 at 14 02 46

it does feel like a very slippery slope to take... what do you think? 😕

vicb: Right.
Maybe add a comment that it should be in sync.

dario-piotrowicz: comment added 🙂 

both here and in the other file

PR #33, opened 2024-09-25
enable `getCloudflareContext` to also work with `next dev`

dario-piotrowicz:
Followup for #31 there we introduced `getCloudflareContext` in a way that would work only in production/preview, this PR  adds support for it under `next dev` (via `getPlatformProxy`)

### DIFF
--- a/examples/api/playwright.dev.config.ts
+++ b/examples/api/playwright.dev.config.ts
@@ -1,0 +1,80 @@
+import { defineConfig, devices } from "@playwright/test";
+
+declare var process: { env: Record<string, string> };
+
+/**
+ * Read environment variables from file.
+ * https://github.com/motdotla/dotenv
+ */
+// import dotenv from 'dotenv';
+// dotenv.config({ path: path.resolve(__dirname, '.env') });
+
+/**
+ * See https://playwright.dev/docs/test-configuration.
+ */
+export default defineConfig({
+  testDir: "./e2e-tests",
+  /* Run tests in files in parallel */
+  fullyParallel: true,
+  /* Fail the build on CI if you accidentally left test.only in the source code. */
+  forbidOnly: !!process.env.CI,
+  /* Retry on CI only */
+  retries: process.env.CI ? 2 : 0,
+  /* Opt out of parallel tests on CI. */
+  workers: process.env.CI ? 1 : undefined,
+  /* Reporter to use. See https://playwright.dev/docs/test-reporters */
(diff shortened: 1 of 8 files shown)


packages/cloudflare/src/api/get-cloudflare-context.ts
```diff
@@ -29,25 +29,55 @@ const cloudflareContextSymbol = Symbol.for("__cloudflare-context__");
 /**
  * Utility to get the current Cloudflare context
  *
- * Throws an error if the context could not be retrieved
```
dario-piotrowicz, line 32: I was thinking, instead of this we could add a sentence saying something like:
```
During development a local approximation/simulation of the real context is provided
```
or something along those lines

but truthfully I think that devs won't really care that much about this and that it's probably better to keep things simpler here...

what do you think?

vicb: Simpler is better 👍

examples/api/playwright.dev.config.ts
```diff
@@ -0,0 +1,80 @@
+import { defineConfig, devices } from "@playwright/test";
+
+declare var process: { env: Record<string, string> };
+
+/**
+ * Read environment variables from file.
+ * https://github.com/motdotla/dotenv
+ */
+// import dotenv from 'dotenv';
```
vicb, line 9: Could this be removed?

examples/api/playwright.dev.config.ts
```diff
@@ -0,0 +1,80 @@
...
+      name: "firefox",
+      use: { ...devices["Desktop Firefox"] },
+    },
+
+    {
+      name: "webkit",
+      use: { ...devices["Desktop Safari"] },
+    },
+
+    /* Test against mobile viewports. */
```
vicb, line 53: As well as this

vicb (approved):

examples/api/playwright.dev.config.ts
dario-piotrowicz: all of our playwright config files contain all of this default boilterplate content

I don't have a preference regarding keeping it or removing it, but if we want to remove it I'd suggest to just do that as a followup PR for all of our playwright config files

examples/api/playwright.dev.config.ts
dario-piotrowicz: as above 🙂

PR #38, opened 2024-09-26
Commerce

vicb:


examples/vercel-commerce/lib/shopify/index.ts
```diff
@@ -205,6 +205,10 @@ const reshapeProducts = (products: ShopifyProduct[]) => {
 };
 
 export async function createCart(): Promise<Cart> {
+  if (!process.env.SHOPIFY_STORE_DOMAIN) {
```
vicb, line 208: I think we should avoid modifying the app more than what is needed to run on workers.
If this is to help we the tests, we can skip them when there is an error message.

examples/vercel-commerce/e2e/base.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  await expect(page.getByRole('link', { name: 'Created by ▲ Vercel' })).toBeVisible();
+});
+
+test('the actual commerce functionality', async ({ page }) => {
+  test.skip(!process.env.SHOPIFY_STORE_DOMAIN, 'Populate the .dev.vars file to run the test');
+
+  await page.goto('/');
+
+  const hanraShirtLink = page.getByRole('link', { name: 'Hanra Shirt Hanra Shirt $108.' });
```
vicb, line 21: This test is brittle as it depends on the content of the shop.
At least we should test what shop we are serving but even then it makes little sense if we do not control the content of that shop.

vicb:
@dario-piotrowicz thanks for fixing the PR.

LGTM with a few comments:
- I wouldn't update the app more than what is needed to run on workers
- I wouldn't test things depending on the shop content that we don't control

examples/vercel-commerce/e2e/base.spec.ts
dario-piotrowicz: yes I totally agree that this is quite brittle, I just didn't want to modify the app more than necessary so, with the html in place there I could not find something better to test this functionality

I am happy to remove this or modify the app to be more easily testable, whatever you prefer 👍

examples/vercel-commerce/lib/shopify/index.ts
dario-piotrowicz: based on this and your other comment I can remove all the modification I did just to enable testing the commerce functionality

examples/vercel-commerce/e2e/base.spec.ts
dario-piotrowicz: oh... I just realized, I could very easily mock the network requests.... 🤔 🤷

examples/vercel-commerce/lib/shopify/index.ts
dario-piotrowicz: removed

examples/vercel-commerce/e2e/base.spec.ts
vicb: IMO a basic test as the first one is enough for now

dario-piotrowicz:
@vicb the tweaks I made to the app were actually necessary to get any e2e to work

so I've removed the e2es altogether... let me know what you prefer, we can either keep the e2es but re-add some slight tweaks to the app or have the app without any sort of e2e...

vicb:
@dario-piotrowicz it's becoming hard to review, could you squash your commits in fewer commits to ease the review.

> the tweaks I made to the app were actually necessary to get any e2e to work

I'm not sure to understand the root cause here.
If the CI creates the `.dev.vars` then it should be able to bring up the app?
But I guess it's not a big deal anyway.

dario-piotrowicz:
> @dario-piotrowicz it's becoming hard to review, could you squash your commits in fewer commits to ease the review.

sure

> > the tweaks I made to the app were actually necessary to get any e2e to work
> 
> I'm not sure to understand the root cause here.
> If the CI creates the .dev.vars then it should be able to bring up the app?

I've remove that part, I think that we do not want the e2es to work only when the `.dev.vars` file is fully populated because this would be problematic for external contributors that would be forced to get a shopify key to get the e2es to work (adding a significant friction point in my opinion)

And the `wrangler dev` command without all the `.dev.vars` although works if you view the app with your browser is not getting detected as working properly by playwright (I suspect that playwright is getting some 500s when trying to see if the server is running)

dario-piotrowicz:
@vicb I've rebased and squashed my commits, please have another look 🙏

vicb:
Looks good, thanks for making this work!

PR #49, opened 2024-09-30
run the `next build` command using the correct package manager

### LINKED ISSUE #45: Yarn Support
Hey, I'm trying to install this package (following the Getting Started guide from https://opennext.js.org/cloudflare/get-started) on an existing project that uses Yarn v1.22.18. However, the build fails because it seems to require PNPM. Is this an intentional design choice, or will the package support other package managers like Yarn?

This also happens with NPM if npm is set on the packageManager  on package.json

```
UsageError: This project is configured to use yarn because package.json has a "packageManager" field
    at Engine.findProjectSpec (.....)

Node.js v22.6.0
node:internal/errors:983
  const err = new Error(message);
              ^

Error: Command failed: pnpm next build```

### PR DESCRIPTION
**Pull request by dario-piotrowicz**
fixes #45

### DIFF
--- a/packages/cloudflare/src/cli/build/build-next-app.ts
+++ b/packages/cloudflare/src/cli/build/build-next-app.ts
@@ -1,3 +1,4 @@
+import { detect } from "package-manager-detector";
 import { execSync } from "node:child_process";
 
 /**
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: packages/cloudflare/src/cli/build/build-next-app.ts [hunk 1]
```diff
@@ -7,17 +8,15 @@ import { execSync } from "node:child_process";
  *
  * @param nextAppDir the directory of the app to build
  */
-export function buildNextjsApp(nextAppDir: string): void {
-  runNextBuildCommand("pnpm", nextAppDir);
+export async function buildNextjsApp(nextAppDir: string): Promise<void> {
+  const pm = await detect();
+
+  runNextBuildCommand(pm?.name ?? "npm", nextAppDir);
```
vicb, line 14: Should the code throw if not pm is detected?

packages/cloudflare/src/cli/build/build-next-app.ts
```diff
@@ -7,17 +8,15 @@ import { execSync } from "node:child_process";
...
 }
 
 // equivalent to: https://github.com/sst/open-next/blob/f61b0e94/packages/open-next/src/build.ts#L175-L186
-function runNextBuildCommand(
-  // let's keep things simple and just support only pnpm for now
-  packager: "pnpm" /*"npm" | "yarn" | "pnpm" | "bun"*/,
-  nextAppDir: string
-) {
-  const command = ["bun", "npm"].includes(packager) ? `${packager} next build` : `${packager} next build`;
+function runNextBuildCommand(packager: "npm" | "pnpm" | "yarn" | "bun", nextAppDir: string) {
```
vicb, line 18: nit: use [`AgentName`](https://github.com/antfu-collective/package-manager-detector/blob/5c494d7a2fcad6a900d0a666bca5601ade003aa9/src/types.ts#L53) for the type

dario-piotrowicz: yeah I considered that, but I dislike that type name 😅 

would be ok if I included it but rename it?
as in:
```
type PM = AgentName;
```

packages/cloudflare/src/cli/build/build-next-app.ts
dario-piotrowicz: sure 👍

packages/cloudflare/src/cli/build/build-next-app.ts
dario-piotrowicz: I've ended up with `import { type AgentName as PackageManager } from "package-manager-detector";`

Hopefully this works for you 😄

packages/cloudflare/src/cli/build/build-next-app.ts
dario-piotrowicz: updated 🙂

PR #51, opened 2024-10-02
specify inspector port to playwright preview:worker commands to avoid flakiness in e2es

dario-piotrowicz:
I think that running different `wrangler dev` instances in parallel can introduce flakiness as wrangler might try to reuse the same inspector port (which [defaults to 9229](https://github.com/cloudflare/workers-sdk/blob/9e977b5297ce77f9b520c620b5c25dc93c46d68e/packages/wrangler/src/index.ts#L87))

Example (with no code changes):
 - [failed run](https://github.com/opennextjs/opennextjs-cloudflare/actions/runs/11142501994/job/30965579858#step:7:13)
 - [failed run](https://github.com/opennextjs/opennextjs-cloudflare/actions/runs/11142501994/job/30965708704#step:7:13)
 - [successful run](https://github.com/opennextjs/opennextjs-cloudflare/actions/runs/11142501994/job/30965921993#step:7:26)

@petebacondarwin what do you think? is this a known wrangler issue?

### DIFF
--- a/examples/api/e2e/playwright.config.ts
+++ b/examples/api/e2e/playwright.config.ts
@@ -46,7 +46,7 @@ export default defineConfig({
 
   /* Run your local dev server before starting the tests */
   webServer: {
-    command: "pnpm preview:worker --inspector-port 9330",
+    command: "pnpm preview:worker",
     url: "http://localhost:8770",
     reuseExistingServer: !process.env.CI,
   },
--- a/examples/api/e2e/playwright.config.ts
+++ b/examples/api/e2e/playwright.config.ts
@@ -46,7 +46,7 @@ export default defineConfig({
 
   /* Run your local dev server before starting the tests */
   webServer: {
-    command: "pnpm preview:worker",
+    command: "pnpm preview:worker --inspector-port 9330",
     url: "http://localhost:8770",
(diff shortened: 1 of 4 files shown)


dario-piotrowicz:
Another flake: https://github.com/opennextjs/opennextjs-cloudflare/actions/runs/11148273104/job/30984521056?pr=52#step:7:13 🥲

examples/api/e2e/playwright.config.ts
```diff
@@ -46,7 +46,7 @@ export default defineConfig({
 
   /* Run your local dev server before starting the tests */
   webServer: {
-    command: "pnpm preview:worker",
+    command: "pnpm preview:worker --inspector-port 9330",
```
workers-frameworks, line 49: I think the correct place to fix this is in the `package.json` `"dev:worker"` scripts, where we already hard code `--port` for each of these examples.

workers-frameworks (changes requested):

examples/api/e2e/playwright.config.ts
dario-piotrowicz: good point 🙂 

updated 👍

petebacondarwin (approved):

PR #55, opened 2024-10-03
fix the wrong eslint version being used in examples/create-next-app

dario-piotrowicz:
it's quite minor, but we do get this eslint error when running the `create-next-app` example:
Screenshot 2024-10-03 at 12 22 12

[example in a gh run](https://github.com/opennextjs/opennextjs-cloudflare/actions/runs/11160744681/job/31021848635#step:7:12)

I think this is caused by us using the wrong eslint version in the app, so that's what I am addressing in this PR


pnpm-workspace.yaml
```diff
@@ -8,7 +8,6 @@ catalog:
   "@types/node": ^22.2.0
   "@types/react": ^18
   "@types/react-dom": ^18
-  "eslint-config-next": 14.2.11
```
vicb, line 11: Could you please sort the lines here?
(I first thought eslint was missing from here)

dario-piotrowicz: done 🙂

vicb (approved):
Thanks!

PR #64, opened 2024-10-06
feat: use incremental cache during rendering

james-elicx:
Changes:

- Removes the `NEXT_PRIVATE_MINIMAL_MODE` env var.
- Inlines the middleware manifest instead of it being dynamic require.
- Removes the no fallback exception bubbling as this results in hanging promises when SSG routes disable fallback dynamic params.

---

We currently enable the private version of minimal mode, alongside the public version of minimal mode. This is problematic as it is intended to be used by cloud providers when they have an additional layer on top of the worker that handles routing systems, cdn caching, static routes, etc.. That's why Vercel uses it. As a result, we're disabling specific behaviour without building our own system that will handle that.

There's two ways to look at this;
1. Enable the internal minimal mode -> build our own routing system, cdn caching logic, static routes/isr handling, etc.
2. Disable the internal minimal mode -> let Next.js handle routing and provide a custom cache handler to the incremental cache, and things _should_ work.

The option that is probably going to be easier to maintain and support in the long run is number 2 - disabling the internal minimal mode and just providing a custom cache handler that the incremental cache can use, which we are already partly doing.

---

Here's what happens in the base server when making a request to a route that is SSG/ISR.

- Finds a match for the route.
- Starts the logic for rendering a route module.
  - Calls the response cache with a callback that will SSR the page. [(src)](https://github.com/vercel/next.js/blob/c53ef9b3cd63845ba8165c720d84669e97d0b513/packages/next/src/server/base-server.ts#L3069)
  - Response cache calls the incremental cache. [(src)](https://github.com/vercel/next.js/blob/c53ef9b3cd63845ba8165c720d84669e97d0b513/packages/next/src/server/response-cache/index.ts#L96)
    - Incremental cache interfaces with the custom cache handler to find a cache entry for the key.
  - If cache entry should be used / is fresh
    - the response cache returns instead of calling the render callback. [(src)](https://github.com/vercel/next.js/blob/c53ef9b3cd63845ba8165c720d84669e97d0b513/packages/next/src/server/response-cache/index.ts#L111-L120)
  - If cache entry is stale / cache miss
    - the response cache calls the render callback that it was given. [(src)](https://github.com/vercel/next.js/blob/c53ef9b3cd63845ba8165c720d84669e97d0b513/packages/next/src/server/response-cache/index.ts#L124)
    - it tries to put the response from SSR into the incremental cache.
    - if there was no response from the render callback due to 404 and there is no fallback function for it to invoke, it bubbled a NoFallbackException back to the worker.
- Response returned to user.

The problem with using the private minimal mode for this is that the incremental cache is [disabled](https://github.com/vercel/next.js/blob/c53ef9b3cd63845ba8165c720d84669e97d0b513/packages/next/src/server/response-cache/index.ts#L96-L102) in the response cache, blocking ssg/isr from functioning!

### DIFF
--- a/packages/cloudflare/src/cli/build/patches/to-investigate/inline-middleware-manifest-require.ts
+++ b/packages/cloudflare/src/cli/build/patches/to-investigate/inline-middleware-manifest-require.ts
@@ -21,7 +21,7 @@ export function inlineMiddlewareManifestRequire(code: string, config: Config) {
   );
 
   if (patchedCode === code) {
-    throw new Error("Cache patch not applied");
+    throw new Error("Cache patch `inlineMiddlewareManifestRequire` not applied");
   }
 
   return patchedCode;
--- a/packages/cloudflare/src/cli/build/patches/to-investigate/inline-middleware-manifest-require.ts
+++ b/packages/cloudflare/src/cli/build/patches/to-investigate/inline-middleware-manifest-require.ts
@@ -1,0 +1,19 @@
+import { existsSync, readFileSync } from "node:fs";
+import { Config } from "../../../config";
+import path from "node:path";
+
+/**
+ * Inlines the middleware manifest from the build output to prevent a dynamic require statement
+ * as they result in runtime failures.
+ */
+export function inlineMiddlewareManifestRequire(code: string, config: Config) {
+  console.log("# inlineMiddlewareManifestRequire");
+
+  const middlewareManifestPath = path.join(config.paths.standaloneAppServer, "middleware-manifest.json");
+
+  const middlewareManifest = existsSync(middlewareManifestPath)
+    ? JSON.parse(readFileSync(middlewareManifestPath, "utf-8"))
+    : {};
+
+  return code.replace(/require\(this.middlewareManifestPath\)/, JSON.stringify(middlewareManifest));
+}
--- a/packages/cloudflare/src/cli/build/patches/to-investigate/inline-middleware-manifest-require.ts
+++ b/packages/cloudflare/src/cli/build/patches/to-investigate/inline-middleware-manifest-require.ts
@@ -15,5 +15,14 @@ export function inlineMiddlewareManifestRequire(code: string, config: Config) {
     ? JSON.parse(readFileSync(middlewareManifestPath, "utf-8"))
     : {};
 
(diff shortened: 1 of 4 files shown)


packages/cloudflare/src/cli/build/patches/to-investigate/inline-middleware-manifest-require.ts
```diff
@@ -0,0 +1,19 @@
...
+export function inlineMiddlewareManifestRequire(code: string, config: Config) {
+  console.log("# inlineMiddlewareManifestRequire");
+
+  const middlewareManifestPath = path.join(config.paths.standaloneAppServer, "middleware-manifest.json");
+
+  const middlewareManifest = existsSync(middlewareManifestPath)
+    ? JSON.parse(readFileSync(middlewareManifestPath, "utf-8"))
+    : {};
+
+  return code.replace(/require\(this.middlewareManifestPath\)/, JSON.stringify(middlewareManifest));
```
vicb, line 18: 2 minor things:
- You should not need a regexp as the first arg of replace,
- Please check that [some patch was applied](https://github.com/opennextjs/opennextjs-cloudflare/blob/bcaaec6766928e362ef23f6b746789f33e1e6db7/packages/cloudflare/src/cli/build/patches/investigated/patch-cache.ts#L20) or throw an error - this will help debugging errors.

The same applies for the other patch.

vicb (changes requested):
Thanks for the PR.

The changes look great - could you please update the PR according to the comment?

packages/cloudflare/src/cli/build/patches/to-investigate/inline-middleware-manifest-require.ts
```diff
@@ -0,0 +1,28 @@
...
+    ? JSON.parse(readFileSync(middlewareManifestPath, "utf-8"))
+    : {};
+
+  const patchedCode = code.replace(
+    "require(this.middlewareManifestPath)",
+    JSON.stringify(middlewareManifest)
+  );
+
+  if (patchedCode === code) {
+    throw new Error("Cache patch not applied");
```
vicb, line 24: Can you please update the name of the patch (here and in the other file)

vicb (changes requested):
Looks great, thanks for the updates!

Could you please update the name of the patches and this will be good to go

vicb (approved):
Thanks again 🎉

PR #65, opened 2024-10-06
feat: ssg support through seeding the incremental cache

james-elicx:
> [!NOTE]
> This PR depends on #64 being merged, and then rebased with main, to be utilised properly.

**Context**

At the moment, static routes are being SSR'd, despite getting pre-rendered at build-time in the app directory. This means we have a series of static assets that the Next.js server wants to be able to serve during rendering. At the moment, there is no way for the incremental cache to retrieve these static assets. This is because we have a cache handler that doesn't interact with the file-system for the seed data.

**Changes**

- Copies prerendered routes to the static assets directory.
- Changes the cache handler logging to be done when debug mode is enabled, like the default Next.js behaviour.
- Improves types used for cache handler to better align with Next.js types.
- Supports serving seed data from the cache handler `get` method when a KV miss happens.
  - Implemented based on the Next.js File System cache at two points in time so that it supports some backwards compatibility over the last year.
  - https://github.com/vercel/next.js/blob/canary/packages/next/src/server/lib/incremental-cache/file-system-cache.ts
  - https://github.com/vercel/next.js/blob/33586b60e3f21779e3a01df02a0bd9774e70010a/packages/next/src/server/lib/incremental-cache/file-system-cache.ts#L52

---

Future thought: Can asset fetchers expose a way to list the assets that they have within them, or to check the existence of an asset instantly?


packages/cloudflare/src/cli/cache-handler/utils.ts
```diff
@@ -0,0 +1,41 @@
...
+export function getSeedBodyFile(key: string, suffix: string) {
+  return getAsset(`${key}${suffix}`, (resp) => resp.arrayBuffer() as Promise<Buffer>);
+}
+
+export function getSeedTextFile(key: string, suffix: string) {
+  return getAsset(`${key}${suffix}`, (resp) => resp.text());
+}
+
+export function getSeedMetaFile(key: string) {
+  return getAsset(`${key}${NEXT_META_SUFFIX}`, (resp) => resp.json<PrerenderedRouteMeta>());
```
vicb, line 34:
```suggestion
  return getAsset(key + NEXT_META_SUFFIX, (resp) => resp.json<PrerenderedRouteMeta>());
```

vicb:
> Future thought: Can asset fetchers expose a way to list the assets that they have within them, or to check the existence of an asset instantly?

We can generate an index at build time but at some point we should populate the KV when building the app

packages/cloudflare/src/cli/build/utils/copy-prerendered-routes.ts
```diff
@@ -0,0 +1,46 @@
...
+      prerenderedRoutes.includes(relativePath.replace(/\.\w+$/, "").replace(/^\/index$/, "/"))
+    );
+
+  prerenderedAssets.forEach(({ fullPath, relativePath }) => {
+    const destPath = join(outputPath, relativePath);
+    mkdirSync(dirname(destPath), { recursive: true });
+
+    if (fullPath.endsWith(NEXT_META_SUFFIX)) {
+      const data = JSON.parse(readFileSync(fullPath, "utf8"));
+      writeFileSync(destPath, JSON.stringify({ ...data, lastModified: Date.now() }));
```
vicb, line 41: nit: pass the date as an argument to `copyPrerenderedRoutes` so that all the assets have a consistent `lastModified` date

packages/cloudflare/src/cli/cache-handler/constants.ts
```diff
@@ -0,0 +1,8 @@
+export const RSC_PREFETCH_SUFFIX = ".prefetch.rsc";
+export const RSC_SUFFIX = ".rsc";
+export const NEXT_DATA_SUFFIX = ".json";
+export const NEXT_META_SUFFIX = ".meta";
+export const NEXT_BODY_SUFFIX = ".body";
+export const NEXT_HTML_SUFFIX = ".html";
+
+export const SEED_DATA_DIR = "_cf_seed_data";
```
vicb, line 8: It would be great to save it under `/cdn-cgi` so that the files are not publicly accessible

vicb (changes requested):
Very nice, thanks for the PR!

I added a few comments. Please address before we merge the PR.

packages/cloudflare/src/cli/build/utils/copy-prerendered-routes.ts
james-elicx: good shout, i've added the timestamp to the config object instead of an arg - that way it's easier to re-use everywhere

james-elicx:
> We can generate an index at build time but at some point we should populate the KV when building the app

My question is around asset bindings exposing a way instead of us generating an index. I recall a year or so a go, Greg posed the question of would it be useful if the Pages asset binding had a way to check if assets existed / get a list. I don't think anything ever came from that though.

packages/cloudflare/src/cli/config.ts
```diff
@@ -11,6 +11,9 @@ const UserConfig = {
 };
 
 export type Config = {
+  // Timestamp for when the build was started
+  timestamp: number;
```
vicb, line 15: nit: can you rename to `buildTimestamp`

packages/cloudflare/src/cli/cache-handler/open-next-cache-handler.ts
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    // Check for seed data from the file-system.
+
+    // we don't check for seed data for fetch or image cache entries
+    if (ctx?.kind === "FETCH" || ctx?.kind === "IMAGE") return null;
+
+    const seedKey = `http://assets.local/${SEED_DATA_DIR}/${key}`.replace(/\/\//g, "/");
+
+    if (ctx?.kind === "APP" || ctx?.kind === "APP_ROUTE") {
```
conico974, line 52: Shouldn't this be `ctx?.kind === "ROUTE"` instead of `APP` ?

packages/cloudflare/src/cli/config.ts
dario-piotrowicz: wouldn't it `buildStartTimestamp` be clearer?

making the variable self descriptive and avoiding the need for the extra comment

vicb (approved):

vicb:
Thanks!

packages/cloudflare/src/cli/cache-handler/open-next-cache-handler.ts
```diff
@@ -0,0 +1,146 @@
...
+          status: meta?.status,
+          headers: meta?.headers,
+        },
+      };
+    }
+
+    return null;
+  }
+
+  async set(...args: Parameters<CacheHandler["set"]>) {
```
conico974, line 114: ISR revalidation runs in the background, it doesn't look like there is anything awaiting for the revalidation to finish

packages/cloudflare/src/cli/config.ts
vicb: Oops sorry @dario-piotrowicz I merged before seeing your comment.
I feel like `buildTimestamp` is good enough but feel free to update

packages/cloudflare/src/cli/cache-handler/open-next-cache-handler.ts
james-elicx: i haven't changed this part - this pr was only implementing seed data. we dont support isr yet from my testing

packages/cloudflare/src/cli/cache-handler/open-next-cache-handler.ts
james-elicx: (not checked pages directory, so that might apply there, but we list pages dir as unsupported at the moment)

conico974: You're right, i missed the `parseCtx` function.
The `.body` is only for cached app route anyway, so that's fine

PR #72, opened 2024-10-07
refactor: retrieve cache handler kv instance inside constructor

james-elicx:
**Context**

At the moment, the KV instance that is used by the cache handler is assigned to the class as a static property as part of the code patch that happens. We should move this to happen during the creation of a new instance of the class.

Normally, all logic related to setting up the cache handler would happen within the cache handler itself, whereas it currently happens externally. This also means that in the long-term, it makes it easier to extend the cache handler to support additional mechanisms/customisability should we desire (e.g. R2).

**Changes**
- Moves the cache handler to the templates dir and builds it to the output directory in the cli with an inlined env var for the name of the KV binding.
- Changes the patch to import the cache handler from the output directory instead of the cli.
- Retrieves KV instance in constructor.


packages/cloudflare/src/cli/build/utils/copy-prerendered-routes.ts
```diff
@@ -1,4 +1,4 @@
-import { NEXT_META_SUFFIX, SEED_DATA_DIR } from "../../cache-handler";
+import { NEXT_META_SUFFIX, SEED_DATA_DIR } from "../../templates/cache-handler";
```
vicb, line 1: nit: It doesn't sound right to import from a template

packages/cloudflare/src/cli/build/build-worker.ts
```diff
@@ -66,6 +69,18 @@ export async function buildWorker(config: Config): Promise<void> {
   patchWranglerDeps(config);
   updateWebpackChunksFile(config);
 
+  await build({
```
vicb, line 72: nit: move this to the patch

packages/cloudflare/src/cli/templates/cache-handler/open-next-cache-handler.ts
```diff
@@ -13,29 +13,31 @@ import {
 } from "./constants";
 import { getSeedBodyFile, getSeedMetaFile, getSeedTextFile, parseCtx } from "./utils";
 import type { IncrementalCacheValue } from "next/dist/server/response-cache";
-import { KVNamespace } from "@cloudflare/workers-types";
+import type { KVNamespace } from "@cloudflare/workers-types";
```
vicb, line 16: thought: we should add this to the ESLint config

vicb (commented):
Thanks for the PR.

It looks great with a few minor comments.

---

A bit off-topic to this PR but more details on our plan since you are a major contributor to this repo:

The plan is that we move/re-use code from [opennextjs-aws](https://github.com/opennextjs/opennextjs-aws).

By sharing code and architecture both repos will benefit for the fixes/features.

The first notable change will be that we will run the routing + middleware layer that is not currently supported in this repo. 

I hope to have some code ready later this week.

Thanks for all you contribs!

packages/cloudflare/src/cli/build/build-worker.ts
james-elicx: Shall do. My thinking here was the reason we'd have the patch is because the dynamic import on the cache handler's path doesn't work when you use the next.js config `cacheHandler` option to specify, so there needs to be a patch for initialising the cache, but the building of the cache handler itself behaves as expected and isn't part of a patch to get it to work.

packages/cloudflare/env.d.ts
```diff
@@ -5,6 +5,7 @@ declare global {
       __NEXT_PRIVATE_STANDALONE_CONFIG?: string;
       SKIP_NEXT_APP_BUILD?: string;
       NEXT_PRIVATE_DEBUG_CACHE?: string;
+      __OPENNEXT_KV_BINDING_NAME: string;
```
vicb, line 8: @james-elicx Is this documented already? 
If not, would you mind sending a PR?
Also maybe add a default value.
Thanks!

james-elicx: The intention was for this to just be used internally to inline a value into the cache handler, but yeah, it would make sense to let users use it as well

PR #73, opened 2024-10-07
chore: defer people to docs for getting started

james-elicx:
I don't think it makes sense to maintain two separate guides, especially when they both have a lot of overlap. We should just maintain the guide for starting/migrating apps in the docs and defer users to that page.


packages/cloudflare/README.md
```diff
@@ -2,109 +2,8 @@
```
vicb, line 2: I think we should keep the content here as it appears on https://www.npmjs.com/package/@opennextjs/cloudflare

What do other think?

dario-piotrowicz (approved):
I'd be happy to proceed with this PR's changes

@vicb please let me know what you think, if you strongly disagree we can think of something else and decline the PR, I'd just avoid to keep this lingering indefinitely

packages/cloudflare/README.md
petebacondarwin: I don't feel strongly but would also be happy to land this if @vicb is not strongly against it. Avoiding having to maintain two lots of docs is a good idea. That being said, do we have a plan (other than @dario-piotrowicz's suggestion above) to ensure the docs are kept up to date?

petebacondarwin (approved):

packages/cloudflare/README.md
vicb: Sure 🚀

PR #83, opened 2024-10-09
feat: configure kv binding name with env var

james-elicx:
**Changes**
- Defaults the `__OPENNEXT_KV_BINDING_NAME` env var to the default kv name.
  - The typescript types mark this variable as defined due to its use for inlining in the cache handler, so defaulting the env var itself makes sense I think.
- Uses the env var in the config for the kv name.
- Adds more context to the cache handler patch.

docs pr: https://github.com/opennextjs/docs/pull/24


vicb (approved):
Awesome, thanks James!

Issue #82, opened 2024-10-09
Error: Not in a Next.js app project


megaconfidence:
I'm trying to deploy this Nextjs template on Workers using opennext but I get this error

https://vercel.com/templates/next.js/empathic-voice-interface-starter
https://vercel.com/templates/next.js/blog-starter-kit
```
Building the Next.js app in the current folder (/Users/megaconfidence/Sandbox/temp/hume-evi-next-js-starter)
file:///Users/megaconfidence/Sandbox/temp/hume-evi-next-js-starter/node_modules/@opennextjs/cloudflare/dist/cli/index.mjs:7291
  throw new Error("Error: Not in a Next.js app project");
        ^

Error: Error: Not in a Next.js app project
    at file:///Users/megaconfidence/Sandbox/temp/hume-evi-next-js-starter/node_modules/@opennextjs/cloudflare/dist/cli/index.mjs:7291:9
    at ModuleJob.run (node:internal/modules/esm/module_job:234:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:473:24)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:123:5)

Node.js v20.17.0
```

shashankboosi:
Hey @megaconfidence,

You are getting the issue because the adapter only supports next projects which have the next.config file. If the adapter doesn't find it, will throw the error. So to fix your issue add the `next.config` file. I have posted the bare minimum code with defaults below

```js
/** @type {import('next').NextConfig} */
const nextConfig = {};

export default nextConfig;
```

I have raised a PR to add a detailed error going forward for other users. I hope this helps!

cc @vicb

megaconfidence:
Thanks @shashankboosi. I've added the config file and local dev in Nodejs works. But when I try to build for Workerd with `preview:workers`, I get the following error in the browser:

<img width="864" alt="image" src="https://github.com/user-attachments/assets/c188b234-7d9f-4bbb-9ad0-db19786fdda7">

vicb:
@shashankboosi both improved the error message and also added the blog starter app, see #86
_It will be merged on the feedback is adressed_

shashankboosi:
Hey @megaconfidence,

As @vicb said, the PR mentioned should work and it is night time here. So will try to get the PR ready for merge tomorrow. Play around with the PR till then :)

megaconfidence:
Thanks, everything seems to work as expected in the PR. Thanks for working on this

PR #84, opened 2024-10-10
Expand missing next.config error message

shashankboosi:
Fixes https://github.com/opennextjs/opennextjs-cloudflare/issues/82

1) Found out that Next can run it's application without next.config but if we are using the adapter we throw an error if we don't find the config. So added a detailed error to point towards the next.config to run the application.
2) Updated wrangler to it's latest version 
3) Also added the vercel blog starter into the examples which is what I used to debug the issue mentioned


vicb (changes requested):
Looks like the PR contains quite a fex unwanted changes!

Could you please also add a patch changeset (see the automated comment on the PR).

Thanks a lot

vicb:
OOps, I should have read the PR description.

Would you mind splitting 1, 2, and 3 into different PRs?

Thanks!

shashankboosi:
@vicb Will create different PRs, should I add patch changeset for all the PRs?

vicb:
Thanks!

> should I add patch changeset for all the PRs?

I would say only the error message fix.
It's not the biggest change but it will be very valuable to users.

shashankboosi:
I will close this and open separate PRs. Thanks!

PR #86, opened 2024-10-10
Add vercel blog starter to examples

shashankboosi:
Added the vercel blog starter to examples as a way to test applications. I used this from vercel which wasn't working but fixed it to the needs of the adapter

You can run the new example using `pnpm --filter vercel-blog-starter preview:worker`


pnpm-lock.yaml
vicb, line 1: Could you please remove this file.
It should not be there in the monorepo setup.
Thanks!

vicb:
>  I used this from vercel which wasn't working but fixed it to the needs of the adapter

One thing you can do is two separate commit:
- first commit would be "import blog start at <sha>"
- second commit would be "adapt for cloudflare worker"

It's nice to have the required changes in the history

shashankboosi:
Made all the recommended changes and had the commits in the flow, that is nice to read

shashankboosi:
I had to delete the commit where I remove `pnpm-lock` because currently there is no way to run github actions without a lock file and there is an open issue about it [here](https://github.com/actions/setup-node/issues/928).

Maybe if you want, I can investigate it as a separate issue

vicb (changes requested):
LGTM, thank you so much!

2 things:
- could you please add [a patch changeset](https://github.com/opennextjs/opennextjs-cloudflare/pull/86#issuecomment-2405701734)
- it would be nice to add a basic e2e test, you can check i.e. the API example. If you are willing to do this, you can do it in a follow up PR when you have time. No rush.

shashankboosi:
@vicb I have added the changeset. In regards to adding an e2e test, I will do it in another PR as I don't want to keep holding this one :)

vicb:
> I don't want to keep holding this one :)

Yep, makes sense.

Thank for adding the example!

vicb (approved):

PR #88, opened 2024-10-10
refactor: project options type shared with build + config init

james-elicx:
**Context**

We have what seems to be a pre-options and then options. It would help in the long-term to have a single set of options that we use to stub the project setup and the config creation, especially if we add more CLI options in the future. Therefore, the types are slightly refactored to share the same thing between the creation of the project config and the project setup function.

I don't mind if you want to keep it as is, but I think it would be simpler to have a single object used between the two.

I wanted to move the config creation to before the project setup outside but it currently depends on the build happening in order to generate the config unfortunately for the .next server directory.

**Changes**

- Shares the same input between the two.
- Renames `builderOutput` to `outputDir` and `nextApp` to `sourceDir`, as it's the output directory and the source directory.

basically centralises where i need to make a change if i want to add a new cli arg (e.g. https://github.com/opennextjs/opennextjs-cloudflare/pull/89) and use it in the config, otherwise you need to update multiple types and args normally

### DIFF
--- a/packages/cloudflare/src/cli/build/index.ts
+++ b/packages/cloudflare/src/cli/build/index.ts
@@ -1,8 +1,9 @@
 import { containsDotNextDir, getConfig } from "../config";
+import type { ProjectOptions } from "../config";
 import { buildNextjsApp } from "./build-next-app";
 import { buildWorker } from "./build-worker";
 import { cpSync } from "node:fs";
-import path from "node:path";
+import { join } from "node:path";
 import { rm } from "node:fs/promises";
 
 /**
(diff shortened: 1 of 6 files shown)


packages/cloudflare/src/cli/build/index.ts
```diff
@@ -1,46 +1,39 @@
...
  *
  * It saves the output in a `.worker-next` directory
  *
- * @param appDir the directory of the Next.js app to build
- * @param opts.outputDir the directory where to save the output (defaults to the app's directory)
- * @param opts.skipBuild boolean indicating whether the Next.js build should be skipped (i.e. if the `.next` dir is already built)
+ * @param opts The options for the project
  */
-export async function build(appDir: string, opts: BuildOptions): Promise<void> {
+export async function build(opts: ProjectOptions): Promise<void> {
```
vicb, line 16: what about renaming `opts` -> `project(Opts)` ?

packages/cloudflare/src/cli/config.ts
```diff
@@ -11,14 +11,18 @@ const UserConfig = {
 };
 
 export type Config = {
-  // Timestamp for when the build was started
-  buildTimestamp: number;
+  build: {
+    // Timestamp for when the build was started
+    timestamp: number;
+    // Whether to skip building the Next.js app or not
+    shouldSkip: boolean;
```
vicb, line 18: what about renaming `shouldSkip` to `skipNextBuild` ?

packages/cloudflare/src/cli/config.ts
```diff
@@ -46,13 +50,14 @@ export type Config = {
 /**
  * Computes the configuration.
  *
- * @param appDir Next app root folder
- * @param outputDir Output of the cloudflare builder
+ * @param opts.sourceDir Next app root folder
```
vicb, line 53: nit: move the comments on the type below

vicb (changes requested):
Thanks for the PR, I added a few minor comments.

For the next time please do me a favor and split changes in multiple commits. For example the first commit should have been the config refactoring and the second the renaming which would be trivial to review. It would be much faster to review.

james-elicx:
> For the next time please do me a favor and split changes in multiple commits. For example the first commit should have been the config refactoring and the second the renaming which would be trivial to review. It would be much faster to review.

Ah sorry about that, totally fair point - will do in the future.

vicb (approved):
Thanks!

PR #89, opened 2024-10-10
feat: cli arg to disable minification

james-elicx:
Introduces a CLI arg to let people choose to disable minification (e.g. for debugging purposes).

Only applies to the cache handler at the moment as that's the only part being minified currently.

### DIFF
--- a/packages/cloudflare/src/cli/args.ts
+++ b/packages/cloudflare/src/cli/args.ts
@@ -5,9 +5,10 @@ import { resolve } from "node:path";
 export function getArgs(): {
   skipBuild: boolean;
   outputDir?: string;
+  disableMinification?: boolean;
 } {
   const {
-    values: { skipBuild, output },
+    values: { skipBuild, output, disableMinification },
   } = parseArgs({
     options: {
       skipBuild: {
@@ -18,6 +19,9 @@ export function getArgs(): {
       output: {
         type: "string",
         short: "o",
+      },
+      disableMinification: {
+        type: "boolean",
       },
     },
     allowPositionals: false,
@@ -32,6 +36,7 @@ export function getArgs(): {
   return {
     outputDir,
     skipBuild: skipBuild || ["1", "true", "yes"].includes(String(process.env.SKIP_NEXT_APP_BUILD)),
+    disableMinification,
   };
 }
 
--- a/packages/cloudflare/src/cli/args.ts
+++ b/packages/cloudflare/src/cli/args.ts
@@ -3,9 +3,9 @@ import { parseArgs } from "node:util";
 import { resolve } from "node:path";
 
 export function getArgs(): {
-  skipBuild: boolean;
+  skipNextBuild: boolean;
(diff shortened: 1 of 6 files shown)


packages/cloudflare/src/cli/args.ts
```diff
@@ -5,9 +5,10 @@ import { resolve } from "node:path";
 export function getArgs(): {
   skipBuild: boolean;
   outputDir?: string;
+  disableMinification?: boolean;
```
vicb, line 8: I think we should prefer positive over negative for variable names in general. i.e. `if (!disable<>) {}` is harder to parse (at least for me).

It's a common pattern for CLI args to have `--<options>` and `--no-<options>` too.

Finally it would make this consistent with the rest of the code (`minify` and `shouldMinify`).

vicb (changes requested):
Nice change, see the inline comment

packages/cloudflare/src/cli/args.ts
james-elicx: i was thinking about that as well. in next-on-pages we had `disableWorkerMinfication` which can be a bit of a mouthful... 😅

packages/cloudflare/src/cli/args.ts
```diff
@@ -5,9 +5,10 @@ import { resolve } from "node:path";
 export function getArgs(): {
   skipBuild: boolean;
   outputDir?: string;
+  noMinify?: boolean;
```
vicb, line 8: I meant `minify` here

packages/cloudflare/src/cli/args.ts
```diff
@@ -5,7 +5,7 @@ import { resolve } from "node:path";
 export function getArgs(): {
   skipBuild: boolean;
```
vicb, line 6: nit:
- skipBuild -> skipNextBuild
- make minify mandatory

packages/cloudflare/src/cli/config.ts
```diff
@@ -66,7 +66,7 @@ export function getConfig(projectOpts: ProjectOptions): Config {
     build: {
       timestamp: Date.now(),
       skipNextBuild: !!projectOpts.skipBuild,
```
vicb, line 68: and here you can remove the !! * 2

packages/cloudflare/src/cli/config.ts
```diff
@@ -104,8 +104,8 @@ export type ProjectOptions = {
   outputDir: string;
   // Whether the Next.js build should be skipped (i.e. if the `.next` dir is already built)
   skipBuild?: boolean;
```
vicb, line 106: skipNextBuild, make mandatory * 2

vicb:
Thanks a lot for updating the code @james-elicx.

Sorry to bother you with nit-picks but I think it's goos to start to improve things little by little.

james-elicx:
> Sorry to bother you with nit-picks

It's no worries at all, I don't mind it

PR #98, opened 2024-10-12
refactor: use ALS for `process.env` object

james-elicx:
**Context**
process.env is currently assigned via spreading env when there is no request handler. The env values available should be scoped to a request, and dealt with on a per-request basis.

**Changes**
- Pulls out ALS proxy creation into a utility.
- Assigns a proxy for an ALS to process.
- Runs the worker inside the ALS with the worker env object spread on the ALS.

I recommend hiding whitespace changes when looking at the diff in this pr.

### DIFF
--- a/packages/cloudflare/src/cli/templates/utils/create-als-proxy.ts
+++ b/packages/cloudflare/src/cli/templates/utils/create-als-proxy.ts
@@ -5,11 +5,14 @@ import type { AsyncLocalStorage } from "node:async_hooks";
  *
  * @param als AsyncLocalStorage instance
  */
-export function createALSProxy<T extends object>(als: AsyncLocalStorage<T>) {
-  return new Proxy({} as T, {
-    ownKeys: () => Reflect.ownKeys(als.getStore()!),
-    getOwnPropertyDescriptor: (_, ...args) => Reflect.getOwnPropertyDescriptor(als.getStore()!, ...args),
-    get: (_, property) => Reflect.get(als.getStore()!, property),
-    set: (_, property, value) => Reflect.set(als.getStore()!, property, value),
-  });
+export function createALSProxy<T>(als: AsyncLocalStorage<T>) {
+  return new Proxy(
+    {},
+    {
+      ownKeys: () => Reflect.ownKeys(als.getStore()!),
+      getOwnPropertyDescriptor: (_, ...args) => Reflect.getOwnPropertyDescriptor(als.getStore()!, ...args),
+      get: (_, property) => Reflect.get(als.getStore()!, property),
+      set: (_, property, value) => Reflect.set(als.getStore()!, property, value),
+    }
+  );
 }
--- a/packages/cloudflare/src/cli/templates/utils/create-als-proxy.ts
+++ b/packages/cloudflare/src/cli/templates/utils/create-als-proxy.ts
@@ -1,9 +1,9 @@
 import type { AsyncLocalStorage } from "node:async_hooks";
 
 /**
- * Creates a proxy for to use for an instance of AsyncLocalStorage.
+ * Creates a proxy that exposes values from an AsyncLocalStorage store
  *
- * @param als AsyncLocalStorage instance.
+ * @param als AsyncLocalStorage instance
  */
 export function createALSProxy<T>(als: AsyncLocalStorage<T>) {
   return new Proxy(
--- a/packages/cloudflare/src/cli/templates/utils/create-als-proxy.ts
+++ b/packages/cloudflare/src/cli/templates/utils/create-als-proxy.ts
@@ -5,14 +5,11 @@ import type { AsyncLocalStorage } from "node:async_hooks";
  *
  * @param als AsyncLocalStorage instance
  */
-export function createALSProxy<T>(als: AsyncLocalStorage<T>) {
-  return new Proxy(
-    {},
-    {
-      ownKeys: () => Reflect.ownKeys(als.getStore()!),
-      getOwnPropertyDescriptor: (_, ...args) => Reflect.getOwnPropertyDescriptor(als.getStore()!, ...args),
-      get: (_, property) => Reflect.get(als.getStore()!, property),
-      set: (_, property, value) => Reflect.set(als.getStore()!, property, value),
-    }
-  );
+export function createALSProxy<T extends object>(als: AsyncLocalStorage<T>) {
+  return new Proxy({} as T, {
+    ownKeys: () => Reflect.ownKeys(als.getStore()!),
+    getOwnPropertyDescriptor: (_, ...args) => Reflect.getOwnPropertyDescriptor(als.getStore()!, ...args),
+    get: (_, property) => Reflect.get(als.getStore()!, property),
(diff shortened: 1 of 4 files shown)


packages/cloudflare/src/cli/templates/worker.ts
```diff
@@ -11,53 +11,61 @@ import { createALSProxy } from "./utils";
...
 const cloudflareContextALS = new AsyncLocalStorage<CloudflareContext>();
 
 // Note: this symbol needs to be kept in sync with the one defined in `src/api/get-cloudflare-context.ts`
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 (globalThis as any)[Symbol.for("__cloudflare-context__")] = createALSProxy(cloudflareContextALS);
 
+globalThis.process = {
+  ...globalThis.process,
+  // @ts-expect-error - populated when we run inside the ALS context
+  env: createALSProxy(processEnvALS),
```
vicb, line 24: @petebacondarwin @dario-piotrowicz is that ok? I think we do not want to replace the env, right?

packages/cloudflare/src/cli/templates/utils/create-als-proxy.ts
```diff
@@ -0,0 +1,18 @@
+import type { AsyncLocalStorage } from "node:async_hooks";
+
+/**
+ * Creates a proxy for to use for an instance of AsyncLocalStorage.
```
dario-piotrowicz, line 4:
```suggestion
 * Creates a proxy that exposes values from an AsyncLocalStorage store
```

packages/cloudflare/src/cli/templates/utils/create-als-proxy.ts
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * @param als AsyncLocalStorage instance.
```
dario-piotrowicz, line 6:
```suggestion
 * @param als AsyncLocalStorage instance
```

packages/cloudflare/src/cli/templates/utils/create-als-proxy.ts
```diff
@@ -0,0 +1,18 @@
...
+ */
+export function createALSProxy<T>(als: AsyncLocalStorage<T>) {
```
dario-piotrowicz, line 8: nice small utility! 😄

packages/cloudflare/src/cli/templates/worker.ts
dario-piotrowicz: I think this is valid, what we currently do is:
```ts
globalThis.process.env = { ...globalThis.process.env, ...env };
```

which just drops everything onto `globalThis.process` potentially causing leaks, so I think that using ALS instead is the right approach

Although I'd argue, why using ALS only for `env` and not the whole `process` object?

(As we do [in next-on-pages](https://github.com/cloudflare/next-on-pages/blob/96c331d20e679ebad0e43153430786685dbbcdec/packages/next-on-pages/src/buildApplication/generateGlobalJs.ts#L20-L30))

james-elicx: Happy to change it if you'd prefer it for for the entire process object, I don't mind either way

dario-piotrowicz: Honestly I don't think that it will make a significant difference...

But on the other side of things, it's also a trivial change to make and makes things potentially more robust without any downside no?

Do you think there's any benefit in just doing this for the `env`?

james-elicx: i dont think there's any benefit to one over the other, hence why i dont mind whichever you'd prefer :)

dario-piotrowicz: I feel like there's a veeeery veeeery slight benefit in wrapping the process object entirely, so yeah I would prefer going with that if you'd be ok with it 🙂

james-elicx: Actually it will cause a problem. If we do it on process instead of process.env, some apps where things may try to access parts of process.env which Cloudflare decides to implement, but we would have no way of making that available to them since we spread globalthis.process because, well, that is a proxy now.

dario-piotrowicz: Ah ok, sorry my bad 🙇 , thanks for giving it a try 🙂👍

dario-piotrowicz (approved):
Looks good to me, thanks for the improvement 🙂 

@james-elicx if you would could we wait for @petebacondarwin to maybe have a quick look tomorrow? Since @vicb seemed against the change I think it'd be great to have also Pete's point of view on this one 🙂

packages/cloudflare/src/cli/templates/worker.ts
```diff
@@ -9,23 +9,19 @@ import { MockedResponse } from "next/dist/server/lib/mock-request";
...
-    getOwnPropertyDescriptor: (_, ...args) =>
-      Reflect.getOwnPropertyDescriptor(cloudflareContextALS.getStore()!, ...args),
-    get: (_, property) => Reflect.get(cloudflareContextALS.getStore()!, property),
-    set: (_, property, value) => Reflect.set(cloudflareContextALS.getStore()!, property, value),
-  }
-);
+(globalThis as any)[Symbol.for("__cloudflare-context__")] = createALSProxy(cloudflareContextALS);
+
+// @ts-expect-error - populated when we run inside the ALS context
+globalThis.process.env = createALSProxy(processEnvALS);
```
petebacondarwin, line 24: I think I agree with @vicb to some extent here.
Before any request, there might already be values on `process.env`, which are being lost by assigning it here.
How about we capture the original `process.env` somehow and then include that in the initialization of the store below.

E.g.

```
const originalEnv = globalThis.process.env;
globalThis.process.env = createALSProxy(processEnvALS);
...
export default {
  async fetch(request, env, ctx) {
    return processEnvALS.run({ NODE_ENV: "production", ...originalEnv, ...env }, () => {
```

petebacondarwin (changes requested):

packages/cloudflare/src/cli/templates/worker.ts
```diff
@@ -20,6 +20,7 @@ const cloudflareContextALS = new AsyncLocalStorage<CloudflareContext>();
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 (globalThis as any)[Symbol.for("__cloudflare-context__")] = createALSProxy(cloudflareContextALS);
 
+const originalEnv: Partial<typeof process.env> = { ...globalThis.process.env };
```
petebacondarwin, line 23: Do we need to clone this object here? We make a copy of the `originalEnv` when setting up the ALS.

petebacondarwin (commented):
Minor nit - but could still do with a test or two...

vicb:
@james-elicx Do you think this PR is still needed?

james-elicx:
> @james-elicx Do you think this PR is still needed?

I don't think so considering process.env is no longer being populated with the request env vars.

PR #99, opened 2024-10-13
refactor: use join instead of path.join

james-elicx:
**Context**
We're inconsistent with how we use default imports with node:path vs every other import in the codebase.

**Changes**
- Pulls the path normalization out into a utility.
- Replaces the path default import with importing the actual stuff that's used.


vicb (approved):
Thanks!

PR #100, opened 2024-10-13
refactor: improve separation of source / output paths in config

james-elicx:
**Context**
Source and output paths are currently mixed in the config with no clear distinction

**Changes**
- Restructures config types into separate objects for source paths vs output paths vs internal paths.

### DIFF
--- a/packages/cloudflare/src/cli/config.ts
+++ b/packages/cloudflare/src/cli/config.ts
@@ -64,7 +64,7 @@ export function getConfig(projectOpts: ProjectOptions): Config {
   const sourceDirDotNext = join(projectOpts.sourceDir, ".next");
 
   const dotNext = join(projectOpts.outputDir, ".next");
-  const appPath = getNextjsApplicationPath(sourceDirDotNext).replace(/\/$/, "");
+  const appPath = getNextjsApplicationPath(dotNext).replace(/\/$/, "");
   const standaloneRoot = join(dotNext, "standalone");
   const standaloneApp = join(standaloneRoot, appPath);
   const standaloneAppDotNext = join(standaloneApp, ".next");
--- a/packages/cloudflare/src/cli/config.ts
+++ b/packages/cloudflare/src/cli/config.ts
@@ -14,24 +14,36 @@ export type Config = {
   };
 
   paths: {
-    // Path to the next application
-    sourceDir: string;
-    // Path to the output folder
-    outputDir: string;
-    // Path to the app's `.next` directory (where `next build` saves the build output)
-    dotNext: string;
-    // Path to the application standalone root directory
-    standaloneRoot: string;
-    // Path to the application standalone directory (where `next build` saves the standalone app)
-    standaloneApp: string;
-    // Path to the `.next` directory specific to the standalone application
-    standaloneAppDotNext: string;
-    // Path to the `server` directory specific to the standalone application
-    standaloneAppServer: string;
-    // Package in the standalone node_modules
-    internalPackage: string;
-    // Templates in the package in the standalone node_modules
-    internalTemplates: string;
+    source: {
+      // Path to the next application
+      root: string;
+      // Path to the app's `.next` directory (where `next build` saves the build output)
+      dotNext: string;
+      // Path to the application standalone root directory
+      standaloneRoot: string;
+    };
+    output: {
+      // Path to the output directory
+      root: string;
+      // Path to the OpenNext static assets directory
+      assets: string;
+      // Path to the app's `.next` directory in the OpenNext output directory
+      dotNext: string;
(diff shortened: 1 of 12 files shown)


packages/cloudflare/src/cli/config.ts
```diff
@@ -49,8 +49,9 @@ export type Config = {
  * @returns The configuration, see `Config`
  */
 export function getConfig(projectOpts: ProjectOptions): Config {
+  const sourceDirDotNext = join(projectOpts.sourceDir, ".next");
   const dotNext = join(projectOpts.outputDir, ".next");
-  const appPath = getNextjsApplicationPath(dotNext).replace(/\/$/, "");
+  const appPath = getNextjsApplicationPath(sourceDirDotNext).replace(/\/$/, "");
```
vicb, line 67: The reason why we copy `.next` to the output is to keep the source `.next` unchanged.
That way we can skip building next during development and we are still sure to have a pristine `.next`.
Let's say you are updating a patch modifying files in the `.next` folder, you would have to build the next app every time you change it if it was working on the source `.next`

james-elicx: The CLI currently fails to build in skip-build mode if no .next folder exists, so i guess that needs fixing as well if it's expected to work entirely independently of the .next directory.

james-elicx: Would you be okay with this doing if the dotnext dir exists, use that, otherwise use the source dir?

vicb: The workflow is: first do a full build (create the `.next`) and then you can iterate without having to rebuild next.

> so i guess that needs fixing as well if it's expected to work entirely independently of the .next directory.

It's mostly for dev usage so I don't think this needs to be perfect. But sure we could still build in skip build if there is no `.next` folder. Honestly not a high priority - but fell free to improve that.

> Would you be okay with this doing if the dotnext dir exists, use that, otherwise use the source dir?

Not sure to understand what you want to do here?

vicb: > not a high priority

If you want to contribute things that will be needed soon and have a higher impact, you can take a look at the [project board](https://github.com/orgs/opennextjs/projects/1).

I am working on integrating this adapter with opennext-aws. I will hopefully have midlleware + server running by the end of the week.

Next we will work on integrating the cache and the image optimization. If you feel like looking at one of this task it would really help. We need to make those run on the cloudflare infra.

You can ping me on Discord if you have questions.

Thanks!

james-elicx: > Not sure to understand what you want to do here?

i was going to enable the config to be creation without depending on the output directory existing. I'll remove that change. This PR is just improving separation of source / output / internal paths in the config now. Can close this if you don't think that's valuable though.

> Next we will work on integrating the cache and the image optimization. If you feel like looking at one of this task it would really help. We need to make those run on the cloudflare infra.

I would quite like to help get some of the cache stuff moving but it's not clear how the merge with the aws adaptor code is supposed to happen - new repos? new monorepo? copy paste?

It talks about multiple possibilities being explored already with no information or context about those explorations or the outcomes, so unsure what the current state is

vicb: > I would quite like to help get some of the cache stuff moving but it's not clear how the merge with the aws adaptor code is supposed to happen - new repos? new monorepo? copy paste?

Great to hear that!

Ultimately we will probably want a monorepo - this is something we will have to discuss with all the contributors.

For now the idea would be to have a `build.ts` that is specific to cloudflare but can still make use of the code in `was`. I have started to submit PRs to `aws` so that it's easier to re-use the code their. I'll create a branch on this repo with the code using `aws` and `aws` will be a dependency.

The maintainer of `aws` are willing to merge PRs making the code easier to re-use. If some parts of the code are really cloudflare specific or harder to integrate, we can copy/update the code and think later what the best to reconcile the code is.

> It talks about multiple possibilities being explored already with no information or context about those explorations or the outcomes, so unsure what the current state is

We only quickly brainstormed with Dario, Pete and Igor about possible ways to implement this. There is nothing decided yet. I guess having a first working implementation would help evaluate of feasibility and pros/cons of other possible solutions.

As mentioned on the cache issue there multiple parts to adapt. If you start working on something please create a dedicated issue and link to the main issue.

I hope this helps,
Thanks

dario-piotrowicz (approved):
Looks good to me 🙂 

Especially since you reverted the change @vicb commented about, since that's not an issue here I think the PR shouldn't really be controversial at this point 🙂 

But we do have an issue there in the sense the CLI fails to build in skip-build mode if no `.next` folder exists. @james-elicx could you open an issue for that so we can keep track of it? 🙂

PR #101, opened 2024-10-13
fix: copying excess files to output directory

james-elicx:
**Context**
We currently copy everything in .next to the output directory when we only utilise .next/standalone.

**Changes**
- Copies the standalone directory instead of the .next directory.


vicb (commented):
We should move away from this copy when we integrate with open next.
Then I don't think there is a need to change that now.

PR #109, opened 2024-10-18
feat: Update ESLint config to fix imports/exports

vicb:
The previously used `sort-imports` can not fix the code.

Changed the plugins to ones that can auto-fix the code (i.e. on save, or via `lint:fix`). 
Getting an error without auto-fix was becoming frustrating.

### DIFF
--- a/packages/cloudflare/eslint.config.mjs
+++ b/packages/cloudflare/eslint.config.mjs
@@ -1,6 +1,8 @@
+import pluginJs from "@eslint/js";
+import importPlugin from "eslint-plugin-import";
+import simpleImportSort from "eslint-plugin-simple-import-sort";
 import eslintPluginUnicorn from "eslint-plugin-unicorn";
 import globals from "globals";
-import pluginJs from "@eslint/js";
 import tseslint from "typescript-eslint";
 
 export default [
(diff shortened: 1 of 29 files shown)


petebacondarwin:
In workers-sdk and in the vite-plugin-cloudflare projects we use Prettier to do the import ordering.
See https://github.com/flarelabs-net/vite-plugin-cloudflare/pull/30/commits/1ad143fc19100534e473ac7873007a2e37d6bf55

This is nice because most people have formatting set up on-save in their IDEs, which keeps imports sorted all the time and avoids having to run eslint fix commands.

Could we consider that approach?

vicb:
> In workers-sdk and in the vite-plugin-cloudflare projects we use Prettier to do the import ordering. See [flarelabs-net/vite-plugin-cloudflare@1ad143f](https://github.com/flarelabs-net/vite-plugin-cloudflare/commit/1ad143fc19100534e473ac7873007a2e37d6bf55)
> 
> This is nice because most people have formatting set up on-save in their IDEs, which keeps imports sorted all the time and avoids having to run eslint fix commands.
> 
> Could we consider that approach?

Can we merge this PR to "unblock" me now and I'll create an issue for the follow up?

Issue #108, opened 2024-10-18
Tests


vicb:
Run:
- opennext tests
- Nextjs tests
- Different versions of Nextjs

dario-piotrowicz:
As part of this we should ideally also add tests for Next.js 15 and so understand to what extend we support it and its features

PR #112, opened 2024-10-29
Pin next to the catalog version

vicb:


dario-piotrowicz (commented):
What's the benefit of pinning `next` to the catalog?

If we were to have different `next` versions that would simply imply that we are testing our solution with different versions of the framework, which seems like a positive thing to me?

vicb:
You should definitely work on improving testing (there is an issue related to that).

But for now I don't want to have to update the apps (to match the latest `Next`) while I'm working on the open next integration.

vicb:
> these are also unpinned: [vercel-commerce/next-deps](https://github.com/opennextjs/opennextjs-cloudflare/blob/main/examples/vercel-commerce/package.json#L24-L26)

It's pinned (but not to the catalog version).

dario-piotrowicz:
> It's pinned (but not to the catalog version).

yeah sorry that's what I meant 😅

dario-piotrowicz (approved):
I don't fully agree with this change but I'm not super against it (since we're doing this in the `create-next-app` example)

vicb:
> I don't fully agree with this change but I'm not super against it (since we're doing this in the `create-next-app` example)

It's only temporary and we will improve tests as part of #108 

Thanks!

PR #113, opened 2024-10-30
Add middleware support

vicb:
It's missing e2e tests in the added middleware example but everything seems to be working fine 🎉


examples/api/open-next.config.ts
```diff
@@ -0,0 +1,25 @@
+import type { OpenNextConfig } from "open-next/types/open-next";
+
+const config: OpenNextConfig = {
+  default: {
+    override: {
+      wrapper: "cloudflare",
```
conico974, line 6: Is this default override necessary ? It seems that you're not using the default server at the moment

examples/api/open-next.config.ts
```diff
@@ -0,0 +1,25 @@
+import type { OpenNextConfig } from "open-next/types/open-next";
+
+const config: OpenNextConfig = {
```
conico974, line 3: It's probably a good idea to create some kind of function that could create the config object automatically and restrict what options could be set. 
I started implemented something like that here https://github.com/opennextjs/opennextjs-aws/blob/main/packages/open-next/src/helpers/withCloudflare.ts
This one won't work but something like that really specific to cloudflare

packages/cloudflare/src/cli/templates/worker.ts
```diff
@@ -35,6 +37,17 @@ let requestHandler: NodeRequestHandler | null = null;
...
+      // The Middleware handler can return either a `Response` or a `Request`:
+      // - `Response`s should be returned early
+      // - `Request`s are handled by the Next server
+      const reqOrResp = await middlewareHandler(request, env);
+
+      if (reqOrResp instanceof Response) {
+        return reqOrResp;
+      }
+
+      request = reqOrResp;
```
conico974, line 49: FYI headers from next-config and those set in the middleware (as response headers) won't work out of the box.
You'll need something like that to retrieve them https://github.com/opennextjs/opennextjs-aws/blob/bfa1a8c4056bd691fb57617dd6287693e51071b4/packages/open-next/src/core/requestHandler.ts#L54-L67 and they'll need to be merged to the response headers

packages/cloudflare/package.json
```diff
@@ -59,6 +59,7 @@
     "vitest": "catalog:"
   },
   "dependencies": {
+    "@opennextjs/aws": "https://pkg.pr.new/@opennextjs/aws@3b372b6",
```
conico974, line 62: I'll make a release tomorrow morning, you'll be able to use 3.2.1 instead of this

conico974 (commented):
Just a few question/comment.
Great job  anyway

examples/api/open-next.config.ts
vicb: Good catch, it was used by an other experiment and should be removed from here.

examples/api/open-next.config.ts
vicb: Definitely something to add to the list. 
For now ensureClouflareConfig checks the required bits (the name of from memory)

packages/cloudflare/src/cli/templates/worker.ts
vicb: That's one of the reason why this PR doesn't target main. I'll work on that next. 
My current plan is to merge back to main when the code uses that request handler and the ON Request and Response.
After that we will need to integrate the cache.

packages/cloudflare/package.json
vicb: Thanks! 
As explained above, a npm release isn't a req for us before we merge back to main.

packages/cloudflare/src/cli/templates/worker.ts
conico974: Ho right, i didn't see that it was not targeting main

packages/cloudflare/src/cli/build/index.ts
```diff
@@ -15,26 +21,84 @@ import { buildWorker } from "./build-worker";
...
+ *
+ * @param config OpenNext configuration.
+ */
+function ensureCloudflareConfig(config: OpenNextConfig) {
+  const requirements = {
+    isExternal: config.middleware?.external == true,
+    useCloudflareWrapper: config.middleware?.override?.wrapper === "cloudflare",
+    useEdgeConverter: config.middleware?.override?.converter === "edge",
+    disableCacheInterception: config.dangerous?.enableCacheInterception !== true,
+  };
```
james-elicx, line 87: Wouldn't it make sense for these to be default values if a user doesn't specify them in their config? As in, they're choosing the Cloudflare adapter, so their intention would be to want a project built for Cloudflare, unless they specifically put other config options.

Looking at the code in the aws adapter, it looks like the default behaviour is to only build an edge config if the middleware.external option is there - I think it would make sense for the cloudflare adapter's default to be to always build the edge config instead of requiring people add the extra config options, and then make sure that their config doesn't try to use an option that goes against the requirements for Cloudflare.

What do you think?

packages/cloudflare/README.md
```diff
@@ -2,8 +2,139 @@
 
 Deploy Next.js apps to Cloudflare!
 
-[OpenNext for Cloudflare](https://opennext.js.org/cloudflare) is an adapter that enables the deployment of Next.js applications to Cloudflare's developer platform.
+[OpenNext for Cloudflare](https://opennext.js.org/cloudflare) is Cloudflare specific adapter that enables deployment of Next.js applications to Cloudflare.
 
 ## Get started
 
-Visit the [OpenNext docs](https://opennext.js.org/cloudflare/get-started) for instructions on starting a new project, or migrating an existing one.
```
james-elicx, line 9: I think the content in this section might have been added back by accident 😅

examples/middleware/e2e/base.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+test("rewrite", async ({ page }) => {
+  await page.goto("/");
+  await page.click('[href="/another"]');
+  expect(page.waitForURL("**/another"));
+  expect(await page.textContent("h1")).toContain("Rewrite");
+});
+
+test("noop middleware", async ({ page }) => {
+  await page.goto("/");
+  await page.click('[href="/about2"]');
```
james-elicx, line 19: My assumption based on the matchers is that this shouldn't match the middleware function - is it worth adding one that also matches the NextResponse.next() part too?

packages/cloudflare/src/cli/build/index.ts
james-elicx: Reading conico's comment again, this sounds like a similar point to what #114 is for, so marking as resolved.

examples/middleware/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "middleware",
+  "private": true,
+  "scripts": {
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start",
+    "lint": "next lint",
+    "build:worker": "pnpm cloudflare",
+    "dev:worker": "wrangler dev --port 8770 --inspector-port 9330",
```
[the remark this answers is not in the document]
vicb: Can you please create an issue for that?
There should be no dependency here.

packages/cloudflare/README.md
vicb: It's not an accident. 
Having the doc in the same repo & branch makes it easier to update it and to learn about the specifics of that branch.

examples/middleware/e2e/base.spec.ts
vicb: > My assumption based on the matchers is that this shouldn't match the middleware function

That's a correct assumption. This test is here to verify that assumption.

> is it worth adding one that also matches the NextResponse.next() part too?

That should be the previous test (/another)

I think this can be marked as resolved unless I miss something

james-elicx: `/another` returns the rewrite so i don't think it would cover the `NextResponse.next()` bit. I can't remember if rewrites would invoke middleware again because it's been a while since i last did a rewrite in a middleware, but the rewritten path doesn't match the matcher so even if it did, i dont think it's covered

vicb: I double checked middleware.ts and you are absolutely right. We could add a /passthrough route and test it.

examples/create-next-app/tsconfig.json
```diff
@@ -22,5 +22,5 @@
     }
   },
   "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", ".next/types/**/*.ts"],
-  "exclude": ["node_modules"]
+  "exclude": ["node_modules", "open-next.config.ts"]
```
dario-piotrowicz, line 25: is this intentional? why? 😕

dario-piotrowicz: (the same question applies to all the other apps too)

examples/middleware/package.json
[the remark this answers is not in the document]
dario-piotrowicz: > The other examples incremented the number to prevent two using the same, so I assume we want to do the same with new ones as well?

Yes, that's because we run the e2e tests in parallel and thus we don't want dev server to clash 👍

packages/cloudflare/src/cli/build/build-next-app.ts
dario-piotrowicz, line 1: 🚀

packages/cloudflare/src/cli/build/build-worker.ts
```diff
@@ -134,6 +135,7 @@ const CustomRequest = class extends globalThis.Request {
 };
 globalThis.Request = CustomRequest;
 Request = globalThis.Request;
+globalThis.__dangerous_ON_edge_converter_returns_request = true;
```
dario-piotrowicz, line 138: what is this for? should we document it with a comment or something?

examples/create-next-app/open-next.config.ts
```diff
@@ -0,0 +1,15 @@
+import type { OpenNextConfig } from "@opennextjs/aws/types/open-next";
```
dario-piotrowicz, line 1: here and in the other examples I get this:
Screenshot 2024-11-04 at 10 17 21

To solve it I've installed `@opennextjs/aws` as a dev dependency of the app
and also updated the import to:
```suggestion
import type { OpenNextConfig } from "@opennextjs/aws/types/open-next.js";
```

packages/cloudflare/src/cli/templates/worker.ts
```diff
@@ -2,14 +2,16 @@ import { AsyncLocalStorage } from "node:async_hooks";
...
-import type { ExportedHandler, Fetcher } from "@cloudflare/workers-types";
 import type { NextConfig } from "next";
 import { NodeNextRequest, NodeNextResponse } from "next/dist/server/base-http/node";
 import { MockedResponse } from "next/dist/server/lib/mock-request";
 import type { NodeRequestHandler } from "next/dist/server/next-server";
 
 import type { CloudflareContext } from "../../api";
 
+// @ts-expect-error: resolved by wrangler build
+import { handler as middlewareHandler } from "./middleware/handler.mjs";
```
dario-piotrowicz, line 13: Instead of a `@ts-expect-error` I think it would be nicer to define a proper `d.ts` file for this (so that we can have proper type safety)

Something like this for example:
Screenshot 2024-11-04 at 10 31 33

(although I don't really know if the type I'm using there is the correct one, or even if the type we want is actually exported, but you get the idea 🙂)

examples/middleware/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+```bash
+yarn create next-app --example middleware middleware-app
+```
+
+```bash
+pnpm create next-app --example middleware middleware-app
+```
+
+Deploy it to the cloud with [Vercel](https://vercel.com/new?utm_source=github&utm_medium=readme&utm_campaign=next-example) ([Documentation](https://nextjs.org/docs/deployment)).
```
dario-piotrowicz, line 31:
```suggestion
```

examples/middleware/e2e/base.spec.ts
vicb: added a /middleware route + test

examples/create-next-app/tsconfig.json
vicb: Looks like you figured out: `import type { OpenNextConfig } from "@opennextjs/aws/types/open-next";` does not compile because `@opennextjs/aws` is not installed.

Hopefully that will not be needed soon when we share code with aws.

packages/cloudflare/src/cli/build/build-worker.ts
vicb: The ona [middleware handler](https://github.com/opennextjs/opennextjs-aws/blob/f7758ac91d29e3cedb86301ecc6ab4e7af581024/packages/open-next/src/adapters/middleware.ts) takes an `InternalEvent` and returns an `InternalResult`.

Cloudflare handlers takes a `Request` and return a `Response`.

To adapt those types, Open Next uses a converter. By default [the edge converter](https://github.com/opennextjs/opennextjs-aws/blob/f7758ac91d29e3cedb86301ecc6ab4e7af581024/packages/open-next/src/overrides/converters/edge.ts) will convert the `Request` to an `InternalEvent` (`convertFrom` is called before the middleware handler) and once the middleware handler returns, `convertTo` is called to convert the `InternalResult` to a `Response`.

In the current cloudflare implementation we do not always want to return a `Response`. We only want want to `Response` when the middleware generates one (redirects or plain response). In all others cases we will send the `Request` to the next server.

`globalThis.__dangerous_ON_edge_converter_returns_request` instructs the edge converter to returns the request before trying to fetch the response from an external next server - our next server is local.

examples/create-next-app/open-next.config.ts
vicb: Touched based in a previous comment: this will ultimately goes away when we actually merge with aws

packages/cloudflare/src/cli/templates/worker.ts
vicb: > so that we can have proper type safety

This is **not** type safety but rather akin to a cast, so I'm -1 on integrating this.

`@ts-expect-error` is also what ona uses and I like how this makes things explicit.

examples/middleware/README.md
vicb: What are the changes here?

packages/cloudflare/src/cli/templates/worker.ts
dario-piotrowicz: > This is not type safety but rather akin to a cast, so I'm -1 on integrating this.

Yes I agree that it is a cast, but if we did have the correct type it would introduce type safety to the code using the handler

The `@ts-expect-error` makes it so that `middlewareHandler` is basically an `any` which I think is clearly wrong in terms of type safety

examples/middleware/package.json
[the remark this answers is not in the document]
dario-piotrowicz: yeah I agree, unfortunately that would require us to do some orchestration when running the tests... maybe it's worth investigating into it 👍

examples/create-next-app/tsconfig.json
dario-piotrowicz: ahhhh ok I see 😕

dario-piotrowicz (approved):
Looks good to me 🙂 

It's very cool and promising seeing the integration with ona 🤩 🚀

packages/cloudflare/src/cli/templates/worker.ts
vicb: > but if we did have the correct type

Which is the exact problem with a cast: by overriding the type there is no way to make sure you have the correct type.

This is "internal" code so I think `@ts-expect-error` is a reasonable thing to do here. If there are breaking changes in the handler the open next build would and it will be caught by the tests _(whether or not there is `.d.ts`)_.

vicb:
Thanks for the reviews and the approval @dario-piotrowicz.

As you noted there are still a few rough edges but they will be solved as the integration progresses - and I think those are ok in the `experimental` branch.

My plan is to merge this PR and then push the changes to the experimental branch so that it's easier for users [to install](https://github.com/opennextjs/opennextjs-cloudflare/pull/115) and test the package.

PR #136, opened 2024-11-17
feat: use the cache api if there is no kv cache available

### LINKED ISSUE #134: [FEATURE] Support Cloudflare Cache API as alternative to KV

Currently @opennextjs/cloudflare uses [Workers KV](https://developers.cloudflare.com/kv/) as the cache for your Next.js app.

Some of our use cases have very high volumes of requests that we would like to cache, but it's not at all cost effective to Write/Read constantly to KV.  

For these use cases, we don't need global-distributed cache, as local colo-specific cache is fine, so we would simply prefer if we could configure the Next Cache to use the Cloudflare Workers Cache API.

As mentioned, it appears the only 2 alternatives today are to either disable caching or use KV for Next caching - We would just really appreciate an option to use the Cloudflare Cache API

0.2.1

### Additional context

_No response_

### Before submitting

- [X] I have checked that there isn't already a similar feature request
- [X] This is a single feature (not multiple features in one request)

james-elicx:
**Changes**

- Adds an abstraction for cache stores that the cache handler interact with.
- Moves the KV interaction to a new store.
- Adds a new Cache API store.
- Defaults to the Cache API store if there is no KV binding available.

resolves #134

### DIFF
--- a/packages/cloudflare/src/cli/templates/cache-handler/cache-store.ts
+++ b/packages/cloudflare/src/cli/templates/cache-handler/cache-store.ts
@@ -1,0 +1,32 @@
+import type { IncrementalCacheValue } from "next/dist/server/response-cache";
+
+export type CacheEntry = {
+  lastModified: number;
+  value: IncrementalCacheValue | null;
+};
+
+export type CacheStore = {
+  get: (key: string) => Promise<CacheEntry | null>;
+  put: (key: string, entry: CacheEntry, ttl?: number) => Promise<void>;
+};
+
+export function getCacheStore() {
+  const kvName = process.env.__OPENNEXT_KV_BINDING_NAME;
+  if (kvName && process.env[kvName]) {
+    return new KVStore(process.env[kvName] as unknown as KVNamespace);
+  }
+}
+
+const defaultTTL = 31536000; // 1 year
+
+class KVStore implements CacheStore {
+  constructor(private store: KVNamespace) {}
+
+  get(key: string) {
+    return this.store.get<CacheEntry>(key, "json");
+  }
+
+  put(key: string, entry: CacheEntry, ttl = defaultTTL) {
+    return this.store.put(key, JSON.stringify(entry), { expirationTtl: ttl });
+  }
+}
--- a/packages/cloudflare/src/cli/templates/cache-handler/cache-store.ts
+++ b/packages/cloudflare/src/cli/templates/cache-handler/cache-store.ts
@@ -12,9 +12,11 @@ export type CacheStore = {
 
 export function getCacheStore() {
   const kvName = process.env.__OPENNEXT_KV_BINDING_NAME;
-  if (kvName && process.env[kvName]) {
+  if (process.env[kvName]) {
     return new KVStore(process.env[kvName] as unknown as KVNamespace);
   }
+
+  return new CacheAPIStore();
 }
 
 const defaultTTL = 31536000; // 1 year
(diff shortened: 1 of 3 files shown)


packages/cloudflare/src/cli/templates/cache-handler/cache-store.ts
```diff
@@ -0,0 +1,65 @@
...
+};
+
+export type CacheStore = {
+  get: (key: string) => Promise<CacheEntry | null>;
+  put: (key: string, entry: CacheEntry, ttl?: number) => Promise<void>;
+};
+
+export function getCacheStore() {
+  const kvName = process.env.__OPENNEXT_KV_BINDING_NAME;
+  if (process.env[kvName]) {
```
vicb, line 15: I think we should remove the bindings from `process.env` and we should move to `getCloudflareContext` - could be done in a follow up PR, in which case it would be nice to add a `//TODO: ...`

packages/cloudflare/src/cli/templates/cache-handler/cache-store.ts
```diff
@@ -0,0 +1,65 @@
...
+    return new KVStore(process.env[kvName] as unknown as KVNamespace);
+  }
+
+  return new CacheAPIStore();
+}
+
+const defaultTTL = 31536000; // 1 year
```
vicb, line 22: Can we rename that to `oneYearInMs` to be explicit (we can add a comment saying it's the default cache TTL)

packages/cloudflare/src/cli/templates/cache-handler/cache-store.ts
```diff
@@ -0,0 +1,65 @@
+import type { IncrementalCacheValue } from "next/dist/server/response-cache";
+
+export type CacheEntry = {
+  lastModified: number;
+  value: IncrementalCacheValue | null;
+};
+
+export type CacheStore = {
+  get: (key: string) => Promise<CacheEntry | null>;
```
vicb, line 9: async?

packages/cloudflare/src/cli/templates/cache-handler/cache-store.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  put(key: string, entry: CacheEntry, ttl = defaultTTL) {
+    return this.store.put(key, JSON.stringify(entry), {
+      expirationTtl: ttl,
+    });
+  }
+}
+
+class CacheAPIStore implements CacheStore {
```
vicb, line 38: @petebacondarwin @dario-piotrowicz I'm not familiar the cf cache API, do you have thoughts about this class?

With my limited understanding of the cache API, I'm not sure if we should use it for the fetch cache (as long as the cache headers are set correctly on the `fetch` call).

I'm also wondering about [custom keys](Currently, it is not possible to purge a URL stored through Cache API that uses a custom cache key set by a Worker).

packages/cloudflare/src/cli/templates/cache-handler/cache-store.ts
james-elicx: They are - the return type is a promise

packages/cloudflare/src/cli/templates/cache-handler/cache-store.ts
vicb: yep, you can create a TODO/issue for that. thanks.

james-elicx: #137

PR #140, opened 2024-11-22
feat: integrate the OpenNext server

vicb:
@dario-piotrowicz @james-elicx @petebacondarwin Here is what I have been working on.
/cc @conico974 - I will never thank you enough for all the support!

disclaimers: 
- it's a preview on an "experimental" branch :)
- I have updated the README

I have only updated the `middleware` example, this is how I test it (from `examples/middleware`):

```bash
OPEN_NEXT_DEBUG=1 node /path/to/opennextjs-cloudflare/packages/cloudflare/dist/cli/index.mjs
wrangler dev
```

You might see errors related to the cache - it is expected ([this PR](https://github.com/opennextjs/opennextjs-aws/pull/641) has more info)

There is still some cleanup to do:
- Added "TODO" from this PR - more generally drop legacy stuff and migrate all to ON
- https://github.com/opennextjs/opennextjs-aws/pull/641 mentioned above
- Update all the examples and make sure tests pass
- Drop tsup and use tsc directly (as in aws)
- Figure out how to handle ON ALS override
- Probably some more that I can't remember right now

Then we should work on the cache implementation.


packages/cloudflare/package.json
```diff
@@ -59,7 +59,7 @@
     "vitest": "catalog:"
   },
   "dependencies": {
-    "@opennextjs/aws": "https://pkg.pr.new/@opennextjs/aws@5c0e121",
+    "@opennextjs/aws": "https://pkg.pr.new/@opennextjs/aws@642",
```
dario-piotrowicz, line 62: @vicb are you sure that this is the correct package?

it seems to 404 😕 

https://pkg.pr.new/@opennextjs/aws@642:
Screenshot 2024-11-22 at 10 42 23

dario-piotrowicz: (this is causing the prerelease of this PR to also not work)

dario-piotrowicz: @Aslemammad any idea? 🙂

conico974: It has been merged, maybe they delete it once merged. You could use https://pkg.pr.new/@opennextjs/aws@main instead

vicb: Oops, sorry. 
I'll update the PR later today.

vicb: Should be updated to https://pkg.pr.new/@opennextjs/aws@0ac604e

vicb: forced push the updated.

packages/cloudflare/src/cli/templates/cache-handler/open-next-cache-handler.ts
james-elicx, line 1: Just a side note that this is what I meant the other day when I was saying that it's hard to try and meaningfully contribute until the merge is done, as most of the things I've been looking at are getting removed as part of it.

Should I close #136 as well if our cache handler is being removed?

vicb: I thought about that exact thing when I deleted the handler.

I mostly removed it because the ON cache API is different.
However we will need to re-integrate the cache ("Then we should work on the cache implementation.").

While the API will is changing, it is similar.
I think we can keep #136 opened before we have an impl in the experimental branch.

packages/cloudflare/src/cli/build/patches/investigated/copy-package-cli-files.ts
```diff
@@ -1,15 +1,22 @@
...
-  const sourceDir = join(packageDistDir, "cli");
-  const destinationDir = join(config.paths.internal.package, "cli");
+  const sourceDir = path.join(packageDistDir, "cli");
+  const destinationDir = path.join(config.paths.internal.package, "cli");
+
+  fs.cpSync(sourceDir, destinationDir, { recursive: true });
 
-  cpSync(sourceDir, destinationDir, { recursive: true });
+  fs.copyFileSync(
+    path.join(packageDistDir, "cli", "templates", "worker.ts"),
```
petebacondarwin, line 19: NIT:
```suggestion
    path.join(packageDistDir, "cli/templates/worker.ts"),
```

petebacondarwin (approved):

packages/cloudflare/env.d.ts
```diff
@@ -6,9 +6,13 @@ declare global {
...
       NEXT_PRIVATE_DEBUG_CACHE?: string;
       __OPENNEXT_KV_BINDING_NAME: string;
-      [key: string]: string | Fetcher;
+      OPEN_NEXT_ORIGIN: string;
     }
   }
+
+  interface Window {
+    [key: string]: string | Fetcher;
+  }
```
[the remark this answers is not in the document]
vicb: I'm not sure... AI autocomplete??

packages/cloudflare/src/cli/build/bundle-server.ts
```diff
@@ -60,15 +61,15 @@ export async function buildWorker(config: Config): Promise<void> {
       // Note: we apply an empty shim to next/dist/compiled/ws because it generates two `eval`s:
       //   eval("require")("bufferutil");
       //   eval("require")("utf-8-validate");
-      "next/dist/compiled/ws": join(config.paths.internal.templates, "shims", "empty.ts"),
+      "next/dist/compiled/ws": path.join(config.paths.internal.templates, "shims", "empty.ts"),
```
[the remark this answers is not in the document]
vicb: I have no clue if this would work with win32?

packages/cloudflare/src/cli/build/open-next/createServerBundle.ts
```diff
@@ -0,0 +1,289 @@
...
+  // Note: in the monorepo case, the handler file is output to
+  //       `.next/standalone/package/path/index.mjs`, but we want
+  //       the Lambda function to be able to find the handler at
+  //       the root of the bundle. We will create a dummy `index.mjs`
+  //       that re-exports the real handler.
+
+  // Always use posix path for import path
+  const packagePosixPath = packagePath.split(path.sep).join(path.posix.sep);
+  fs.writeFileSync(path.join(outputPath, "index.mjs"), `export * from "./${packagePosixPath}/index.mjs";`);
+}
```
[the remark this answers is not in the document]
vicb: I'll add a TODO here - the code is copied over from aws, we should use the helper (I believe James added one) when we switch to a monorepo.

packages/cloudflare/src/cli/build/open-next/createServerBundle.ts
```diff
@@ -0,0 +1,289 @@
...
+  //       Next.js app.
+
+  const disableNextPrebundledReact =
+    buildHelper.compareSemver(options.nextVersion, "13.5.1") >= 0 ||
+    buildHelper.compareSemver(options.nextVersion, "13.4.1") <= 0;
+
+  const overrides = fnOptions.override ?? {};
+
+  const isBefore13413 = buildHelper.compareSemver(options.nextVersion, "13.4.13") <= 0;
+  const isAfter141 = buildHelper.compareSemver(options.nextVersion, "14.0.4") >= 0;
```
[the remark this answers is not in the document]
vicb: This should be modified in aws

packages/cloudflare/src/cli/build/open-next/createServerBundle.ts
```diff
@@ -0,0 +1,289 @@
...
+    fnOptions.routes ?? ["app/page.tsx"],
+    isBundled
+  );
+
+  // Build Lambda code
+  // note: bundle in OpenNext package b/c the adapter relies on the
+  //       "serverless-http" package which is not a dependency in user's
+  //       Next.js app.
+
+  const disableNextPrebundledReact =
```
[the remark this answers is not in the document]
vicb: I think so (the code comes from aws)

packages/cloudflare/src/cli/build/patches/investigated/copy-package-cli-files.ts
vicb: win32...

packages/cloudflare/src/cli/build/open-next/createServerBundle.ts
```diff
@@ -0,0 +1,289 @@
...
+  if (fs.existsSync(path.join(serverPath, "pages"))) {
+    const pagePath = path.join(serverPath, "pages");
+    buildHelper.traverseFiles(
+      pagePath,
+      ({ relativePath }) => relativePath.endsWith(".js"),
+      ({ relativePath }) => {
+        const route = `pages/${relativePath.replace(/\.js$/, "")}`;
+        if (!foundRoutes.has(route)) {
+          remainingRoutes.add(route);
+        }
```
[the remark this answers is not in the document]
vicb: It is just a Set.

Nope, those are **two** Sets.

packages/cloudflare/src/cli/build/open-next/createServerBundle.ts
```diff
@@ -0,0 +1,289 @@
...
+      }
+    );
+  }
+
+  // Generate default function
+  await generateBundle("default", options, {
+    ...defaultFn,
+    // @ts-expect-error - Those string are RouteTemplate
+    routes: Array.from(remainingRoutes),
```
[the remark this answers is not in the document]
vicb: This needs to be changed in aws

packages/cloudflare/src/cli/build/open-next/createServerBundle.ts
```diff
@@ -0,0 +1,289 @@
...
+
+  const promises = functions.map(async ([name, fnOptions]) => {
+    const routes = fnOptions.routes;
+    routes.forEach((route) => foundRoutes.add(route));
+    if (fnOptions.runtime === "edge") {
+      await generateEdgeBundle(name, options, fnOptions);
+    } else {
+      await generateBundle(name, options, fnOptions);
+    }
+  });
```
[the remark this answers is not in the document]
vicb: you can only await in async function, no?

packages/cloudflare/src/cli/build/bundle-server.ts
```diff
@@ -151,10 +161,15 @@ async function updateWorkerBundledCode(workerOutputFile: string, config: Config)
...
   patchedCode = patchFindDir(patchedCode, config);
   patchedCode = inlineEvalManifest(patchedCode, config);
-  patchedCode = await patchCache(patchedCode, config);
+  patchedCode = await patchCache(patchedCode, openNextOptions);
   patchedCode = inlineMiddlewareManifestRequire(patchedCode, config);
   patchedCode = patchExceptionBubbling(patchedCode);
 
+  patchedCode = patchedCode
+    // workers do not support dynamic require nor require.resolve
+    .replace("patchAsyncStorage();", "//patchAsyncStorage();")
```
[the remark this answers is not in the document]
vicb: patchAsyncStorage relies on require.resolve.
Added a TODO to implement later.

packages/cloudflare/src/cli/build/patches/investigated/copy-package-cli-files.ts
petebacondarwin: **`path.join()`:**

> The path.join() method joins all given path segments together using the platform-specific separator as a delimiter, then normalizes the resulting path.

https://nodejs.org/api/path.html#pathjoinpaths

**`path.normalize()`**

> When multiple, sequential path segment separation characters are found (e.g. / on POSIX and either \ or / on Windows), they are replaced by a single instance of the platform-specific path segment separator (/ on POSIX and \ on Windows). Trailing separators are preserved.

https://nodejs.org/api/path.html#path_path_normalize_path

petebacondarwin: So you can use `/` on Windows in paths but not the other way round.

petebacondarwin: I'm not what the concern is there, but I have used forward slashes in paths on many Node.js projects that run on Windows, and it has never been a problem.

Only a NIT. Don't feel obliged to do this. It just feels easier to read for me. Especially if you ended up with a path that had a large number of segments, where Prettier will "format" the args for you:

```
const p = path.join(
  "a",
  "b",
  "c",
  "d",
  "e",
  "f",
);
```

vicb:
```
Welcome to Node.js v22.11.0.
> p = require('path/win32')
> p.join('c:\\base\\path', 'to/file');
'c:\\base\\path\\to\\file'
```

I'll create an issue to follow up

vicb: https://github.com/opennextjs/opennextjs-cloudflare/issues/142

packages/cloudflare/src/cli/build/open-next/createServerBundle.ts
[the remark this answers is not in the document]
vicb: right.
Should be changed in aws

petebacondarwin (approved):

vicb:
Thanks for the review!

Issue #143, opened 2024-11-27
[BUG] Deployment Failure: k.resolve is not a function in experimental Branch During Cloudflare Workers Deployment


martcarrefour:
### Describe the bug

```
Total Upload: 13091.56 KiB / gzip: 2395.29 KiB
Your worker has access to the following bindings:
- KV Namespaces:
  - NEXT_CACHE_WORKERS_KV: c0c379d...
  
[ERROR] A request to the Cloudflare API (/accounts/***/workers/scripts/frontend-dev/versions) failed.

Uncaught TypeError: k.resolve is not a function
    at null.<anonymous> (file:///home/runner/work/frontend/frontend/.open-next/server-functions/default/handler.mjs:32942:12)
    [code: 10021]

```

### Steps to reproduce

1. Switch to the experimental branch.
2. Run the Cloudflare Workers deployment using OpenNext.
3. Observe the error in the deployment output.

### Expected behavior

The deployment should complete successfully.

### @opennextjs/cloudflare version

experimental

### Node.js version

22.11.0

### Wrangler version

3.87.0

### next info output

```shell
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Pro
  Available memory (MB): 49084
  Available CPU cores: 16
Binaries:
  Node: 22.11.0
  npm: N/A
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  next: 14.2.15 // An outdated version detected (latest is 15.0.3), upgrade is highly recommended!
  eslint-config-next: 14.2.15
  react: 18.3.1
  react-dom: 18.3.1
  typescript: 5.5.4
Next.js Config:
  output: N/A
 ⚠ An outdated version detected (latest is 15.0.3), upgrade is highly recommended!
   Please try the latest canary version (`npm install next@canary`) to confirm the issue still exists before creating a new issue.
   Read more - https://nextjs.org/docs/messages/opening-an-issue
```

### Additional context

**Location**:
- Branch: `experimental`
- Commit: `fe29fee` ([View Commit on GitHub](https://github.com/opennextjs/opennextjs-cloudflare/commit/fe29fee))
- File: `.open-next/server-functions/default/handler.mjs:32942:12`

vicb:
What app are you trying to deploy?

vicb:
Repro from a similar issue (posted by Tempo on ON Discord): https://github.com/tempo22/test-opennext-cloudflare

vicb:
[...]

PR #144, opened 2024-11-28
fix(build): Do not minify OpenNext bundles

vicb:
This is required to apply string replacement patches.
Fixes https://github.com/opennextjs/opennextjs-cloudflare/issues/143


packages/cloudflare/src/cli/build/bundle-server.ts
```diff
@@ -168,6 +167,7 @@ async function updateWorkerBundledCode(
 
   patchedCode = patchedCode
     // workers do not support dynamic require nor require.resolve
+    // TODO: implement for cf
```
conico974, line 170: Maybe add a note that this can be done in `@opennextjs/aws`, not necessarily here

conico974 (approved):
LGTM

vicb:
Thanks for the review!

Issue #161, opened 2024-12-10
rename the binary from "cloudflare" to "opennextjs-cloudflare" or similar


IgorMinar:
Currently the binary provided by the `@opennextjs/cloudflare` package is called `cloudflare`, which means we invoke the app builder as:

```json
    "deploy": "cloudflare && wrangler deploy",
    "preview": "cloudflare && wrangler dev",
```

This is due to how the package.json is set up:
https://github.com/opennextjs/opennextjs-cloudflare/blob/021ddb7ef630e0162b9fb55226b82ce6bfd13efc/packages/cloudflare/package.json#L14

We should come up with a less ambiguous name than `cloudflare`, for example `opennextjs-cloudflare` and update package.json as:

```json
    "bin": {
        "opennextjs-cloudflare": "dist/cli/index.mjs", 
    }
```

docs: https://docs.npmjs.com/cli/v10/configuring-npm/package-json#bin

This change will require additional follow-ups:
- [ ] update docs (README for now, we'll sync to ON docs when experimental is merged to main)
- [ ] update examples: https://github.com/opennextjs/opennextjs-cloudflare/tree/main/examples
- [ ] update create-cloudflare template: https://github.com/cloudflare/workers-sdk/blob/8937b01df9d39548cddbd8378d2b68ba5221b1fd/packages/create-cloudflare/templates-experimental/next/c3.ts#L43-L44

PR #163, opened 2024-12-11
feat: rename the binary from "cloudflare" to "opennextjs-cloudflare"

petebacondarwin:
**BREAKING CHANGE**:
After this change the old way of running the tool (e.g. `pnpm cloudflare`) no longer works.
Going forward use the new binary name (e.g. `pnpm opennextjs-cloudflare`).

See [#161](https://github.com/opennextjs/opennextjs-cloudflare/issues/161)

### DIFF
--- a/.changeset/happy-worms-rest.md
+++ b/.changeset/happy-worms-rest.md
@@ -1,0 +1,11 @@
+---
+"@opennextjs/cloudflare": minor
+---
+
+feat: rename the binary from "cloudflare" to "opennextjs-cloudflare"
+
+**BREAKING CHANGE**:
+After this change the old way of running the tool (e.g. `pnpm cloudflare`) no longer works.
+Going forward use the new binary name (e.g. `pnpm opennextjs-cloudflare`).
+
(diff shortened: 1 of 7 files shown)


vicb (changes requested):
LGTM

could you please update https://github.com/opennextjs/opennextjs-cloudflare/blob/experimental/packages/cloudflare/README.md so that we have a single source of truth when we update the official doc as experimental is merged back to main.

Thanks!

petebacondarwin:
Ah! I forgot most of the work is going on in experimental. It feels like we should just merge that branch into main, no?

petebacondarwin:
I'll make a second PR against `experimental`.

vicb:
> Ah! I forgot most of the work is going on in experimental. It feels like we should just merge that branch into main, no?

I think that experimental is ok (I approved that other PR).
It should be merged back to main shortly and there will be other required changes (wrangler.toml, open next config, ...) so maybe do all the breaking change at once?

vicb:
#166 was merged to experimental

PR #141, opened 2024-11-22
[WIP] add patches to support nodejs.org

dario-piotrowicz:
I'm trying to see if we could run the nodejs.org website using our adaptor, as part of that I am opening this PR that introduces some patches that seem necessary to support the nodejs.org website.

I'm opening this PR now mainly to get a pkg.pr.new package so that I can share my progress with others.

(but the changes here might be generally needed/valid)

See #160

PR #169, opened 2024-12-12
temporary patch the `loadInstrumentationModule` methods to support Next.js 15

### LINKED ISSUE #160: Instrumentation with Next15
Loading instrumentation fails with Next15.

dario-piotrowicz:
resolves #160 

___

This PR patches the bundled worker code by removing the content of all the problematic `loadInstrumentationModule` methods.

Such removal however most likely breaks [Next.js' instrumentation](https://nextjs.org/docs/app/building-your-application/optimizing/instrumentation), so I think that it should just be considered a temporary workaround for supporting Next.js 15.

### DIFF
--- a/packages/cloudflare/src/cli/build/bundle-server.ts
+++ b/packages/cloudflare/src/cli/build/bundle-server.ts
@@ -5,7 +5,6 @@ import { fileURLToPath } from "node:url";
 
 import type { BuildOptions } from "@opennextjs/aws/build/helper.js";
 import { build, Plugin } from "esbuild";
-import * as ts from "ts-morph";
 
 import { Config } from "../config";
 import { copyPackageCliFiles } from "./patches/investigated/copy-package-cli-files";
@@ -17,9 +16,10 @@ import { inlineMiddlewareManifestRequire } from "./patches/to-investigate/inline
 import { inlineNextRequire } from "./patches/to-investigate/inline-next-require";
 import { patchExceptionBubbling } from "./patches/to-investigate/patch-exception-bubbling";
 import { patchFindDir } from "./patches/to-investigate/patch-find-dir";
+import { patchLoadInstrumentationModule } from "./patches/to-investigate/patch-load-instrumentation-module";
 import { patchReadFile } from "./patches/to-investigate/patch-read-file";
 import { patchWranglerDeps } from "./patches/to-investigate/wrangler-deps";
-import { copyPrerenderedRoutes, tsParseFile } from "./utils";
+import { copyPrerenderedRoutes } from "./utils";
 
 /** The dist directory of the Cloudflare adapter package */
 const packageDistDir = path.join(path.dirname(fileURLToPath(import.meta.url)), "..");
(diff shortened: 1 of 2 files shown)


dario-piotrowicz:
> [!Note]
> I wanted to use ts-morph to make the replacement more robust.... only after applying it I actually looked into the instrumentation docs and understood that this is a temporary workaround 😓, so the regex replacement I had in #141 was actually good enough 😓, I am happy to revert to that if people prefer that for such a temporary workaround

dario-piotrowicz:
> [!Note]
> I think this change is valuable for unblocking Next.js 15, I will still look into what a potential proper fix for #160 should look like

packages/cloudflare/src/cli/build/bundle-server.ts
```diff
@@ -164,6 +200,7 @@ async function updateWorkerBundledCode(
   patchedCode = await patchCache(patchedCode, openNextOptions);
   patchedCode = inlineMiddlewareManifestRequire(patchedCode, config);
   patchedCode = patchExceptionBubbling(patchedCode);
+  patchedCode = patchLoadInstrumentationModule(tsParseFile(patchedCode)).print();
```
dario-piotrowicz, line 203: the `tsParseFile()` + `print()` looks a bit out of place here, but I think this is the appropriate direction actually, ideally we should call `tsParseFile()` at line 193, and use that `ts.SourceFile` for all the patching, and end with a single `print()`

Before we get to that, in my opinion, the more patches we can do here with AST the better.

(if someone disagree I am happy to keep going with the regex replacement)

packages/cloudflare/src/cli/build/bundle-server.ts
```diff
@@ -139,6 +140,41 @@ globalThis.__dangerous_ON_edge_converter_returns_request = true;
...
+ * calls `module.findSourceMap` (https://nodejs.org/api/module.html#modulefindsourcemappath) which we haven't implemented causing a runtime error.
+ *
+ * To solve this issue this function gets all the `loadInstrumentationModule` declarations found in the file and removes all the statements
+ * from their bodies (making them no-op methods).
+ *
+ * Instrumentation is a Next.js feature for monitoring and logging (see: https://nextjs.org/docs/app/building-your-application/optimizing/instrumentation),
+ * the removal of this method's logic most likely breaks this feature (see: https://nextjs.org/docs/app/api-reference/file-conventions/instrumentation),
+ * so this function is likely temporary and something that we'll have to fix in the future.
+ *
+ * TODO: investigate and re-enable instrumentation (https://github.com/opennextjs/opennextjs-cloudflare/issues/160)
```
vicb, line 154: Can you please create a separate issue for this todo and add the link here. 
We can close the current issue when we merge this to main but keep the new one for further investigation. 

You can optionally move this function to the patch folder. 

Thanks

vicb (approved):
Thanks!

I have added a minor comment. Please move the existing issue to the experimental column of the board when you merge this.

packages/cloudflare/src/cli/build/bundle-server.ts
dario-piotrowicz: > move the parsing/stringify inside the function

I didn't want to do it inside the function since as I mentioned in my comment, I think that these functions should ideally all take an AST and return an AST, so actually following this structure already (without hiding the AST inside the function itself) make this clearer in my opinion, I am happy to move the parsing/stringifying into the function if you prefer that (it doesn't really make much of a difference in this case) 🙂

packages/cloudflare/src/cli/build/bundle-server.ts
dario-piotrowicz: > Can you please create a separate issue for this todo and add the link here.
> We can close the current issue when we merge this to main but keep the new one for further investigation.

Done 👍 

> You can optionally move this function to the patch folder.

oh yes! sorry I forgot how our files structure looks like! done! 🙇 👍

packages/cloudflare/src/cli/build/bundle-server.ts
dario-piotrowicz: updated as suggested 👍

vicb:
Thanks!
I moved #160 to the done in experimental column of the board.

PR #173, opened 2024-12-14
Fix + update examples

vicb:
See https://github.com/cloudflare/workerd/issues/3245 for details/rationale about the fix.

Updated the examples.

The blog and commerce examples fails, I'll create separate issues for them.
-> See #174 and #175, both added to the board.

### DIFF
--- a/packages/cloudflare/src/cli/build/bundle-server.ts
+++ b/packages/cloudflare/src/cli/build/bundle-server.ts
@@ -109,11 +109,13 @@ fetch = globalThis.fetch;
 const CustomRequest = class extends globalThis.Request {
   constructor(input, init) {
     if (init) {
-      delete init.cache;
-      if (init.body?.__node_stream__ === true) {
+      init = {
+        ...init,
+        cache: undefined,
         // https://github.com/cloudflare/workerd/issues/2746
-        init.body = __cf_stream.Readable.toWeb(init.body);
-      }
+        // https://github.com/cloudflare/workerd/issues/3245
+        body: init.body instanceof __cf_stream.Readable ? ReadableStream.from(init.body) : init.body,
+      };
     }
     super(input, init);
   }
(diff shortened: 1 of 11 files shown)


packages/cloudflare/src/cli/build/bundle-server.ts
```diff
@@ -109,11 +109,13 @@ fetch = globalThis.fetch;
...
-      delete init.cache;
-      if (init.body?.__node_stream__ === true) {
+      init = {
+        ...init,
+        cache: undefined,
         // https://github.com/cloudflare/workerd/issues/2746
-        init.body = __cf_stream.Readable.toWeb(init.body);
-      }
+        // https://github.com/cloudflare/workerd/issues/3245
+        body: init.body instanceof __cf_stream.Readable ? ReadableStream.from(init.body) : init.body,
```
conico974, line 117: Would this fail if `init.body` is a Readable from unenv ?

conico974 (approved):
LGTM

packages/cloudflare/src/cli/build/bundle-server.ts
vicb: From the top of my head I would say no.
It works with Node (/undici) only because a Readable is async iterable.
And we moved away from unenv Readable because it does not implement async iterator.

james-elicx (approved):

packages/cloudflare/src/cli/build/bundle-server.ts
conico974: I was thinking about user code actually, since this is overriding the global `Request`, user code or one of their lib could be using unenv or similar and `instanceof` would be false

vicb: Ah, got it now.

There should be no more instances of unenv `Readable` after https://github.com/unjs/unenv/pull/363 (https://github.com/cloudflare/workerd/issues/3245 has some details). Before that it was still possible to create an unenv Readable via the http module.

PR #181, opened 2024-12-15
inline project env files in worker

### LINKED ISSUE #157: Next .env file
We have a [`@next/env` shim](https://github.com/opennextjs/opennextjs-cloudflare/blob/021ddb7ef630e0162b9fb55226b82ce6bfd13efc/packages/cloudflare/src/cli/build/build-worker.ts#L83) to disable the loading of the .env files. This was needed to prevent accessing the filesystem.

We should inline the values at build time so that we can set them on `process.env`.

We will probably need a `NEXTJS_ENV` var that we could set to `development` in a `.dev.vars` to load the dev variables when running locally via `wrangler dev`. When deployed `NEXTJS_ENV` should load the prod vars if unset or whatever vars it points to when set.

james-elicx:
**Changes**

- Extracts env vars from .env files.
- Chucks them into a .env.mjs that is dynamically imported in the worker.


packages/cloudflare/src/cli/build/utils/extract-project-env-vars.ts
```diff
@@ -0,0 +1,38 @@
...
+ * 1. `.env.{mode}.local`
+ * 2. `.env.local`
+ * 3. `.env.{mode}`
+ * 4. `.env`
+ *
+ * In a monorepo, the env files in an app's directory will take precedence over
+ * the env files at the root of the monorepo.
+ */
+export function extractProjectEnvVars(options: BuildOptions) {
+  const wranglerDevVars = readEnvFiles([".dev.vars"], options);
```
vicb, line 33: What about doing this at runtime?

A wrangler "environment variable" could be defined `.dev.vars`, `wrangler.toml` and the dashboard.

We would have to embed all the `env.*` files in the code, I don't think this would be an issue?

packages/cloudflare/src/cli/templates/worker.ts
```diff
@@ -21,9 +21,13 @@ const cloudflareContextALS = new AsyncLocalStorage<CloudflareContext>();
   }
 );
 
+declare const __OPENNEXT_BUILD_TIME_ENV: Record<string, string>;
+
 export default {
   async fetch(request, env, ctx) {
-    return cloudflareContextALS.run({ env, ctx, cf: request.cf }, async () => {
+    const combinedEnv = { ...__OPENNEXT_BUILD_TIME_ENV, ...env };
```
vicb, line 28: I think Next env vars should be on `process.env` ?

packages/cloudflare/src/cli/build/utils/extract-project-env-vars.ts
james-elicx: My initial thought was we could use the readConfig function that is on the wrangler backlog somewhere, but this makes sense.

vicb:
A few thoughts (I'll do a deeper review later today)

I find "build time" to be confusing. Looking at the [next docs](https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables) there are 2 things: Non-`NEXT_PUBLIC_` and `NEXT_PUBLIC_` vars.

Does the current code support [referencing other vars](https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#referencing-other-variables) - maybe add a test.

It looks like we should support `development`, `production` and `test`. It looks like get the env from `NODE_ENV` so maybe `applyBuildTimeEnv(env["NEXTJS_ENV"])` should actually be `applyBuildTimeEnv(env["NEXTJS_ENV"] ?? process.env.NODE_ENV)`

To be more consistent with how Open Next work, we might want to generate `.env.development.mjs`, ... in the build folder of open next and include the file from worker.ts (rather than ESBulding the file). That would be similar to [compiling the config](https://github.com/opennextjs/opennextjs-cloudflare/blob/132b75e4a8492ca4c3c74d3b3ec07f50862722f5/packages/cloudflare/src/cli/build/index.ts#L36)

packages/cloudflare/src/cli/templates/worker.ts
```diff
@@ -34,6 +49,8 @@ export default {
         },
       });
 
+      await applyProjectEnvVars(env.NEXTJS_ENV ?? globalThis.process.env.NODE_ENV);
```
james-elicx, line 52: This was super weird. For some bizarre reason, NODE_ENV was requiring it be accessed through globalThis to obtain the value:

image
image

james-elicx:
> I find "build time" to be confusing. Looking at the [next docs](https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables) there are 2 things: Non-`NEXT_PUBLIC_` and `NEXT_PUBLIC_` vars.

Removed the reference to 'build-time'.

> Does the current code support [referencing other vars](https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#referencing-other-variables) - maybe add a test.

Yeah, that was actually why i used the newer dotenvx instead of dotenv :). Added a test.

> It looks like we should support `development`, `production` and `test`. It looks like get the env from `NODE_ENV` so maybe `applyBuildTimeEnv(env["NEXTJS_ENV"])` should actually be `applyBuildTimeEnv(env["NEXTJS_ENV"] ?? process.env.NODE_ENV)`
> 
> To be more consistent with how Open Next work, we might want to generate `.env.development.mjs`, ... in the build folder of open next and include the file from worker.ts (rather than ESBulding the file). That would be similar to [compiling the config](https://github.com/opennextjs/opennextjs-cloudflare/blob/132b75e4a8492ca4c3c74d3b3ec07f50862722f5/packages/cloudflare/src/cli/build/index.ts#L36)

Sorry I hadn't realised it was an intentional part of the design. I've undone that change. Instead, it's not putting all of them into an `.env.mjs` file, exporting each mode. I originally did `.env.${mode}.mjs`, but this was causing runtime issues with variable-based dynamic imports due to wrangler's bundle step, so i couldn't do that unfortunately.

packages/cloudflare/src/cli/build/patches/investigated/copy-package-cli-files.ts
```diff
@@ -6,7 +6,7 @@ import type { BuildOptions } from "@opennextjs/aws/build/helper.js";
 import { Config } from "../../../config";
 
 /**
- * Copies the template files present in the cloudflare adapter package into the standalone node_modules folder
+ * Copies the template files present in the cloudflare adapter package into the standalone node_modules folder.
```
james-elicx, line 9:
```suggestion
 * Copies the template files present in the cloudflare adapter package into the standalone node_modules folder
```

packages/cloudflare/src/cli/build/patches/investigated/copy-env-files.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import { BuildOptions } from "@opennextjs/aws/build/helper.js";
+
+import { extractProjectEnvVars } from "../../utils";
+
+/**
+ * Copies the values extracted from the project's env files to the output directory for use in the worker.
+ */
+export function copyEnvFiles(options: BuildOptions) {
+  ["production", "development", "test"].forEach((mode) =>
```
vicb, line 12: it looks like we need the `.local` version for `wrangler dev` too
ref https://nextjs.org/docs/app/building-your-application/configuring/environment-variables#environment-variable-load-order

packages/cloudflare/src/cli/build/utils/extract-project-env-vars.ts
```diff
@@ -0,0 +1,34 @@
+import * as fs from "node:fs";
+import * as path from "node:path";
+
+import { parse } from "@dotenvx/dotenvx";
+import type { BuildOptions } from "@opennextjs/aws/build/helper.js";
+
+function readEnvFiles(fileNames: string[], { monorepoRoot, appPath }: BuildOptions) {
```
vicb, line 7: nit: maybe it would be clearer to separate reading the file and merging the values in 2 different functions?

Also a comment to indicate the merge order would we helpful

packages/cloudflare/src/cli/build/utils/extract-project-env-vars.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Merged variables respect the following priority order.
+ * 1. `.env.{mode}.local`
+ * 2. `.env.local`
+ * 3. `.env.{mode}`
+ * 4. `.env`
+ *
+ * In a monorepo, the env files in an app's directory will take precedence over
+ * the env files at the root of the monorepo.
+ */
+export function extractProjectEnvVars(mode: string, options: BuildOptions) {
```
vicb, line 32: Doesn't that mean that we will always use the `.local` files while we only want to use them for `wrangler dev` but not in prod?

packages/cloudflare/src/cli/templates/worker.ts
```diff
@@ -21,6 +21,21 @@ const cloudflareContextALS = new AsyncLocalStorage<CloudflareContext>();
   }
 );
 
+async function applyProjectEnvVars(mode = "production") {
```
vicb, line 24: Would doing this once at top-level remove the need for `process.env.__OPENNEXT_PROCESSED_ENV` ?

nit: maybe rename secrets to nextEnvVars? (those are not all secrets, i.e. `NEXT_PUBLIC_`)

james-elicx: We can't do this at the top-level because of the desire to have the `NEXTJS_ENV` .dev.vars variable. If that's not needed, then I can move it to the top-level

packages/cloudflare/src/cli/build/utils/extract-project-env-vars.ts
james-elicx: Yeah. How come you wouldn't want them in production? Next.js would use those files. I suppose you're talking about if someone builds it locally and then deploys, but in that case, you could make the argument that we shouldnt include any .env files at all because technically they're all local to the machine. I'm not quite sure what else you could do here - open to suggestions though.

packages/cloudflare/src/cli/build/patches/investigated/copy-env-files.ts
james-elicx: it's extracting the local ones in the `extractProjectEnvVars` function

packages/cloudflare/src/cli/build/utils/extract-project-env-vars.ts
vicb: @james-elicx I did a quick test and you are right, `.local` files are used if present, regardless if it is a prod or dev build. I was wrong to think that they would only be used in dev.

packages/cloudflare/src/cli/build/utils/extract-project-env-vars.ts
```diff
@@ -0,0 +1,35 @@
...
+    .filter((filePath) => fs.existsSync(filePath) && fs.statSync(filePath).isFile())
+    .map((filePath) => parse(fs.readFileSync(filePath).toString()));
+}
+
+/**
+ * Extracts the environment variables defined in various .env files for a project.
+ *
+ * The `NEXTJS_ENV` environment variable in `.dev.vars` determines the mode.
+ *
+ * Merged variables respect the following priority order.
```
vicb, line 22: Maybe add a link to https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#environment-variable-load-order

packages/cloudflare/src/cli/build/utils/extract-project-env-vars.ts
```diff
@@ -0,0 +1,35 @@
...
+ * 1. `.env.{mode}.local`
+ * 2. `.env.local`
+ * 3. `.env.{mode}`
+ * 4. `.env`
+ *
```
james-elicx, line 27:
```suggestion
 *
 * https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#environment-variable-load-order
 *
```

packages/cloudflare/src/cli/build/utils/extract-project-env-vars.ts
vicb: Maybe this can be `readEnvFile` (single file) and you use `.map()` - up to you

packages/cloudflare/src/cli/build/patches/investigated/copy-env-files.ts
```diff
@@ -0,0 +1,18 @@
+import fs from "node:fs";
+import path from "node:path";
+
+import { BuildOptions } from "@opennextjs/aws/build/helper.js";
+
+import { extractProjectEnvVars } from "../../utils";
+
+/**
+ * Copies the values extracted from the project's env files to the output directory for use in the worker.
```
vicb, line 9: Maybe:

```suggestion
 * Compiles the values extracted from the project's env files to the output directory for use in the worker.
```

And rename the function `compilesEnvFiles` - similar to `compileOpenNextConfig`, `cvompiloeCache`, ...

packages/cloudflare/src/cli/templates/worker.ts
vicb: Weird!

What about we default to "production" if `env.NEXTJS_ENV === undefined`
As discussed, we should add a `.dev.vars` for new projects

james-elicx: And ignore `process.env.NODE_ENV` altogether I assume then?

james-elicx: I've created https://github.com/opennextjs/opennextjs-cloudflare/issues/189 for the `.dev.vars` file.

vicb: > And ignore `process.env.NODE_ENV` altogether I assume then?

The thing is `process.env` should be empty on first request to a worker anyway

vicb:
Do you think you can add a simple test, maybe to the api example app?

vicb (approved):
Awesome PR, thanks!

I would move the code out of patch*, I think it should be part of the build.

Approving the PR, you can pick whatever your prefer. 

🙏

Issue #28, opened 2024-09-23
Fail build (/throw errors) when code patching fails


dario-piotrowicz:
In various places we do some code patching that is completely necessary for the worker to run, for example:

https://github.com/opennextjs/opennextjs-cloudflare/blob/42320e78e4e8f1abafd2336c3a384fedc2f8a394/packages/cloudflare/src/build/patches/to-investigate/patch-find-dir.ts#L12-L22

But we do not check if such patches have been completed successfully (e.g. in the example above we never check wether `"function findDir(dir, name) {"` is actually found and replaced in the code or not)

In order not to complete the build process in a seemingly successful way just to end up with a non-fully (or not-at-all) functional worker we should throw errors and fail the build process in case some of these patches aren't getting successfully applied.

dario-piotrowicz:
> [!NOTE]
> Later we should probably consider refactoring the code regex replacements with AST manipulations (using `ts-morph` as it is already use in the [`updateWebpackChunksFile` utility](https://github.com/opennextjs/opennextjs-cloudflare/tree/main/packages/cloudflare/src/build/patches/investigated/update-webpack-chunks-file))

PR #184, opened 2024-12-17
make sure `patchFindDir` works with Next.js 15 and add patching validation check

### LINKED ISSUE #172: [BUG] 404 - on experimental branch with Nextjs 15.1
### Describe the bug

I updated Next.js to the latest version on an experimental branch, and now I’m getting 404 errors everywhere.

### Steps to reproduce

1) clone https://github.com/himanshu-ntml/opennext-15
2) npm run preview

### Expected behavior

Should work as expected

experimental

### Node.js version

v22.11.0

### Wrangler version

3.95.0

### next info output

```shell
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.1.0: Thu Oct 10 21:03:15 PDT 2024; root:xnu-11215.41.3~2/RELEASE_ARM64_T6000
  Available memory (MB): 65536
  Available CPU cores: 10
Binaries:
  Node: 22.11.0
  npm: 10.9.0
  Yarn: 1.22.22
  pnpm: 9.12.3
Relevant Packages:
  next: 15.1.0 // Latest available version is detected (15.1.0).
  eslint-config-next: 14.2.5
  react: 19.0.0
  react-dom: 19.0.0
  typescript: 5.7.2
Next.js Config:
  output: N/A
```

### Additional context

_No response_

dario-piotrowicz:
resolves #172 

___

The `patchFindDir` function: https://github.com/opennextjs/opennextjs-cloudflare/blob/acec1c3ce61602a5bbb3f1330b5d1e0d0110b3f6/packages/cloudflare/src/cli/build/patches/to-investigate/patch-find-dir.ts#L12

Stopped working since a new `name` variable got introduced in the Next.js source code, causing the bundled code to rename the function's `name` parameter to `name2` instead:
Screenshot 2024-12-17 at 19 33 16

I fixed this by updating the code replacement to also accept the variables with a numerical suffix. I also added a check to make sure that the patching is applied (if we already had the check in place that would have saved me lots of time here! 😭)

A few notes:
 - I wanted to update the code to use AST manipulation, but since this patch is in the `to-investigate` directory I guess that we assume that this is temporary, so maybe it's not worth investing time to update the code to use ASTs? (if it is I am more than happy to do so!)
 - again, having the `patchedCode !== code` check would have saved me lots of time here, we need to add it to all the patch functions to avoid wasting time again, I will do that in a followup PR
 - potentially we can hit this issue with other patching functions, I can update those as well to accept numerical suffixes in their variables/paramenters, but I think that adding the `patchedCode !== code` check would be good enough for now

### DIFF
--- a/packages/cloudflare/src/cli/build/patches/to-investigate/patch-find-dir.ts
+++ b/packages/cloudflare/src/cli/build/patches/to-investigate/patch-find-dir.ts
@@ -11,18 +11,24 @@ import { Config } from "../../../config";
  */
 export function patchFindDir(code: string, config: Config): string {
   console.log("# patchFindDir");
-  return code.replace(
-    "function findDir(dir, name) {",
-    `function findDir(dir, name) {
-			if (dir.endsWith(".next/server")) {
-			if (name === "app") {
+  const patchedCode = code.replace(
+    /function findDir\((dir\d*), (name\d*)\) {/,
+    `function findDir($1, $2) {
+			if ($1.endsWith(".next/server")) {
+			if ($2 === "app") {
 			  return ${existsSync(`${join(config.paths.output.standaloneAppServer, "app")}`)};
 	    }
-			if (name === "pages") {
+			if ($2 === "pages") {
 			  return ${existsSync(`${join(config.paths.output.standaloneAppServer, "pages")}`)};
 	    }
 		}
-		throw new Error("Unknown findDir call: " + dir + " " + name);
+		throw new Error("Unknown findDir call: " + $1 + " " + $2);
 		`
   );
+
+  if (patchedCode === code) {
+    throw new Error("Patch `patchFindDir` not applied");
+  }
+
+  return patchedCode;
 }
--- a/packages/cloudflare/src/cli/build/patches/to-investigate/patch-find-dir.ts
+++ b/packages/cloudflare/src/cli/build/patches/to-investigate/patch-find-dir.ts
@@ -12,17 +12,17 @@ import { Config } from "../../../config";
 export function patchFindDir(code: string, config: Config): string {
   console.log("# patchFindDir");
   const patchedCode = code.replace(
-    /function findDir\((dir\d*), (name\d*)\) {/,
-    `function findDir($1, $2) {
-			if ($1.endsWith(".next/server")) {
-			if ($2 === "app") {
+    /function findDir\((?<dir>dir\d*), (?<name>name\d*)\) {/,
+    `function findDir($dir, $name) {
+			if ($dir.endsWith(".next/server")) {
+			if ($name === "app") {
 			  return ${existsSync(`${join(config.paths.output.standaloneAppServer, "app")}`)};
 	    }
-			if ($2 === "pages") {
+			if ($name === "pages") {
 			  return ${existsSync(`${join(config.paths.output.standaloneAppServer, "pages")}`)};
 	    }
 		}
-		throw new Error("Unknown findDir call: " + $1 + " " + $2);
+		throw new Error("Unknown findDir call: " + $dir + " " + $name);
 		`
   );
 


packages/cloudflare/src/cli/build/patches/to-investigate/patch-find-dir.ts
```diff
@@ -11,18 +11,24 @@ import { Config } from "../../../config";
  */
 export function patchFindDir(code: string, config: Config): string {
   console.log("# patchFindDir");
-  return code.replace(
-    "function findDir(dir, name) {",
-    `function findDir(dir, name) {
-			if (dir.endsWith(".next/server")) {
-			if (name === "app") {
+  const patchedCode = code.replace(
```
vicb, line 14: Maybe add a comment here with the pre-15.1 code and 15.1 code

packages/cloudflare/src/cli/build/patches/to-investigate/patch-find-dir.ts
```diff
@@ -11,18 +11,24 @@ import { Config } from "../../../config";
...
+    /function findDir\((dir\d*), (name\d*)\) {/,
```
vicb, line 15: Maybe use [named capturing groups](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Named_capturing_group) to make the replacement more readable

vicb (approved):
Thanks Dario 🙏

Does this also resolve the warnings (`eval(require)...`?)

Feel to to assign https://github.com/opennextjs/opennextjs-cloudflare/issues/28 to yourself and add it to the board :)

packages/cloudflare/src/cli/build/patches/to-investigate/patch-find-dir.ts
dario-piotrowicz: are you sure? it feels a bit overkill to me here 😕

packages/cloudflare/src/cli/build/patches/to-investigate/patch-find-dir.ts
dario-piotrowicz: good point 👍

packages/cloudflare/src/cli/build/patches/to-investigate/patch-find-dir.ts
vicb: It might help if we later need to convert the patch to AST but your call :)

dario-piotrowicz:
> Does this also resolve the warnings (`eval(require)...`?)

no 🥲 

I can look into those if you want, but since they are only warnings I assumed that they are lower priority

Regarding warnings we also get this one:
Screenshot 2024-12-17 at 19 56 47

vicb:
> > Does this also resolve the warnings (`eval(require)...`?)
> 
> no 🥲
> 
> I can look into those if you want, but since they are only warnings I assumed that they are lower priority
> 
> Regarding warnings we also get this one: Screenshot 2024-12-17 at 19 56 47

Maybe create an issue for `eval(require)...` so that we look into it at some point.

packages/cloudflare/src/cli/build/patches/to-investigate/patch-find-dir.ts
dario-piotrowicz: done 👍

packages/cloudflare/src/cli/build/patches/to-investigate/patch-find-dir.ts
dario-piotrowicz: I'd leave it as is if it's ok by you 🙂 

but if you do prefer a comment, could you please add a code suggestion here and I'd be happy to just go with whatever you'd fine useful here 🙂

PR #186, opened 2024-12-18
make sure all code patch steps are validated

### LINKED ISSUE #28: Fail build (/throw errors) when code patching fails

dario-piotrowicz:
resolves #28 

___

This PR makes sure that all our code patches are applied (that they do change the worker's code)

As part of this I did a bit of refactoring to remove every single function's log and validation and consolidate then
in a single function `patchCodeWithValidations`

___

The result/logs in case everything works as expected:
Screenshot 2024-12-18 at 10 25 37

The result/logs in case a step fails:
Screenshot 2024-12-18 at 10 31 17


packages/cloudflare/src/cli/build/bundle-server.ts
```diff
@@ -139,6 +128,7 @@ globalThis.__dangerous_ON_edge_converter_returns_request = true;
     );
   }
 
+  console.log();
```
vicb, line 131: It's this intended?

packages/cloudflare/src/cli/build/bundle-server.ts
```diff
@@ -192,3 +192,33 @@ function createFixRequiresESBuildPlugin(config: Config): Plugin {
...
+ * @param code the code to apply the patches to
+ * @param patches array of functions that take a string (pre-patch code) and return a string (post-patch code)
+ * @returns the patched code
+ */
+async function patchCodeWithValidations(
+  code: string,
+  patches: [string, (code: string) => string | Promise<string>][]
+): Promise<string> {
+  console.log(`Applying code patches:`);
+  let prePatchCode = code;
```
vicb, line 209: Nit: you can remove this line and add a let l215

vicb (approved):
Thanks!

packages/cloudflare/src/cli/build/bundle-server.ts
dario-piotrowicz: Yes, for this empty line:
Screenshot 2024-12-18 at 11 36 06

probably a `\n` in the other console.log would be better/clearer?

vicb: Moving this log to l222 inside `patchCodeWithValidations` as "Applied N patches" might be a bit clearer, up to you.

dario-piotrowicz: sounds good, I don't have a strong preference here 🙂

PR #190, opened 2024-12-18
prevent eval patch from failing build

james-elicx:
The e2es are currently failing to run due to the eval patch not being required for Next.js 14 (which the e2es are currently using).

image

This PR makes that patch optional, i.e. doesn't throw an error if the patch hasn't made any changes.


vicb (approved):
Thanks!

PR #191, opened 2024-12-18
add default to exports

wesbos:
I'm trying to write a script that works alongside my Next.js project, and since Next.js is still CJS - I'm unable to use `"type": "module"` in my package.json

This package is required as part of my database setup, and it even works outside Next.js, but for whatever cursed CJS → ESM compat reason, I can't import it.

default is a generic fallback - and allows me to import it.


vicb (approved):
Thanks!

PR #192, opened 2024-12-19
Drop tsup in favor of tsc

vicb:
Few reasons:
- `tsup` is not really needed
- consistency with Open Next (aws)
- A cache adapter will be added next and it needs to be compiled with tsc

Note: `wrangler.toml` needs to be updated gto point to `worker.js` (was `.ts` before)


packages/cloudflare/package.json
```diff
@@ -69,6 +70,7 @@
     "@dotenvx/dotenvx": "catalog:"
   },
   "peerDependencies": {
+    "rimraf": "^6.0.1",
```
james-elicx, line 73: Probably worth moving this to the workspace file for consistency (the two eslint ones above might need moving too)

james-elicx (approved):

PR #193, opened 2024-12-19
create an open-next.config.ts file for the user in case one is not already present

### LINKED ISSUE #188: Create a new `open-next.config.ts` file for the user if there isn't already one in their directory
To reduce friction and ease people's onboarding we should create an appropriate default `open-next.config.ts` file in the directory if it doesn't already exist (maybe after asking the user if that's ok?)

> [!Note]
> This is for Next.js app migrations, for new applications we'll have C3 create the file)

> [!Note]
> Long term the file should include a `withCloudflare` utility as mentioned here: https://github.com/opennextjs/opennextjs-cloudflare/issues/114 I can look into doing that here or we can do it separately, I am fine with either (as I tend to prefer small focused changes I'd probably go for not treating the two as separate (related) feature requests)

dario-piotrowicz:
resolves #188

### DIFF
--- a/packages/cloudflare/src/cli/build/index.ts
+++ b/packages/cloudflare/src/cli/build/index.ts
@@ -1,4 +1,4 @@
-import { cpSync } from "node:fs";
+import { cpSync, existsSync } from "node:fs";
 import { createRequire } from "node:module";
 import { dirname, join } from "node:path";
 
@@ -11,6 +11,7 @@ import * as buildHelper from "@opennextjs/aws/build/helper.js";
 import { printHeader, showWarningOnWindows } from "@opennextjs/aws/build/utils.js";
 import logger from "@opennextjs/aws/logger.js";
 import type { OpenNextConfig } from "@opennextjs/aws/types/open-next.js";
+import Enquirer from "enquirer";
 
 import type { ProjectOptions } from "../config";
 import { containsDotNextDir, getConfig } from "../config";
@@ -33,6 +34,8 @@ export async function build(projectOpts: ProjectOptions): Promise<void> {
   const baseDir = projectOpts.sourceDir;
   const require = createRequire(import.meta.url);
   const openNextDistDir = dirname(require.resolve("@opennextjs/aws/index.js"));
+
+  await createOpenNextConfigIfNotExistent(baseDir);
 
   const { config, buildDir } = await compileOpenNextConfig(baseDir);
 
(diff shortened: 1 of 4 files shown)


packages/cloudflare/src/cli/build/index.ts
```diff
@@ -94,6 +97,34 @@ export async function build(projectOpts: ProjectOptions): Promise<void> {
...
+        message: "Missing required `open-next.config.ts` file, do you want to create one?",
+        type: "confirm",
+        initial: "y",
+      })) as { [questionName]: boolean }
+    )[questionName];
+    if (!answer) {
+      throw new Error("A `open-next.config.ts` file is mandatory, aborting!");
+    }
+
+    cpSync(`${import.meta.dirname}/templates/defaults/open-next.config.ts`, openNextConfigPath);
```
vicb, line 124: That will not work any more after my pending PR compiling all the files.

Maybe return that as a string instead?

dario-piotrowicz: what do you mean by "return as a string"? inlining the file's content as a string here?

vicb: That's one possibility.

Or you can copy the compiled .js and rename it to .ts (but probably we want some types later)

Or you can create a /template folder alongside /src, add it to the npm package and use that

...

packages/cloudflare/templates/defaults/open-next.config.ts
```diff
@@ -0,0 +1,28 @@
+// default open-next.config.ts file created by @opennextjs/cloudflare
+
+const config = {
+  default: {
+    override: {
+      wrapper: "cloudflare-node",
+      converter: "edge",
+      incrementalCache: "dummy",
```
vicb, line 8: may be setup the cache (see the blog example)

packages/cloudflare/src/cli/build/index.ts
```diff
@@ -100,6 +104,34 @@ export async function build(projectOpts: ProjectOptions): Promise<void> {
...
 }
 
+/**
+ * Creates a `open-next.config.ts` file for the user if it doesn't exist, but only after asking for the user's confirmation.
+ *
+ * If the user refuses an error is thrown (since the file is mandatory).
+ *
+ * @param baseDir the Next.js app root folder
+ */
+async function createOpenNextConfigIfNotExistent(baseDir: string): Promise<void> {
```
vicb, line 114: nit: can you pass projectOpts and derive the src/dst path from that?
It will be easier when we will later move to BuildOptions (OpenNext)
Thx

vicb (approved):

PR #195, opened 2024-12-20
fix: No body for 101, 204, 205, or 304 status code

vicb:
Fix a KI when browser cache was enabled by pulling the fix from https://github.com/opennextjs/opennextjs-aws/pull/683

### DIFF
--- a/packages/cloudflare/README.md
+++ b/packages/cloudflare/README.md
@@ -70,7 +70,6 @@ export default config;
 
 - Next cache is not supported in the experimental branch yet
 - `▲ [WARNING] Suspicious assignment to defined constant "process.env.NODE_ENV" [assign-to-define]` can safely be ignored
-- You should test with cache disabled in the developer tools
 - Maybe more, still experimental...
 
 ## Local development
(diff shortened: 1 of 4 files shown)


petebacondarwin (commented):
What is a KI?

I feel uncomfortable about landing "fixes" with no tests.
How hard would it be to add tests that prove this bump to the aws dependency fixes the problem?

vicb:
KI = known issue (See the README update).

I tested locally with the blog example, reloading the page would generated the error as it adds a body to a 304. This might be possible to e2e but I'm not sure how e2e behaves regarding to cache.

I'm confident that the issue is fixed.

packages/cloudflare/README.md
```diff
@@ -70,7 +70,6 @@ export default config;
 
 - Next cache is not supported in the experimental branch yet
```
conico974, line 71: This can be removed ( or changed in the cache PR as well )

conico974 (approved):
LGTM

packages/cloudflare/README.md
vicb: ^that should actually go in the #194

PR #208, opened 2024-12-27
update location of output path in success message

dario-piotrowicz:
It's quite minor, but given how the structure of the output has changed I think that the location indicated in the success message should be updated:
Screenshot 2024-12-27 at 13 08 22

So in this PR I am simply updating this to the location of the `worker.js` file (which I think it's the one we should really point users to)


vicb (approved):
Thanks!

PR #211, opened 2024-12-27
remove `process.env.NODE_ENV` warning by bumping `@opennextjs/aws` dependency

dario-piotrowicz:
A quick (potentially temporary) solution for the `process.env.NODE_ENV` warning:
Screenshot 2024-12-27 at 18 20 06

___

This change consists in simply bumping the aws dependency to its prerelease package generated in https://github.com/opennextjs/opennextjs-aws/pull/686 ([comment](https://github.com/opennextjs/opennextjs-aws/pull/686#issuecomment-2564340650))

### DIFF
--- a/.changeset/real-rings-walk.md
+++ b/.changeset/real-rings-walk.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+bump `"@opennextjs/aws` dependency to `https://pkg.pr.new/@opennextjs/aws@686`
--- a/packages/cloudflare/package.json
+++ b/packages/cloudflare/package.json
@@ -71,7 +71,7 @@
   },
   "dependencies": {
     "@dotenvx/dotenvx": "catalog:",
-    "@opennextjs/aws": "https://pkg.pr.new/@opennextjs/aws@684",
+    "@opennextjs/aws": "https://pkg.pr.new/@opennextjs/aws@686",
     "glob": "catalog:",
     "rimraf": "catalog:",
     "ts-morph": "catalog:",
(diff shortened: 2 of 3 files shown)


vicb (commented):
It looks like the right (safe) thing to do but because this is Open Next code, it would be easier to modify the code at https://github.com/opennextjs/opennextjs-aws/blob/9595714ac23e5f131b879d04d5cfb2a5d11bdbdd/packages/open-next/src/adapters/util.ts#L4

dario-piotrowicz:
> It looks like the right (safe) thing to do but because this is Open Next code, it would be easier to modify the code at https://github.com/opennextjs/opennextjs-aws/blob/9595714ac23e5f131b879d04d5cfb2a5d11bdbdd/packages/open-next/src/adapters/util.ts#L4

yes I agree, but I am not sure how to discern there if the code is being used by Cloudflare and also I wouldn't want to overcomplicate things there only for our use case, I think that it's probably something worth tackling/looking into when we move to a shared monorepo setup, right now I would just add this quick patch to remove the warning not to annoy/scare users away

but I agree that ideally it should be solved in a better way, I just don't think that it's a high enough priority right for us to tackle in such a way right now

if you disagree with the patch I am totally fine not including it 👍

vicb:
Yes the line I pointed to is the one causing the warning.

It is better to make the function 2 lines there to avoid adding a patch that we would migrate to AST later.

dario-piotrowicz:
> Yes the line I pointed to is the one causing the warning.
> 
> It is better to make the function 2 lines there to avoid adding a patch that we would migrate to AST later.

https://github.com/opennextjs/opennextjs-aws/pull/686 👍

dario-piotrowicz:
@vicb PR updated to only bump the aws dependency

vicb (approved):

Issue #206, opened 2024-12-26
[BUG] NextAuth errors with [unenv] crypto.createCipheriv is not implemented yet!


RUNFUNRUN:
> [!TIP]
> @arnavgupta00 found a workaround, see https://github.com/cloudflare/workerd/issues/3277#issuecomment-2614656148

https://github.com/cloudflare/workerd/issues/3277#issuecomment-2614656148

### Describe the bug

When next-auth is used and next-auth related functions are executed, the following error occurs.

```
[wrangler:err] TypeError: Cannot assign to read only property 'cache' of object '#<eA>'
    at new CustomRequest (file:///Users/runfunrun/projects/test/opennext-cf/.open-next/server-functions/default/handler.mjs:20:18)
    at new $ (file:///Users/runfunrun/projects/test/opennext-cf/.open-next/middleware/handler.mjs:797:62)
    at o5 (file:///Users/runfunrun/projects/test/opennext-cf/.open-next/middleware/handler.mjs:5655:16)
    at so (file:///Users/runfunrun/projects/test/opennext-cf/.open-next/middleware/handler.mjs:5797:19)
    at Object.handler (file:///Users/runfunrun/projects/test/opennext-cf/.open-next/middleware/handler.mjs:5782:20)
    at null.<anonymous> (file:///Users/runfunrun/projects/test/opennext-cf/.open-next/middleware/handler.mjs:1268:174)
    at Object.wrap (file:///Users/runfunrun/projects/test/opennext-cf/.open-next/middleware/handler.mjs:1194:20)
    at null.<anonymous> (file:///Users/runfunrun/projects/test/opennext-cf/.open-next/middleware/handler.mjs:1266:276)
    at null.<anonymous> (file:///Users/runfunrun/projects/test/opennext-cf/.open-next/middleware/handler.mjs:1106:25)
    at i2.with (file:///Users/runfunrun/projects/test/opennext-cf/.open-next/middleware/handler.mjs:6421:25)
```

### Steps to reproduce

1. clone this repository https://github.com/RUNFUNRUN/opennext-cf
2. install packages
```bash
bun i
```
3. copy env
```bash
cp sample.env .env
```
(Environment variables can be reproduced as in the sample.)
4. run preview 
```bash
bun run preview
```
5. open http://localhost:8787/
If I try deleting middleware.ts, I can access http://localhost:8787/ , but when I press the Sign In button, I get the same error.

### Expected behavior

next-auth works

[...]

PR #212, opened 2024-12-27
fix: delete init.cache rather than assign undefined

vicb:
Fix an issue with NextAuth (#206)

Creating a PR so that users can test, but need to come with a proper test

### DIFF
--- a/packages/cloudflare/src/cli/build/bundle-server.ts
+++ b/packages/cloudflare/src/cli/build/bundle-server.ts
@@ -94,7 +94,7 @@ fetch = globalThis.fetch;
 const CustomRequest = class extends globalThis.Request {
   constructor(input, init) {
     if (init) {
-      init.cache = undefined;
+      delete init.cache;
       // https://github.com/cloudflare/workerd/issues/2746
       // https://github.com/cloudflare/workerd/issues/3245
       Object.defineProperty(init, "body", {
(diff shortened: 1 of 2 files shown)


packages/cloudflare/src/cli/build/bundle-server.ts
```diff
@@ -94,7 +94,7 @@ fetch = globalThis.fetch;
 const CustomRequest = class extends globalThis.Request {
   constructor(input, init) {
     if (init) {
-      init.cache = undefined;
+      delete init.cache;
```
vicb, line 97: Using `Object.defineProperty()` would throw later on l88.

dario-piotrowicz (approved):
Looks good to me 👍 

The first thing that came to mind when seeing the change is that a comment could help clarify why we're not setting the value to `undefined` but not having the comment is not a huge deal, especially since this seems to be a bit of a specific case

Besides (and related to) the above one worry I have here is that I feel that there is nothing to prevent this code to be changed in the future reintroducing the issue, I think it would be very good if we could have some e2e testing here, but I also understand that this might be a bit overkill and not worth it right now, so I am fine skipping that too

packages/cloudflare/src/cli/build/bundle-server.ts
dario-piotrowicz: I find this and l88 quite cryptic 🥲, hopefully this code will be improved/made clearer at some point

vicb:
We should definitely find a way to test that (as mentioned in the description).
I'll create a separate issue for that.
Merging as the reporter of #206 saw an improvement.

PR #220, opened 2025-01-02
create a wrangler config file for the user in case one is not already present (alongside ways to opt out of this)

dario-piotrowicz:
Similarly to https://github.com/opennextjs/opennextjs-cloudflare/pull/193 I thought it'd be a nice touch if we created the `wrangler.toml` file for the user instead of telling them to create and populate it themselves 🙂 

___

Related docs update: https://github.com/opennextjs/docs/pull/45

### DIFF
--- a/.changeset/lazy-balloons-report.md
+++ b/.changeset/lazy-balloons-report.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+create a wrangler.toml file for the user in case one is not already presentwq
--- a/.changeset/lazy-balloons-report.md
+++ b/.changeset/lazy-balloons-report.md
@@ -2,7 +2,7 @@
 "@opennextjs/cloudflare": patch
 ---
 
-checks and creates a wrangler.jsonc file for the user in case one wrangler.(toml|json|jsonc) file is not already present
+checks and creates a `wrangler.jsonc` file for the user in case a `wrangler.(toml|json|jsonc)` file is not already present
 
 also introduce a new `--skipWranglerConfigCheck` cli flag and a `SKIP_WRANGLER_CONFIG_CHECK`
 environment variable that allows users to opt out of the above check (since developers might
--- a/.changeset/lazy-balloons-report.md
+++ b/.changeset/lazy-balloons-report.md
@@ -2,7 +2,7 @@
 "@opennextjs/cloudflare": patch
 ---
 
-checks and creates a `wrangler.jsonc` file for the user in case a `wrangler.(toml|json|jsonc)` file is not already present
+check and create a `wrangler.jsonc` file for the user in case a `wrangler.(toml|json|jsonc)` file is not already present
 
 also introduce a new `--skipWranglerConfigCheck` cli flag and a `SKIP_WRANGLER_CONFIG_CHECK`
 environment variable that allows users to opt out of the above check (since developers might
--- a/.changeset/lazy-balloons-report.md
+++ b/.changeset/lazy-balloons-report.md
@@ -2,4 +2,8 @@
 "@opennextjs/cloudflare": patch
 ---
 
-create a wrangler.toml file for the user in case one is not already presentwq
+checks and creates a wrangler.jsonc file for the user in case one wrangler.(toml|json|jsonc) file is not already present
+
+also introduce a new `--skipWranglerConfigCheck` cli flag and a `SKIP_WRANGLER_CONFIG_CHECK`
+environment variable that allows users to opt out of the above check (since developers might
+want to use custom names for their config files)
--- a/.changeset/lazy-balloons-report.md
+++ b/.changeset/lazy-balloons-report.md
@@ -6,4 +6,4 @@ check and create a `wrangler.jsonc` file for the user in case a `wrangler.(toml|
 
 also introduce a new `--skipWranglerConfigCheck` cli flag and a `SKIP_WRANGLER_CONFIG_CHECK`
 environment variable that allows users to opt out of the above check (since developers might
-want to use custom names for their config files)
+want to use custom locations for their config files)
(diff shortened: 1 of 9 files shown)


packages/cloudflare/src/cli/build/index.ts
```diff
@@ -178,3 +180,79 @@ function ensureCloudflareConfig(config: OpenNextConfig) {
...
+
+/**
+ * Creates a `wrangler.toml` file for the user if it doesn't exist, but only after asking for the user's confirmation.
+ *
+ * If the user refuses an error is thrown (since the file is mandatory).
+ *
+ * @param projectOpts The options for the project
+ */
+async function createWranglerTomlIfNotExistent(projectOpts: ProjectOptions): Promise<void> {
+  const wranglerTomlPath = join(projectOpts.sourceDir, "wrangler.toml");
```
vicb, line 192: We should probably also check for .json or .jsonc.

Maybe we should create a .json file? It's probably easier for new comers.

packages/cloudflare/src/cli/build/index.ts
```diff
@@ -178,3 +180,79 @@ function ensureCloudflareConfig(config: OpenNextConfig) {
...
+
+  if (!existsSync(wranglerTomlPath)) {
+    const answer = await askConfirmation("Missing required `wrangler.toml` file, do you want to create one?");
+
+    if (!answer) {
+      throw new Error("The `wrangler.toml` file is required, aborting!");
```
vicb, line 198: Not sure if we should throw. Users might not want to generate the toml (or json) format?

packages/cloudflare/src/cli/build/index.ts
```diff
@@ -178,3 +180,79 @@ function ensureCloudflareConfig(config: OpenNextConfig) {
...
+      (await resp.json()) as {
+        "dist-tags": { latest: string };
+      }
+    )["dist-tags"].latest;
+
+    // The format of the workerd version is `major.yyyymmdd.patch`.
+    const match = latestWorkerdVersion.match(/\d+\.(\d{4})(\d{2})(\d{2})\.\d+/);
+
+    if (match) {
+      const [, year, month, date] = match ?? [];
```
vicb, line 250: Nit: ?? Should not be needed?

packages/cloudflare/templates/defaults/wrangler.toml
```diff
@@ -0,0 +1,7 @@
+main = ".open-next/worker.js"
+name = "app-name"
+
+compatibility_date = "2024-12-30"
+compatibility_flags = ["nodejs_compat"]
+
+assets = { directory = ".open-next/assets", binding = "ASSETS" }
```
vicb, line 7: Maybe add a commented out kv binding for the cache

packages/cloudflare/src/cli/build/index.ts
dario-piotrowicz: good callout, no, I think that that's totally unnecessary 👍 

I just copied the code from https://github.com/cloudflare/workers-sdk/blob/2e78812ade7cd7361b023c90afe06221a52b79eb/packages/create-cloudflare/src/helpers/compatDate.ts#L30 without looking into it much 😅

packages/cloudflare/src/cli/build/index.ts
dario-piotrowicz: > We should probably also check for .json or .jsonc.

mhhh.... oh yeah that's a good call 🙂 

> Maybe we should create a .json file? It's probably easier for new comers.

I'm happy with whatever 👍 (we do recommend creating a toml file in the [docs](https://opennext.js.org/cloudflare/get-started#2-install-wrangler-and-add-a-wranglertoml-file))

packages/cloudflare/src/cli/build/index.ts
dario-piotrowicz: I can set it as a warning if you prefer 👍

packages/cloudflare/src/cli/build/index.ts
vicb: I think it's because json still required a flag when the docs were created.

packages/cloudflare/src/cli/build/index.ts
vicb: +1 for a warning

dario-piotrowicz: updated 👍

packages/cloudflare/templates/defaults/wrangler.toml
dario-piotrowicz: done 👍

packages/cloudflare/src/cli/build/index.ts
dario-piotrowicz: done 👍  (a28f9976a7c6344ad4788bdb7ca91fa5e9eb5ed5)

packages/cloudflare/src/cli/build/index.ts
dario-piotrowicz: Updated 👍

.changeset/lazy-balloons-report.md
```diff
@@ -0,0 +1,9 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+checks and creates a wrangler.jsonc file for the user in case one wrangler.(toml|json|jsonc) file is not already present
+
+also introduce a new `--skipWranglerConfigCheck` cli flag and a `SKIP_WRANGLER_CONFIG_CHECK`
```
dario-piotrowicz, line 7: I think that this makes sense, generally I expect the vast majority of people to just use standard wrangler config files, those that however do not want to do that and want to instead use a custom config location should have a way to opt out of the config check

vicb (approved):
Thanks!

packages/cloudflare/src/cli/build/index.ts
dario-piotrowicz: I've copied this minor cleanup to the original `workers-sdk` codebase: https://github.com/cloudflare/workers-sdk/pull/7658

PR #228, opened 2025-01-06
fix: exclude `.env.local` files for `test` mode

james-elicx:
Align the behavior with [what NextJs does](https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#test-environment-variables)


vicb (approved):
Thanks!

PR #234, opened 2025-01-08
fix: host not included in route handler urls

### LINKED ISSUE #210: [BUG] Route Handler request always return "n" instead of the worker's origin
### Describe the bug

```
export async function GET(
  request: NextRequest,
  { params }: { params: Promise<{ provider: string }> }
) {
  const url = new URL(request.nextUrl);
  const supabase = await createClient();
  const provider = (await params).provider as Provider;

  console.log("raw url: " + request.nextUrl);
  console.log("url: " + url);
  const { data, error } = await supabase.auth.signInWithOAuth({
    provider,
    options: {
      redirectTo: `${url.origin}/auth/callback`,
    },
  });

  console.log(`${url.origin}/auth/callback`);

  if (error || !data || !data.url) {
    return Response.json({ msg: "Failed to sign in" }, { status: 500 });
  }

  return redirect(data.url);
}

```
image

request.nextUrl always return "https://n/" instead of my worker custom domain.

### Steps to reproduce

Create a route handler and call request.nextUrl

### Expected behavior

It should return the correct url

0.3.2

### Wrangler version

3.99.0

### next info output

```shell
Operating System:
  Platform: linux
  Arch: x64
  Version: #1 SMP Tue Nov 5 00:21:55 UTC 2024
  Available memory (MB): 7782
  Available CPU cores: 32
Binaries:
  Node: 20.14.0
  npm: 10.9.2
  Yarn: N/A
  pnpm: 9.4.0
Relevant Packages:
  next: 15.1.3 // Latest available version is detected (15.1.3).
  eslint-config-next: 15.1.3
  react: 19.0.0
  react-dom: 19.0.0
  typescript: 5.7.2
Next.js Config:
  output: N/A
```

### Additional context

_No response_

james-elicx:
fixes #210

Research into what happens internally in Next.js is available in https://github.com/opennextjs/opennextjs-cloudflare/issues/210#issuecomment-2573260525.

OpenNext is removing the `trustHostHeader: true` option from the Next.js config if routing is disabled, and it considers routing disabled when `config.middleware.external` is truthy. This value is required to be true in the OpenNext config used for deploying to Cloudflare.

This PR changes the check to not strip it if there is external middleware.

### DIFF
--- a/examples/api/e2e/base.spec.ts
+++ b/examples/api/e2e/base.spec.ts
@@ -35,3 +35,10 @@ test("sets environment variables from the Next.js env file", async ({ page }) =>
   const res = await page.request.get("/api/env");
   await expect(res.json()).resolves.toEqual(expect.objectContaining({ TEST_ENV_VAR: "TEST_VALUE" }));
 });
+
+test("returns correct information about the request from a route handler", async ({ page }) => {
+  const res = await page.request.get("/api/request");
+  // Next.js can fall back to `localhost:3000` or `n` if it doesn't get the host - neither of these are expected.
+  const expectedURL = expect.stringMatching(/https?:\/\/localhost:(?!3000)\d+\/api\/request/);
+  await expect(res.json()).resolves.toEqual({ nextUrl: expectedURL, url: expectedURL });
+});
--- a/packages/cloudflare/src/cli/build/open-next/createServerBundle.ts
+++ b/packages/cloudflare/src/cli/build/open-next/createServerBundle.ts
@@ -182,8 +182,7 @@ async function generateBundle(
       target: /core(\/|\\)util\.js/g,
       deletes: [
         ...(disableNextPrebundledReact ? ["requireHooks"] : []),
-        ...(disableRouting ? ["trustHostHeader"] : []),
-        ...(!isBefore13413 ? ["requestHandlerHost"] : []),
+        ...(isBefore13413 ? ["trustHostHeader"] : ["requestHandlerHost"]),
         ...(isAfter141 ? ["experimentalIncrementalCacheHandler"] : ["stableIncrementalCache"]),
       ],
     }),
--- a/.changeset/short-cats-kneel.md
+++ b/.changeset/short-cats-kneel.md
@@ -1,0 +1,7 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+fix: host not included in route handler urls
+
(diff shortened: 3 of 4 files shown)


examples/api/e2e/base.spec.ts
```diff
@@ -35,3 +35,10 @@ test("sets environment variables from the Next.js env file", async ({ page }) =>
   const res = await page.request.get("/api/env");
   await expect(res.json()).resolves.toEqual(expect.objectContaining({ TEST_ENV_VAR: "TEST_VALUE" }));
 });
+
+test("returns correct information about the request from a route handler", async ({ page }) => {
+  const res = await page.request.get("/api/request");
+  // Next.js can fall back to `localhost:3000` or `n` if it doesn't get the host - neither of these are expected.
+  const expectedURL = expect.stringMatching(/https:\/\/localhost:(?!3000)\d+\/api\/request/);
```
james-elicx, line 42: fyi, couldn't use page.url() from playwright here as it was always returning about:blank

vicb:
@conico974 could you please take a look when you have time? Thanks!

vicb:
> OpenNext is removing the trustHostHeader: true option from the Next.js config if routing is disabled, and it considers routing disabled when config.middleware.external is truthy

I was also thinking that we might want to switch to not using external middleware at least until we have a better story for multi worker app.

packages/cloudflare/src/cli/build/open-next/createServerBundle.ts
```diff
@@ -182,8 +182,7 @@ async function generateBundle(
       target: /core(\/|\\)util\.js/g,
       deletes: [
         ...(disableNextPrebundledReact ? ["requireHooks"] : []),
-        ...(disableRouting ? ["trustHostHeader"] : []),
-        ...(!isBefore13413 ? ["requestHandlerHost"] : []),
+        ...(isBefore13413 ? ["trustHostHeader"] : ["requestHandlerHost"]),
```
conico974, line 185: Next probably won't work for version below 13.5.1 in cloudflare anyway. For Next <13.4.13, we don't provide any routing at all and rely entirely on NextServer itself (with the fake edge runtime for the middleware and all the filesystem related call in Next internal routing). For Next <13.5.1 and >13.4.1, we override `require` to provide the correct version of react depending on if app or page router is used, and i'm pretty sure it won't work as is in cloudflare.

It's probably better to throw or at least show a big warning message for these versions.

To be honest, i'm not sure it makes sense in aws to remove `trustHostHeader` as well unless we are before 13.4.13

vicb (approved):
Thanks for the PR and adding a test!

PR #237, opened 2025-01-08
show error for next.js versions older than v14

james-elicx:
image

### DIFF
--- a/.changeset/stupid-geese-reply.md
+++ b/.changeset/stupid-geese-reply.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+show error on Next.js versions older than v14
--- a/packages/cloudflare/src/cli/build/index.ts
+++ b/packages/cloudflare/src/cli/build/index.ts
@@ -55,7 +55,7 @@ export async function build(projectOpts: ProjectOptions): Promise<void> {
   buildHelper.checkRunningInsideNextjsApp(options);
   logger.info(`App directory: ${options.appPath}`);
   buildHelper.printNextjsVersion(options);
-  showErrorForOldNextjsVersions(options);
+  ensureNextjsVersionSupported(options);
   buildHelper.printOpenNextVersion(options);
 
   if (projectOpts.skipNextBuild) {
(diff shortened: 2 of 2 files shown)


packages/cloudflare/src/cli/build/index.ts
```diff
@@ -276,3 +277,10 @@ export async function getLatestCompatDate(): Promise<string | undefined> {
     /* empty */
   }
 }
+
+function showErrorForOldNextjsVersions(options: buildHelper.BuildOptions) {
+  if (buildHelper.compareSemver(options.nextVersion, "14.0.0") < 0) {
+    logger.error("Next.js version unsupported, please upgrade to version 14 or greater.");
+    process.exit(1);
```
dario-piotrowicz, line 284: Something I've picked up from @petebacondarwin, I usually prefer `throw`s  instead of hard `process.exit` calls so that we still have the possibility in the future if needed for whatever reason to catch it

packages/cloudflare/src/cli/build/index.ts
```diff
@@ -276,3 +277,10 @@ export async function getLatestCompatDate(): Promise<string | undefined> {
     /* empty */
   }
 }
+
+function showErrorForOldNextjsVersions(options: buildHelper.BuildOptions) {
```
dario-piotrowicz, line 281: this function name doesn't really convey the fact that this function exits the build process (by the name alone I would assume that this function simply logs the error)

I would maybe call it something like:
```
crushIfOnOldNextjsVersion
```
or something like that

If you where to throw as I suggested in my other comment it could also be:
```
warnAndThrowOnOldNextjsVersions
```
or something like that

.changeset/stupid-geese-reply.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+show error on Next.js versions older than v14
```
dario-piotrowicz, line 5: sorry if it's a silly question, is there a reason why v14 specifically? (I think it's ok but I am just curious if there is something specific that breaks us in <= 13)

dario-piotrowicz (commented):
Looks good to me

.changeset/stupid-geese-reply.md
vicb: See https://github.com/opennextjs/opennextjs-cloudflare/pull/234#discussion_r1907320812 and that's something we already discussed with @conico974

packages/cloudflare/src/cli/build/index.ts
vicb: +1 for a better name, my take: `ensureNextjsVersionSupported`

packages/cloudflare/src/cli/build/index.ts
vicb: I'm fine with exiting.
We can change later if needed.

vicb (approved):
Thanks,
I added some minor comments

.changeset/stupid-geese-reply.md
dario-piotrowicz: great, thanks 🙂

PR #241, opened 2025-01-09
chore: move rimraf to devDependencies

vicb:


james-elicx (approved):

vicb:
Thanks James. I forgot to update the lock file. Tests should be happier/greenier now

Issue #239, opened 2025-01-09
[FEATURE] Allow specifying environment for calls to getCloudflareContext in next dev mode


jmif:
### Is your feature request related to a problem?

`getCloudflareContext` doesn't support environments when when running locally in next dev mode.  I have two environments specified in my wrangler.toml (development and production) and each of those has different configuration values (different KV bindings and different env vars).  When I run `next dev` the env returned from `getCloudflareContext` is empty.  It works when running `wrangler dev --env development`.

### Describe the solution you'd like

I think we need to support supplying the environment [here](https://github.com/opennextjs/opennextjs-cloudflare/blob/main/packages/cloudflare/src/api/get-cloudflare-context.ts#L74).  I tested this by changing the line to:

```
    const { env, cf, ctx } = await getPlatformProxy({ environment: 'development' });
```

And my bindings were exposed.

Was going to submit a PR for this, but it's not exactly clear to me the best way to approach this.  It doesn't feel right to expose environment on the `getCloudflareContext` call as it's only really relevant in `next dev`.   I'm wondering if some kind of process.env or global variable should be supported?

### Describe alternatives you've considered

I can run wrangler dev but it runs a prod build so the DX isn't as nice as `next dev`

### @opennextjs/cloudflare version

latest

### Before submitting

- [X] I have checked that there isn't already a similar feature request
- [X] This is a single feature (not multiple features in one request)

vicb:
Sounds good, do you want to submit a PR?

PR #243, opened 2025-01-09
Add support for wrangler environment selection in next dev mode

jmif:
I opted for an environment variable as it was the closest to the `--env` flag that you'd use with wrangler.  I also considered exposing it as via an API method and saving the result in a module level variable as well as a global variable, but I figured keeping this configuration as close to how you'd do in wrangler made the most sense.  Happy to discuss though!

Addresses #239

### DIFF
--- a/packages/cloudflare/src/api/get-cloudflare-context.ts
+++ b/packages/cloudflare/src/api/get-cloudflare-context.ts
@@ -71,7 +71,10 @@ async function getCloudflareContextInNextDev<
     const { getPlatformProxy } = await import(
       /* webpackIgnore: true */ `${"__wrangler".replaceAll("_", "")}`
     );
-    const { env, cf, ctx } = await getPlatformProxy();
+    const { env, cf, ctx } = await getPlatformProxy({
+      // This allows the selection of a wrangler environment while running in next dev mode
+      environment: process.env.NEXT_DEV_WRANGLER_ENV,
+    });
     global[cloudflareContextInNextDevSymbol] = {
       env,
       cf: cf as unknown as CfProperties,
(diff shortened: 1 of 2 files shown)


packages/cloudflare/src/api/get-cloudflare-context.ts
```diff
@@ -71,7 +71,10 @@ async function getCloudflareContextInNextDev<
     const { getPlatformProxy } = await import(
       /* webpackIgnore: true */ `${"__wrangler".replaceAll("_", "")}`
     );
-    const { env, cf, ctx } = await getPlatformProxy();
+    const { env, cf, ctx } = await getPlatformProxy({
+      // This allows the selection of a wrangler environment while running in next dev mode
+      environment: process.env.CF_WRANGLER_ENV,
```
vicb, line 76:
```suggestion
      environment: process.env.NEXT_DEV_WRANGLER_ENV,
```

Unfortunately `.env` files are also loaded by `wrangler` (we have a collision here) so using a prefix specific to NEXT can help (See existing [env vars](https://github.com/cloudflare/workers-sdk/blob/f6132761c80d90c3521c93d48d0f0aed62bb360a/packages/wrangler/src/environment-variables/factory.ts#L1) used by wrangler)

vicb (changes requested):
Thanks for implementing this.

Could you please add a patch changeset (`pnpm changeset`) so that we have an entry for this change in the changelog - make sure to document the env var name there?

🙏

packages/cloudflare/src/api/get-cloudflare-context.ts
jmif: Makes sense - updated

vicb (approved):

vicb:
Thanks 🙏

PR #258, opened 2025-01-15
refactor: shuffle code around

vicb:
- Move utils to the utils folder to keep the main files (build/bundle) small and focused
- rename index.ts -> build.ts, I see no point in having index.ts and they make code navigation harder. If we should have index.ts I think they should only re-export not contain any other code.


dario-piotrowicz:
> rename index.ts -> build.ts, I see no point in having index.ts and they make code navigation harder. If we should have index.ts I think they should only re-export not contain any other code.

Regarding `index.ts` files in general I was the one advocating for those as I think that they generally make the code cleaner, however after the changes in https://github.com/opennextjs/opennextjs-cloudflare/pull/192 and the bundle resolution removal I don't really think that they make the code any cleaner at all (actually the oppositve), so I'm actually in favour of not using index.ts anymore at all if you want 🙂

dario-piotrowicz (approved):

vicb:
Thanks for the review!

> so I'm actually in favour of not using index.ts anymore at all if you want

I don't mind if they only re-export. _(The thing that I want is to go to `build.ts` to see the build related code and not having to pick amongst 5 different `index.ts`)_

Issue #14, opened 2024-09-20
🐛 Bug:  everything that's needed should be in the output directory


dario-piotrowicz:
The [build](https://github.com/flarelabs-net/poc-next/tree/main/builder) does output a worker in the specified [output directory](https://github.com/dario-piotrowicz/poc-build-nextjs-app-for-cf-workers/blob/69fb9be75670f82f13c7fdaf07b2dbd8ea1c7a51/nextjs-worker-builder/src/build/index.ts#L23) (defaulting to `.next-worker`).

This directory should contain everything that's required to run the worker (assets included).

But this is not so, as there are clear signs that files within `.next-worker` make use of files outside of the directory (making the dev/deployment of unbundled workers (`--no-bundle`) impossible).

For example:
 - the need to [patch the webpack-runtime.js file](https://github.com/flarelabs-net/poc-next/blob/0d304fdad460c174ca80dc99408daaf58f936663/builder/src/build/build-worker.ts#L40)
 - the need to [patch the `pages.runtime.prod.js` file](https://github.com/flarelabs-net/poc-next/blob/545428066cf3f30b7839f2d2355690bdb5456c12/builder/src/build/patches/to-investigate/wrangler-deps.ts#L10-L25)

I'm not sure how those files are not getting bundled by the [esbuild build step](https://github.com/dario-piotrowicz/poc-build-nextjs-app-for-cf-workers/blob/69fb9be75670f82f13c7fdaf07b2dbd8ea1c7a51/nextjs-worker-builder/src/build/build-worker/index.ts#L69) 😕

We should make sure that they are actually bundled in, or if that's not possible, everything that the worker requires should just need to live in the output directory and no external file/dependency should be needed (in other words, I think that we need to make sure that `wrangler dev --no-bundle` works).

vicb:
@dario-piotrowicz I took a quick look at this and have some comments:

> I'm not sure how those files are not getting bundled by the esbuild build step

At least part of why some file are not bundled by this pass is that they are not required.
You can see that [patchRequire](https://github.com/opennextjs/opennextjs-cloudflare/blob/baf8ff425040b1ad3c2dfa2ec2b48b4f1e3dc42b/packages/cloudflare/src/cli/build/bundle-server.ts#L147) add `require`s after this first pass.

That is not the only patch adding deps that will be pick up by `wrangler dev`

> I think that we need to make sure that wrangler dev --no-bundle works

I think this is no more applicable.
[...]

PR #259, opened 2025-01-15
instead of patching the `tracer.js` file to throw on `@opentelemetry/api` imports, delete the `@opentelemetry/api` dependency itself

dario-piotrowicz:
the problem that this addresses is that the `@opentelemetry/api` package is not only imported by the `tracer.js` file
we patch, so just deleting the library itself makes sure that all files requiring it get the same throwing behavior
(besides decreasing the overall worker size)

### DIFF
--- a/.changeset/forty-jobs-press.md
+++ b/.changeset/forty-jobs-press.md
@@ -1,0 +1,9 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+instead of patching the `tracer.js` file to throw on `@opentelemetry/api` imports, delete the `@opentelemetry/api` dependency itself
+
+the problem that this addresses is that the `@opentelemetry/api` package is not only imported by the `tracer.js` file
+we patch, so just deleting the library itself makes sure that all files requiring it get the same throwing behavior
+(besides decreasing the overall worker size)
--- a/packages/cloudflare/src/cli/build/open-next/createServerBundle.ts
+++ b/packages/cloudflare/src/cli/build/open-next/createServerBundle.ts
@@ -17,6 +17,7 @@ import { openNextReplacementPlugin } from "@opennextjs/aws/plugins/replacement.j
 import { openNextResolvePlugin } from "@opennextjs/aws/plugins/resolve.js";
 import type { FunctionOptions, SplittedFunctionOptions } from "@opennextjs/aws/types/open-next.js";
 
+import { deleteOpenTelemetryDep } from "../patches/index.js";
 import { normalizePath } from "../utils/index.js";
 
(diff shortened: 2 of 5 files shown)


dario-piotrowicz:
In the apps I've tested my solution worked just fine, in the [create-next-app example](https://github.com/opennextjs/opennextjs-cloudflare/tree/main/examples/create-next-app) it's instead failing 😓 
(I've tested my solution with `npm` maybe `pnpm` causes this, I need to double check that)

Anyways the issue this [code](https://github.com/vercel/next.js/blob/9e8266a7850cd2bc7dba3060465c5d0fc15ff07a/packages/next/src/server/lib/trace/tracer.ts#L25) usually the problematic require gets tree shaken away:
Screenshot 2025-01-15 at 22 08 43

That doesn't seem to happen in the app's case though 😓 
Screenshot 2025-01-15 at 22 07 52
(I patched the node_modules file with a console log there just to get a worker outputted via `wrangler deploy --dry-run --outdir`, the original code is the one inside the console log)

I think this is extra problematic because as you can see, the file is inside the `.next` directory, outside of our `.open-next` output directory, reaffirming that we're still got some issues with the output directory not containing everything as mentioned in https://github.com/opennextjs/opennextjs-cloudflare/issues/14 😓

dario-piotrowicz:
Given my comment above I am a bit of at a loss... if the tracer is outside of our output directory (`.open-next`) I think that there isn't much we can do about it... (I don't think we should touch/patch any files outside of our output directory)

It feels like for the time being (until we properly solve #14) we have to ask people to use wrangler aliases to make `@opentelemetry/api` requires throw...

Unless someone has some idea? @vicb, @petebacondarwin, @james-elicx? 😖

vicb:
Not looked at it deeply but I was also wondering about

> (The apps were using npm but the create-next-app is under our pnpm monorepo, I wonder if that could causes this)

Maybe with pnpm you will delete a link, resulting in the dep being deleted in the root `node_modules` while with npm you will only delete the local `node_modules` and the dep can still be resolved from the root.

dario-piotrowicz:
> while with npm you will only delete the local `node_modules` and the dep can still be resolved from the root.

Nono the dep is not resolved, I'm sure of it, I've debugged it multiple times and also that would break the applications (since the non-precompiled dep errors in workerd).

I do think that the dep deleting is correct, at least, if everything the worker needs (including `node_modules`) is supposed to be in `.open-next` then I do think that removing the dep there is the correct solution (since that's exactly what Next.js is checking: "is the opentelemetry dep available? if not use our precompiled version")

Issue #255, opened 2025-01-15
e2e failing locally on chromium & webkit


vicb:
Some e2e are failing locally at sha b397635

Testing on a MBP

<img width="1094" alt="image" src="https://github.com/user-attachments/assets/0362cdc1-cf4d-4adb-a2de-18f8e1102405" />

PR #265, opened 2025-01-17
introduce new `initOpenNextCloudflareForDev` utility and make `getCloudflareContext` synchronous

### LINKED ISSUE #137: [FEATURE] Synchronous getCloudflareContext()

The `getCloudflareContext()` utility cannot be used in certain places as it is async function, and therefore requires awaiting. Class constructors and global variables often depend on the expectation of being synchronous, and therefore are prevented from leveraging the utility for bindings, and have to use process.env instead.

`getCloudflareContext()` should be able to be used synchronously.

N/A.

0.2.1

### Additional context

The next-on-pages approach to this problem involved a way to setup the environment in the Next.js config file, and then interacting with a global symbol on process.env in dev mode.

https://github.com/cloudflare/next-on-pages/tree/main/internal-packages/next-dev

### Before submitting

- [X] I have checked that there isn't already a similar feature request
- [X] This is a single feature (not multiple features in one request)

### LINKED ISSUE #226: [BUG] TypeError: A dynamic import callback was not specified when using getCloudflareContext
### Describe the bug

I am using a service binding in my middleware to let another worker handle authentication. I have configured this as following and this works properly when deployed and when running in preview mode locally.

```
import { Fetcher } from "@cloudflare/workers-types";

declare module "next/server" {
  interface NextRequest {
    apiClient: ApiClient;
  }
}

export interface Env {
  API: Fetcher;
}

function createApiClient(env: Env) {
  // fetch client using env.API
}

export async function middleware(request: NextRequest) {
  const env = (await getCloudflareContext()).env as Env;
  const apiClient = await createApiClient(env);
  request.apiClient = apiClient;
  
  // Redacted for brevity: check using api client whether user's token is valid. If not return 401
  
  return NextResponse.next();
}
```

But when in `next dev` this throws
```
TypeError: A dynamic import callback was not specified.
```
image

### Steps to reproduce

1. Use getCloudflareContext in middleware.ts
2. Run app with `next dev`
3. Refresh the page

### Expected behavior

No error to be thrown

experimental

### Wrangler version

3.99.0

### next info output

```shell
pnpm next info

Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 23.0.0: Fri Sep 15 14:41:43 PDT 2023; root:xnu-10002.1.13~1/RELEASE_ARM64_T6000
  Available memory (MB): 65536
  Available CPU cores: 10
Binaries:
  Node: 22.12.0
  npm: 10.9.0
  Yarn: N/A
  pnpm: 9.15.0
Relevant Packages:
  next: 14.2.5 // An outdated version detected (latest is 15.1.3), upgrade is highly recommended!
  eslint-config-next: 15.0.3
  react: 18.3.1
  react-dom: 18.3.1
  typescript: 5.7.2
```
```


### Additional context

_No response_

### PR DESCRIPTION
**Pull request by dario-piotrowicz**
This PR introduces a new `initOpenNextCloudflareForDev` utility to add to the Next.js config file that makes the `getCloudflareContext` work in middlewares (or more generally, the edge runtime) during local development (via `next dev`)

It `getCloudflareContext` has also been converted to be synchronous

For more details see the changeset file: [`.changeset/chilly-dryers-begin.md`](https://github.com/opennextjs/opennextjs-cloudflare/pull/265/files#diff-75dff73666acc35e672da1f4ed34bef1d39ca6a3ca72ce956ed09ac978ea7ed4)
___

C3 PR: https://github.com/cloudflare/workers-sdk/pull/7903

Docs relative PR: https://github.com/opennextjs/docs/pull/57

___

fixes #137
fixes #226

### CODE REVIEW

File: examples/middleware/e2e/playwright.config.ts [hunk 1]
```diff
@@ -49,5 +49,7 @@ export default defineConfig({
     command: "pnpm preview:worker",
     url: "http://localhost:8774",
     reuseExistingServer: !process.env.CI,
+    // the app takes a bit to boot up for some reason, that's why we need a longer timeout here
+    timeout: 2 * 60 * 1000,
```
dario-piotrowicz, line 53: not too sure why this is needed now and it wasn't before 🤷

dario-piotrowicz: ah.... probably because the boot up process takes a bit longer with `enableEdgeDevGetCloudflareContext`... weird the difference should be practically negligible (locally I don't notice any difference)

examples/middleware/e2e/base.spec.ts
```diff
@@ -3,28 +3,28 @@ import { test, expect } from "@playwright/test";
 test("redirect", async ({ page }) => {
```
vicb, line 3: This looks like a fix for https://github.com/opennextjs/opennextjs-cloudflare/issues/255

Would you mind extracting the code in a distinct PR for easier tracking?

packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -0,0 +1,149 @@
...
+  ).default.createRequire(import.meta.url);
+
+  // eslint-disable-next-line unicorn/prefer-node-protocol -- the `next dev` compiler doesn't accept the node prefix
+  const vmModule = require("vm");
+
+  const originalRunInContext = vmModule.runInContext.bind(vmModule);
+
+  const context = await getCloudflareContextFromWrangler();
+
+  vmModule.runInContext = (...args: [string, RuntimeContext, ...unknown[]]) => {
```
vicb, line 127: Extract this to a distinct function to add comments?
ref: https://github.com/cloudflare/next-on-pages/blob/2cd4c3c704a00e6b693229f1f14102abc6318d11/internal-packages/next-dev/src/shared.ts#L22

examples/middleware/next.config.mjs
```diff
@@ -1,4 +1,8 @@
+import { enableEdgeDevGetCloudflareContext } from "@opennextjs/cloudflare";
+
 /** @type {import('next').NextConfig} */
 const nextConfig = {};
 
+enableEdgeDevGetCloudflareContext();
```
vicb, line 6: What about renaming this function to `initOpenNextCloudflare()`

I think it's easier for the user to understand as we expose lesst techincal details and we could add to this function later if needed.

packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -0,0 +1,149 @@
...
+/**
+ * Utility to get the current Cloudflare context
+ *
+ * @returns the cloudflare context
+ */
+export async function getCloudflareContext<
+  CfProperties extends Record<string, unknown> = IncomingRequestCfProperties,
+  Context = ExecutionContext,
+>(): Promise<CloudflareContext<CfProperties, Context>> {
+  const global = globalThis as unknown as {
```
vicb, line 38: could we add the type to `declare global` higher up in this file instead?

packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -0,0 +1,149 @@
...
+ * Gets a local proxy version of the cloudflare context (created using `getPlatformProxy`) when
+ * running in the standard next dev server (via `next dev`)
+ *
+ * @returns the local proxy version of the cloudflare context
+ */
+async function getCloudflareContextInNextDev<
+  CfProperties extends Record<string, unknown> = IncomingRequestCfProperties,
+  Context = ExecutionContext,
+>(): Promise<CloudflareContext<CfProperties, Context>> {
+  const global = globalThis as unknown as {
```
vicb, line 65: same comment as above for the type

packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -0,0 +1,149 @@
...
+    [cloudflareContextInNextDevSymbol]: CloudflareContext<CfProperties, Context> | undefined;
+  };
+
+  if (!global[cloudflareContextInNextDevSymbol]) {
+    try {
+      const cloudflareContext = await getCloudflareContextFromWrangler<CfProperties, Context>();
+      global[cloudflareContextInNextDevSymbol] = cloudflareContext;
+    } catch (e: unknown) {
+      if (e instanceof Error && e.message.includes("A dynamic import callback was not specified")) {
```
vicb, line 74: Is this to detect the edge runtime?
I think NEXT defines a `process.env.RUNTIME` (not sure about the exact name) and it could be clearer if we can use it?

packages/cloudflare/src/api/cloudflare-context.ts
vicb, line 127: wrt `...args: [string, RuntimeContext, ...unknown[]]`

doesn't `runInContext` only have 3 args? In such a case it would be clearer to spell them out and avoid the `args[1]`. If it can have more, you can use ...args for the more part

vicb (changes requested):

vicb:
Thought: Maybe we can `getPlatformProxy(...)` in `initOpenNextCloudflare()` so that `getCloudflareContext` could be made sync?

[This comment](https://github.com/cloudflare/next-on-pages/blob/2cd4c3c704a00e6b693229f1f14102abc6318d11/internal-packages/next-dev/src/index.ts#L10) says "the function is an async one but it doesn't need to be awaited", it might be nice to add the rationale if we add the implementation to our repo.

packages/cloudflare/src/api/cloudflare-context.ts
dario-piotrowicz: I am not sure if you can extend the global declaration with a value indexed by a symbol, a bunch of googling lets me think that it's not possible 😕

vicb: Too bad :(

We could either switch to a string or left as is then

vicb: Another option:

```ts
const mySymbol = Symbol.for('internal.fake');
type MyCustomGlobal = typeof globalThis & {
  [mySymbol]: { loader: true}
}

(globalThis as MyCustomGlobal)[mySymbol] = { loader: true};
```

examples/middleware/next.config.mjs
dario-piotrowicz: mh... my idea was that people would not have to always include this but only when they face the cloudflare context in the edge runtime issue (a problem that can happen to a subset of users and one that hopefully can go away in the future if Next.js decides to support nodejs middlewares), that's why I made the name more specific

`initOpenNextCloudflare()` sounds like something they always need to have (similar to [setupDevPlatform](https://github.com/cloudflare/next-on-pages/blob/2cd4c3c704a00e6b693229f1f14102abc6318d11/internal-packages/next-dev/src/index.ts#L14) in next-on-pages)

> we could add to this function later if needed.

That's true, but again your comment makes it sound to me like moving forward we'd want to recommend all users to include this, which I am not sure about 😕

So based on the above I think I do prefer something like `enableEdgeDevGetCloudflareContext` which makes it very clear to the user what this function does and when/why is needed besides making clear(er) that this is optional

But if you strongly prefer a generic name like `initOpenNextCloudflare()` I'd be ok with using that (although that's not my preference)

examples/middleware/e2e/base.spec.ts
dario-piotrowicz: yes of course: https://github.com/opennextjs/opennextjs-cloudflare/pull/267, sorry I didn't realise there was an issue for this 👍

dario-piotrowicz: (once we merge #267 I can simply rebase this PR)

packages/cloudflare/src/api/cloudflare-context.ts
dario-piotrowicz: > We could either switch to a string

I could really prefer not to do that because if we were to use a string and add it to the global declaration then that would leak in the user types, so with autocompletion etc... they would see such a field in their global scope (besides the fact that they could easily access it)

dario-piotrowicz: > Another option:
> 
> ```ts
> const mySymbol = Symbol.for('internal.fake');
> type MyCustomGlobal = typeof globalThis & {
>   [mySymbol]: { loader: true}
> }
> 
> (globalThis as MyCustomGlobal)[mySymbol] = { loader: true};
> ```

I'm happy to go with that if you prefer, but to me looks more verbose than the current one, and I don't think it makes things clearer (a type defined inside a function is not super clear/readable in my opinion):
Screenshot 2025-01-20 at 10 54 09

packages/cloudflare/src/api/cloudflare-context.ts
dario-piotrowicz: > Is this to detect the edge runtime?
> I think NEXT defines a process.env.RUNTIME (not sure about the exact name) and it could be clearer if we can use it?

Yes I used this try-catch to detect the edge runtime, since with debugging I couldn't see any global flag that indicated what runtime is being used, I did not think of checking values on process.env, good call, thanks 🙂, I've updated the code 👍

dario-piotrowicz:
> Thought: Maybe we can `getPlatformProxy(...)` in `initOpenNextCloudflare()` so that `getCloudflareContext` could be made sync?

Yes, we could, this would mean that everyone would always have to call `initOpenNextCloudflare()`, I mentioned in [my comment](https://github.com/opennextjs/opennextjs-cloudflare/pull/265#discussion_r1922175232) that we don't need to go that route as not forcing users to do that might provide a better DX

but again if you strongly prefer us to go that route I'm ok with it

> [This comment](https://github.com/cloudflare/next-on-pages/blob/2cd4c3c704a00e6b693229f1f14102abc6318d11/internal-packages/next-dev/src/index.ts#L10) says "the function is an async one but it doesn't need to be awaited", it might be nice to add the rationale if we add the implementation to our repo.

Yes, given the implementation of `next dev`, it takes some time from when Next.js loads the config and when it serves the app, we did see that that amount of time is enough for the proxy to be ready in time. It would require very little changes here, just removing the on-the-fly proxy generation from `getCloudflareContext` and changing the function signature to be sync

packages/cloudflare/src/api/cloudflare-context.ts
dario-piotrowicz: good point 👍

dario-piotrowicz: code updated 👍

dario-piotrowicz:
@vicb I've addressed all the feedback, the only thing left is to decide whether we want to force this sort of thing via a `initOpenNextCloudflare()` and potentially convert `getCloudflareContext` to be sync or not

up to you, just let me know if you want to me make the change 👍

packages/cloudflare/src/api/cloudflare-context.ts
vicb: Very good point for string.

The idea `typeof globalThis & {...}` would be to have this close to `declare global {...}` at the top of the file.
But then you would probably have a typing issue... which might be fixed with a function... so probably not worth it

vicb:
> @vicb I've addressed all the feedback, the only thing left is to decide whether we want to force this sort of thing via a `initOpenNextCloudflare()` and potentially convert `getCloudflareContext` to be sync or not
> 
> up to you, just let me know if you want to me make the change 👍 (it doesn't really take much effort to do code-wise, it's just a matter of DX)

Give me some time to think about it, I'm not settled.

Some thoughts I have for now about `initOpenNextCloudflare()`

##Cons:##

`initOpenNextCloudflare()` always need to be called but we can have an error message when it's not called

##Pros##

- I think it's easier to document (i.e. "always call `initOpenNextCloudflare()`" vs " call enableEdgeDevGetCloudflareContext if you call getConflareContext from a middleware"). I think Vercel is working on the ability to run middlewares on the node runtime and we will add support for the edge runtime at some point. It will make it harder to explain
- I think it's easier for user to understand (They don't have to undestand what "EdgeDev" is)
- We could setup the cloudflare context in there so that `getCloudflareContext()` becomes sync. A lot of user are complaining about that on Discord and GH - See how it impacts us in the [KVCache](https://github.com/opennextjs/opennextjs-cloudflare/blob/0b54cf4c1f5850234dda206d75305761c17d32c8/packages/cloudflare/src/api/kvCache.ts#L144-L149)
- Calling the init on a per need basis (vs always calling `initOpenNextCloudflare()`) could introduce unrelated test failures - i.e. how you had to bump the timeout in this PR

packages/cloudflare/src/api/cloudflare-context.ts
dario-piotrowicz: as above, I think we agreed that it's not really that worth looking into changing this

dario-piotrowicz:
> Give me some time to think about it, I'm not settled.
> 
> Some thoughts I have for now about `initOpenNextCloudflare()`
> 
> ...

Well there is only one con and many pros, I dislike that we need to then ask users to always call the function, but given the benefits it does seem like it might be worth it

By the way, either way I am not too convinced about the name since this is a dev (`next dev`) only thing and in my opinion  extremely unlikely to be useful outside of dev (because we have other places where we could accept options/tweak things, like in our templates, in the open-next config, etc...) for this reason I would really prefer it if the function had `dev` somewhere in its name (e.g. `initOpenNextCloudflareForDev()`)

examples/middleware/next.config.mjs
dario-piotrowicz: Resolving as this has been addressed, although I went with the name `initOpenNextCloudflareForDev` 😄 

I'm happy to reconsider if someone thinks that it is too verbose/long

.changeset/chilly-dryers-begin.md
```diff
@@ -0,0 +1,47 @@
...
+
+```js
+// next.config.mjs
+
+import { initOpenNextCloudflareForDev } from "@opennextjs/cloudflare";
+
+/** @type {import('next').NextConfig} */
+const nextConfig = {};
+
+initOpenNextCloudflareForDev();
```
vicb, line 22: init first ?

```suggestion
initOpenNextCloudflareForDev();

/** @type {import('next').NextConfig} */
const nextConfig = {};
```

.changeset/chilly-dryers-begin.md
```diff
@@ -0,0 +1,47 @@
...
+thanks to this new utility we are also able to make `getCloudflareContext` synchronous, so `await`ing such function is no longer necessary
+
+Example:
+
+- Before:
+
+  ```js
+  import { getCloudflareContext } from "@opennextjs/cloudflare";
+
+  const cloudflareContext = await getCloudflareContext();
```
vicb, line 36: We still should not call this at top level as we are not in a request context

examples/api/e2e/playwright.config.ts
```diff
@@ -49,5 +49,8 @@ export default defineConfig({
     command: "pnpm preview:worker",
     url: "http://localhost:8770",
     reuseExistingServer: !process.env.CI,
+    // the app uses the `initOpenNextCloudflareForDev` which in CI apparently makes the boot up take slightly longer,
+    // that's why we need a longer timeout here (we just add 10 seconds to the default 60)
+    timeout: !process.env.CI ? undefined : 60 * 1000 + 10 * 1000,
```
vicb, line 54: nit:

```suggestion
    timeout: 70_000,
```

examples/middleware/next.config.mjs
```diff
@@ -1,4 +1,8 @@
+import { initOpenNextCloudflareForDev } from "@opennextjs/cloudflare";
+
 /** @type {import('next').NextConfig} */
 const nextConfig = {};
 
+initOpenNextCloudflareForDev();
+
```
vicb, line 7:
```suggestion
initOpenNextCloudflareForDev();

/** @type {import('next').NextConfig} */
const nextConfig = {};

```

packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -0,0 +1,175 @@
...
+  const cloudflareContext = global[cloudflareContextSymbol];
+
+  if (!cloudflareContext) {
+    // the cloudflare context is initialized by the worker and is always present in production/preview
+    // during local development (`next dev`) it might be missing only if the developers hasn't called
+    // the `initOpenNextCloudflareForDev` function in their Next.js config file
+    const getContextFunctionName = getCloudflareContext.name;
+    const initFunctionName = initOpenNextCloudflareForDev.name;
+    throw new Error(
+      `\n\n\`${getContextFunctionName}\` has been called during development without having used` +
```
vicb, line 57:
```suggestion
      `\n\n\`${getContextFunctionName}\` has been called during development without having called` +
```

packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -0,0 +1,175 @@
...
+}
+
+/**
+ * Performs some initial setup to integrate as best as possible the local Next.js dev server (run via `next dev`)
+ * with the open-next Cloudflare adapter
+ *
+ * (Currently all the setup that this function performs is making sure that `getCloudflareContext` works as intended,
+ * in the future mode improvements might be added)
+ *
+ * Note: this function should only be called inside the Next.js config file
```
vicb, line 81: add not that it doesn't need to be `await`ed?

packages/cloudflare/src/api/kvCache.ts
```diff
@@ -2,7 +2,7 @@ import type { KVNamespace } from "@cloudflare/workers-types";
 import type { CacheValue, IncrementalCache, WithLastModified } from "@opennextjs/aws/types/overrides";
 import { IgnorableError, RecoverableError } from "@opennextjs/aws/utils/error.js";
 
-import { getCloudflareContext } from "./get-cloudflare-context.js";
+import { getCloudflareContext } from "./cloudflare-context.js";
```
vicb, line 5: in this file you can remove the following pattern:

```ts
    if (!this.initialized) {
      await this.init();
    }
```

That was only required because `getCloudflareContext` was async

vicb (changes requested):
Cant't wait for this PR to get in 🎉

packages/cloudflare/src/api/cloudflare-context.ts
vicb, line 57: Could this be a single ```...string...``` to avoid `\n`, `+`?

packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -0,0 +1,175 @@
...
+
+/**
+ * Adds the cloudflare context to the global scope in which the Next.js dev node.js process runs in, enabling
+ * future calls to `getCloudflareContext` to retrieve and return such context
+ *
+ * @param cloudflareContext the cloudflare context to add to the node.sj global scope
+ */
+function addCloudflareContextToNodejsGlobal(cloudflareContext: CloudflareContext<CfProperties, Context>) {
+  const global = getInternalGlobalThis<CfProperties, Context>();
+  global[cloudflareContextSymbol] = cloudflareContext;
```
vicb, line 99: When I see `getInternalGlobalThis(...)` I feel like the function is "doing" something. It is actually a noop and its only purpose it to cast. Can we just directly:

```suggestion
  const global = globalThis as InternalGlobalThis<CfProperties, Context>;
  global[cloudflareContextSymbol] = cloudflareContext;
```

_(global can even be inlined here)_

examples/middleware/e2e/playwright.dev.config.ts
```diff
@@ -0,0 +1,53 @@
+import { defineConfig, devices } from "@playwright/test";
+
+declare var process: { env: Record<string, string> };
```
vicb, line 3: import { type process } from 'node:process' ?

.changeset/chilly-dryers-begin.md
dario-piotrowicz: it's just an example, I didn't want to get too verbose...

would `// ...` lines solve the issue?
or would you prefer a more verbose example?
or maybe I can remove the example altogether since this is also pretty obvious anyways

packages/cloudflare/src/api/cloudflare-context.ts
dario-piotrowicz: we could use string literals but then we'd have to de-dent them, for example using a library like [`ts-dedent`](https://www.npmjs.com/package/ts-dedent) (like wrangler does for example) or by doing that manually (like with a regex replace)

personally I prefer `\n` + `+` (no extra dependency/brittleness to deal with and it is practically as readable), but if you want to use string literals + some de-dent functionality I'm ok with that too

packages/cloudflare/src/api/cloudflare-context.ts
dario-piotrowicz: yes I was thinking that maybe `getInternalGlobalThis` might be too much as well, I'm happy to remove it 👍

packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -0,0 +1,175 @@
...
+
+type InternalGlobalThis<
+  CfProperties extends Record<string, unknown> = IncomingRequestCfProperties,
+  Context = ExecutionContext,
+> = typeof globalThis & {
+  [cloudflareContextSymbol]: CloudflareContext<CfProperties, Context> | undefined;
+};
+
+/**
+ * Gets the globalThis for internal usage
```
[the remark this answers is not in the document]
dario-piotrowicz: done 👍

packages/cloudflare/src/api/cloudflare-context.ts
dario-piotrowicz: done 👍

packages/cloudflare/src/api/cloudflare-context.ts
dario-piotrowicz: added 👍

examples/middleware/e2e/cloudflare-context.spec.ts
```diff
@@ -0,0 +1,14 @@
...
+test("cloudflare context env object is populated", async ({ page }) => {
+  await page.goto("/middleware");
+  const cloudflareContextHeaderElement = page.getByTestId("cloudflare-context-header");
+  // Note: the text in the span is "keys of `cloudflareContext.env`: MY_VAR, MY_KV, ASSETS" for previews
+  //       and "keys of `cloudflareContext.env`: MY_VAR, MY_KV" in dev (`next dev`)
+  //       that's why we use `toContain` instead of `toEqual`, this is incorrect and the `ASSETS` binding
+  //       should ideally also be part of the dev cloudflare context
+  //       (this is an upstream wrangler issue: https://github.com/cloudflare/workers-sdk/issues/7812)
+  expect(await cloudflareContextHeaderElement.textContent()).toContain(
+    "keys of `cloudflareContext.env`: MY_VAR, MY_KV"
```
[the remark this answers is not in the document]
dario-piotrowicz: the dependence on the ordering can also easily be fixed by sorting the keys: https://github.com/opennextjs/opennextjs-cloudflare/pull/265/commits/6006cf24cd4d4046213636f08b33183778d9a9a0
(I also had to filter the variables to avoid the ASSETS issue)

I am happy to also add (or instead) an API route and do `array.toContain` whatever you prefer, I'm happy with anything 🙂

.changeset/chilly-dryers-begin.md
```diff
@@ -0,0 +1,47 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+introduce new `initOpenNextCloudflareForDev` utility and make `getCloudflareContext` synchronous
+
+introduce a new `initOpenNextCloudflareForDev` function that when called in the [Next.js config file](https://nextjs.org/docs/app/api-reference/config/next-config-js) integrates the Next.js dev server with the open-next Cloudflare adapter. Most noticeably this enables `getCloudflareContext` to work in
```
vicb, line 7: suggestion: maybe start simple with the TL;DR and add details (or not) in following paragraph.

something like

```
`initOpenNextCloudflareForDev()` must be called from the [Next.js config file](https://nextjs.org/docs/app/api-reference/config/next-config-js) so that your bindings can be accessed  using `getCloudflareContext()` when running the Next.js development server (`next dev`).

`getCloudflareContext` is now sync.

File: .changeset/chilly-dryers-begin.md (hunk 9 above)
Comment (vicb) [reply .changeset/chilly-dryers-begin.md:36]: I don't like the fact that the example is broken.

As the changeset is already long, I would rather shorten it and remove the example but add details to our documentation.

File: packages/cloudflare/src/api/kvCache.ts (hunk 14 above)
Comment (vicb) [reply packages/cloudflare/src/api/kvCache.ts:5]: I'll amend the PR

File: examples/middleware/e2e/cloudflare-context.spec.ts (hunk 18 above)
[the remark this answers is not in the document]
Comment (vicb) [reply examples/middleware/e2e/cloudflare-context.spec.ts]: I think API + `array.toContain` would be simpler but the code is already there and working so up to you.

File: packages/cloudflare/src/api/cloudflare-context.ts (hunk 12 above)
Comment (vicb) [reply packages/cloudflare/src/api/cloudflare-context.ts]: You could just:

```
      throw new Error(`

\`${getContextFu...
Another line

`);
```

**Comment by vicb** [comment]
I've pushed the `kvCache` update

**Comment by vicb** [comment]
Could you please update the README and opennext.js.org to tell users to add the init?

File: .changeset/chilly-dryers-begin.md (hunk 9 above)
Comment (dario-piotrowicz) [reply .changeset/chilly-dryers-begin.md:36]: example removed 👍

File: examples/create-next-app/e2e/playwright.config.ts [hunk 20]
```diff
@@ -49,8 +49,6 @@ export default defineConfig({
     command: "pnpm preview:worker",
     url: "http://localhost:8771",
     reuseExistingServer: !process.env.CI,
-    // the app uses the `initOpenNextCloudflareForDev` which in CI apparently makes the boot up take slightly longer,
-    // that's why we need a longer timeout here (we just add 10 seconds to the default 60)
-    timeout: !process.env.CI ? undefined : 60 * 1000 + 10 * 1000,
+    timeout: 7_000,
```
vicb, line 52: oops, should be 70_000... I'm not good at math ;)

.changeset/chilly-dryers-begin.md
dario-piotrowicz: the changeset title is already a good TL;DR, isn't it?

I think I don't fully understand what you're suggesting, if you want feel free to suggest a code change or to push on the branch

packages/cloudflare/src/api/kvCache.ts
dario-piotrowicz: ah ok I see what you meant, thanks for the update 👍

packages/cloudflare/src/api/cloudflare-context.ts
dario-piotrowicz: oh.... I really really dislike that! 🥲

(in my opinion that sort of thing completely breaks files indentation and looks just super messy)

.changeset/chilly-dryers-begin.md
vicb: I think the change set should first highlight what is most important for the users:

1. that the init function must be called from the config
2. that getCloudflareContext is sync

IMO mentioning the edge runtime of our recommendation could be pushed down or moved to the doc site.

packages/cloudflare/src/api/cloudflare-context.ts
vicb: keep you version then 👍

examples/api/e2e/playwright.config.ts
```diff
@@ -1,6 +1,7 @@
 import { defineConfig, devices } from "@playwright/test";
+import nodeProcess from "node:process";
```
vicb, line 2: Doesn't that work:
```suggestion
import type nodeProcess from "node:process";
```

dario-piotrowicz: no 🤷 

Screenshot 2025-01-24 at 10 24 22

vicb: (the difference being that `import type` does not import anything (i.e. TS only construct)

examples/middleware/e2e/cloudflare-context.spec.ts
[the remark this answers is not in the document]
dario-piotrowicz: I just simplified the test there: [6e6d076](https://github.com/opennextjs/opennextjs-cloudflare/pull/265/commits/6e6d076b6071e9d83e904fad20f8e77549f27909)

All we care about here is checking that we can get access to the cloudflare context in the middleware, regarding the context variables we already test that they are present here: https://github.com/opennextjs/opennextjs-cloudflare/blob/a18f826f6b7ece350cadb7e5f9bcc56fdcd84c7f/examples/api/e2e/base.spec.ts#L18-L26

So I think that testing the existence of the `env` object is enough here (and if we wanted to have a comprehensive e2e test for `getCloudflareContext` with all the various variables, `cf`, etc... that should probably live in the api example), what do you think?

examples/api/e2e/playwright.config.ts
dario-piotrowicz: updated 👍

.changeset/chilly-dryers-begin.md
dario-piotrowicz: I've updated the changeset I hope the new version works for you

.changeset/chilly-dryers-begin.md
```diff
@@ -0,0 +1,28 @@
+---
+"@opennextjs/cloudflare": patch
```
vicb, line 2: what about cutting a minor?

It would signal users that there are changes:

- calling the`initOpenNextCloudflareForDev` function,
- `getCloudflareContext` becoming sync

If you do that, please also prepare:
- a PR to update the c3 template to 0.4.x
- a PR to the doc site (copy the current content to 0.3, add the changes in the current content, add a migration from 0.3)

vicb (approved):
Could you please update https://github.com/opennextjs/opennextjs-cloudflare/blob/main/packages/cloudflare/README.md

Either to ask the user to add the call to the init function
Or if you can update opennext.js.org, remove the content and link to the doc site (I was mostly using the content for the experimental branch when the experimental features were not documented anywhere else)

Great stuff, thanks 🚀

.changeset/chilly-dryers-begin.md
dario-piotrowicz: done 👍

examples/middleware/middleware.ts
```diff
@@ -1,7 +1,9 @@
 import { NextRequest, NextResponse, NextFetchEvent } from "next/server";
 import { clerkMiddleware } from "@clerk/nextjs/server";
 
-export function middleware(request: NextRequest, event: NextFetchEvent) {
+import { getCloudflareContext } from "@opennextjs/cloudflare";
+
+export async function middleware(request: NextRequest, event: NextFetchEvent) {
```
vicb, line 6: I don't think this needs to be `async` ?

```suggestion
export function middleware(request: NextRequest, event: NextFetchEvent) {
```

packages/cloudflare/README.md
```diff
@@ -6,65 +6,7 @@ Deploy Next.js apps to Cloudflare!
...
-      wrapper: "cloudflare-edge",
-      converter: "edge",
-      proxyExternalRequest: "fetch",
-    },
-  },
-};
-
-export default config;
-```
+To get started with the adapter visit the [official get started documentation](https://opennext.js.org/cloudflare/get-started).
```
vicb, line 9: Can you please also drop the KI section below which is out of date

vicb (approved):
Looks great, thanks!

examples/middleware/middleware.ts
dario-piotrowicz: good spot 👍 , I must have added it before making `getCloudflareContext` sync and forgot to revert it later

packages/cloudflare/README.md
dario-piotrowicz: done [da78a85](https://github.com/opennextjs/opennextjs-cloudflare/pull/265/commits/da78a853aaadb46feae197ed90c0274d5c4624d9) 👍

james-elicx (approved):

vicb:
@dario-piotrowicz, would you mind posting this on Discord once you push a release? (0.4 + getContext sync) 🙏

PR #270, opened 2025-01-21
refactor: add wrangler chema, fix CLERK error

vicb:
`CLERK_ENCRYPTION_KEY` is added to fix an error log during e2e

### DIFF
--- a/examples/middleware/wrangler.json
+++ b/examples/middleware/wrangler.json
@@ -1,4 +1,5 @@
 {
+  "$schema": "node_modules/wrangler/config-schema.json",
   "main": ".open-next/worker.js",
   "name": "middleware",
   "compatibility_date": "2024-12-30",
@@ -6,5 +7,9 @@
   "assets": {
     "directory": ".open-next/assets",
     "binding": "ASSETS"
+  },
+  "vars": {
+    // Note: Do not define keys in wrangler config.
+    "CLERK_ENCRYPTION_KEY": "key"
   }
 }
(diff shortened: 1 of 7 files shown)


examples/middleware/wrangler.json
```diff
@@ -1,10 +1,15 @@
...
   "name": "middleware",
   "compatibility_date": "2024-12-30",
   "compatibility_flags": ["nodejs_compat"],
   "assets": {
     "directory": ".open-next/assets",
     "binding": "ASSETS"
+  },
+  "vars": {
+    // Note: Do not define keys in wrangler config.
+    "CLERK_ENCRYPTION_KEY": "key"
```
james-elicx, line 13: would it be better to commit a .dev.vars file for this?

vicb: I was wondering if a `.env` or `.dev.vars` would be better but adding them to the repo is also a bad practice.
At least here I can add a comment.
What do you think would be the best way?

james-elicx: As an example it feels less consequential than demoing doing it directly in wrangler.json as peoples' gitignore would exclude env files by default. Instead of committing them in the file people associate with secrets / interpret as not being something that should be committed, we're doing it directly in the config, which feels like the wrong place an example should put something. 

Alternatively, we could have a .env.example file and in the git workflow rename all .env.example files to .env before the e2es are run perhaps - that feels like a decent option for an example

vicb: I updated the PR to add a `.env` - I feel like renaming would be a bit overkill?

james-elicx (approved):

vicb:
Thanks for the review!

PR #271, opened 2025-01-21
refactor: use the new regex utility for cross-platform paths

james-elicx:
Migrates our esbuild filter usage to use the new utility.

### DIFF
--- a/packages/cloudflare/src/cli/build/bundle-server.ts
+++ b/packages/cloudflare/src/cli/build/bundle-server.ts
@@ -4,6 +4,7 @@ import path from "node:path";
 import { fileURLToPath } from "node:url";
 
 import type { BuildOptions } from "@opennextjs/aws/build/helper.js";
+import { getCrossPlatformPathRegex } from "@opennextjs/aws/utils/regex.js";
 import { build, Plugin } from "esbuild";
 
 import { Config } from "../config.js";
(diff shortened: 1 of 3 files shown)


packages/cloudflare/src/cli/build/bundle-server.ts
```diff
@@ -183,7 +184,7 @@ function createFixRequiresESBuildPlugin(config: Config): Plugin {
     name: "replaceRelative",
     setup(build) {
       // Note: we (empty) shim require-hook modules as they generate problematic code that uses requires
-      build.onResolve({ filter: /^\.(\/|\\)require-hook$/ }, () => ({
+      build.onResolve({ filter: getCrossPlatformPathRegex("^\\./require-hook$", { escape: false }) }, () => ({
```
vicb, line 187: nit: use `String.raw` to avoid double escaping (`\\`)

vicb (approved):
Thanks, looks nicer!

packages/cloudflare/src/cli/build/bundle-server.ts
james-elicx:
```suggestion
      build.onResolve({ filter: getCrossPlatformPathRegex(String.raw`^\./require-hook$`, { escape: false }) }, () => ({
```

PR #275, opened 2025-01-22
fix: fix the build of examples/e2e/app-router

vicb:


vicb:
Thanks for the tip @james-elicx, `ignoreBuildErrors: true,` did the trick 🙏

packages/cloudflare/src/cli/build/patches/ast/util.ts
```diff
@@ -0,0 +1,48 @@
...
+  matches.forEach((match) => {
+    edits.push(
+      match.replace(
+        // Replace known placeholders by their value
+        fix
+          .replace(/\$\$\$([A-Z0-9_]+)/g, (_m, name) =>
+            match
+              .getMultipleMatches(name)
+              .map((n) => n.text())
+              .join()
```
james-elicx, line 40: is it possible that you would get multiple matches for the placeholder name in the fix here? If so, is it right that all the matches are joined together when being replaced? I'm just wondering if you might end up with an edge case that results in incorrect values here due to multiple matches for the placeholder.

vicb: ast-grep has placeholders for single node `$SINGLE` or multiple nodes `$$$MULTI` see [meta-variables](https://ast-grep.github.io/guide/pattern-syntax.html#meta-variable).

If your question is whether the variable can be re-used multiple times in the rule, then [yes](https://ast-grep.github.io/guide/pattern-syntax.html#meta-variable-capturing) but it would capture the same thing so it would like serialize to this one thing only.

Does it answer the question?

james-elicx: Those links were helpful in understanding what's going on.

So what I was trying to say would probably be something like this example:
if your code contained `t = require('caniuse-lite', 'test');`, I believe the output would be incorrect - testing locally gives me `t = require('caniuse-lite',,,'test');`.

I know, that's not a valid require statement anyway because it's only supposed to have one arg. For this patch, it doesn't matter, but for other patches I think how the multi matches are replaced could potentially cause issues. I'm not quite sure, but happy to leave that for now

vicb: Oh you are absolutely right, I should have used `.join("")`, I keep forgetting that `,` is the default :(

vicb, line 40:
```suggestion
              .join("")
```

james-elicx: it might be useful if we start writing unit tests for these patches to try and catch things like that

james-elicx (approved):

vicb:
Thanks for the review and for catching this bug (yes, tests would be better than comments on the rule).

packages/cloudflare/src/cli/build/patches/ast/util.ts
vicb: I agree

image

I will start by creating a P2 issue

PR #278, opened 2025-01-23
refactor: join join arguments

### LINKED ISSUE #142: [FEATURE] handle path.join / separator consistently
We should handle path.join consistently
See https://github.com/opennextjs/opennextjs-cloudflare/pull/140#discussion_r1856826863

No rush, we can wait to have a monorepo

vicb:
fixes #142

### DIFF
--- a/packages/cloudflare/src/cli/build/patches/to-investigate/inline-eval-manifest.ts
+++ b/packages/cloudflare/src/cli/build/patches/to-investigate/inline-eval-manifest.ts
@@ -14,7 +14,7 @@ import { normalizePath } from "../../utils/index.js";
  */
 export function inlineEvalManifest(code: string, config: Config): string {
   const manifestJss = globSync(
-    normalizePath(join(config.paths.output.standaloneAppDotNext, "**", "*_client-reference-manifest.js"))
+    normalizePath(join(config.paths.output.standaloneAppDotNext, "**/*_client-reference-manifest.js"))
   ).map((file) =>
     normalizePath(file).replace(normalizePath(config.paths.output.standaloneApp) + posix.sep, "")
   );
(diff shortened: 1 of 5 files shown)


dario-piotrowicz:
I don't understand the refactoring here... what's the issue you're trying to address? consistency? in that case shouldn't we always use either `path.join` or `/`s? I don't understand what benefit there is in having both `join` and `/`s 😕

james-elicx (approved):

vicb:
> I don't understand the refactoring here... what's the issue you're trying to address? consistency? in that case shouldn't we always use either `path.join` or `/`s? I don't understand what benefit there is in having both `join` and `/`s 😕 (it feels to me like you are consistently making the code inconsistent here 🤔)

TL;DR: Pete noted that `join` normalizes the paths and using `"sub/path/to"` is nicer than `"sub", "path", "to"`.

You can follow the link in link in the PR description and you'll end up the [discussion](https://github.com/opennextjs/opennextjs-cloudflare/pull/140#discussion_r1856826863)

vicb:
_I'm tackling this now because I'm working on simplyfying the 15 different paths we have in the config, and dropping the `.open-next/.next` folder which is a duplicate of what's in `.open-next/server-bundle/...`_

dario-piotrowicz:
> > I don't understand the refactoring here... what's the issue you're trying to address? consistency? in that case shouldn't we always use either `path.join` or `/`s? I don't understand what benefit there is in having both `join` and `/`s 😕 (it feels to me like you are consistently making the code inconsistent here 🤔)
> 
> TL;DR: Pete noted that `join` normalizes the paths and using `"sub/path/to"` is nicer than `"sub", "path", "to"`.
> 
> You can follow the link in link in the PR description and you'll end up the [discussion](https://github.com/opennextjs/opennextjs-cloudflare/pull/140#discussion_r1856826863)

if that's the case then I think using `path.normalize` with `/`s instead would be the most explicit clear and consistent solution (i.e. if you're using `path.join` to normalize paths, why not just use `path.normalize` and make the intent explicit?)

vicb:
If you're suggesting to use `path.normalize(path.join(...))`, I don't really see the point.

I think the two reasonable choice are:

- the current way `path.join(base, "sub", "dir")` or
- `path.join(base, "sub/dir")`

I'll let @petebacondarwin decide - I don't mind either way.

vicb:
I think it wouldn't work

image

PR #282, opened 2025-01-25
refactor: use the traced file copy from OpenNext

vicb:
fixes #251 
relates to #14 

Before this PR, the Next compiled output was copied to `.open-next/.next` and some files were used/patched from there.

This PR makes sure we always use the files from `.open-next/.server-functions/<fnName>` instead.

This allows to simplify a lot of the config object that was used before.

PR #283, opened 2025-01-26
fix: patch @vercel/og usage to use the edge runtime version

james-elicx:
PR is based off of #282 and should not be merged before that one.

This PR does the following:
- Changes the function for getting rule edits to also enable returning matches.
- Patches the import of `index.node.js` to `index.edge.js`.
- Patches the fallback font fetch to use an import.
- Renames the fallback font to be suffixed with `.bin`.
- Copies the edge version to the OpenNext `node_modules`.
- Unit tests + e2e test.

Might be worth noting that the edge runtime version is only available in the node_modules location referenced by the traced files in .next/server, and not .next/standalone. So we need to use that output to get the edge file and work our way back up to node_modules, and bring that over the right place for the standalone build in our output.

Not sure if this approach is entirely what you had in mind.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches: [main, experimental]
   pull_request:
-    branches: [main, experimental]
 
 jobs:
   checks:
--- a/examples/api/e2e/base.spec.ts
+++ b/examples/api/e2e/base.spec.ts
@@ -1,4 +1,16 @@
 import { test, expect } from "@playwright/test";
+import type { BinaryLike } from "node:crypto";
+import { createHash } from "node:crypto";
+
+const OG_MD5 = "2f7b724d62d8c7739076da211aa62e7b";
+
+export function validateMd5(data: Buffer, expectedHash: string) {
+  return (
+    createHash("md5")
+      .update(data as BinaryLike)
+      .digest("hex") === expectedHash
+  );
+}
 
 test("the application's noop index page is visible and it allows navigating to the hello-world api route", async ({
   page,
@@ -47,4 +59,5 @@ test("generates an og image successfully", async ({ page }) => {
   const res = await page.request.get("/og");
   expect(res.status()).toEqual(200);
   expect(res.headers()["content-type"]).toEqual("image/png");
+  expect(validateMd5(await res.body(), OG_MD5)).toEqual(true);
 });
--- a/examples/api/e2e/base.spec.ts
+++ b/examples/api/e2e/base.spec.ts
@@ -42,3 +42,9 @@ test("returns correct information about the request from a route handler", async
   const expectedURL = expect.stringMatching(/https?:\/\/localhost:(?!3000)\d+\/api\/request/);
   await expect(res.json()).resolves.toEqual({ nextUrl: expectedURL, url: expectedURL });
 });
+
+test("generates an og image successfully", async ({ page }) => {
+  const res = await page.request.get("/og");
+  expect(res.status()).toEqual(200);
+  expect(res.headers()["content-type"]).toEqual("image/png");
+});
--- a/packages/cloudflare/src/cli/build/bundle-server.ts
+++ b/packages/cloudflare/src/cli/build/bundle-server.ts
@@ -176,7 +176,7 @@ async function updateWorkerBundledCode(workerOutputFile: string, buildOpts: Buil
 
   const bundle = parse(Lang.TypeScript, patchedCode).root();
 
-  const edits = patchOptionalDependencies(bundle);
+  const { edits } = patchOptionalDependencies(bundle);
 
   await writeFile(workerOutputFile, bundle.commitEdits(edits));
 }
--- a/packages/cloudflare/src/cli/build/bundle-server.ts
+++ b/packages/cloudflare/src/cli/build/bundle-server.ts
@@ -34,7 +34,7 @@ export async function bundleServer(buildOpts: BuildOptions): Promise<void> {
 
   patches.patchWranglerDeps(buildOpts);
   await patches.updateWebpackChunksFile(buildOpts);
-  await patches.patchVercelOgLibrary(buildOpts);
+  patches.patchVercelOgLibrary(buildOpts);
 
   const outputPath = path.join(outputDir, "server-functions", "default");
   const packagePath = getPackagePath(buildOpts);
--- a/packages/cloudflare/src/cli/build/bundle-server.ts
+++ b/packages/cloudflare/src/cli/build/bundle-server.ts
@@ -33,7 +33,8 @@ export async function bundleServer(buildOpts: BuildOptions): Promise<void> {
   console.log(`\x1b[35m⚙️ Bundling the OpenNext server...\n\x1b[0m`);
 
   patches.patchWranglerDeps(buildOpts);
-  patches.updateWebpackChunksFile(buildOpts);
+  await patches.updateWebpackChunksFile(buildOpts);
+  await patches.patchVercelOgLibrary(buildOpts);
 
   const outputPath = path.join(outputDir, "server-functions", "default");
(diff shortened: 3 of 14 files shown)


vicb:
> Might be worth noting that the edge runtime version is only available in the node_modules location referenced by the traced files in .next/server, and not .next/standalone. So we need to use that output to get the edge file and work our way back up to node_modules, and bring that over the right place for the standalone build in our output.

Yep, I created https://github.com/opennextjs/opennextjs-cloudflare/issues/285 to track this

.github/workflows/checks.yml
```diff
@@ -4,7 +4,6 @@ on:
   push:
     branches: [main, experimental]
   pull_request:
-    branches: [main, experimental]
```
vicb, line 7: why changing this?

examples/api/e2e/base.spec.ts
```diff
@@ -42,3 +42,9 @@ test("returns correct information about the request from a route handler", async
   const expectedURL = expect.stringMatching(/https?:\/\/localhost:(?!3000)\d+\/api\/request/);
   await expect(res.json()).resolves.toEqual({ nextUrl: expectedURL, url: expectedURL });
 });
+
+test("generates an og image successfully", async ({ page }) => {
+  const res = await page.request.get("/og");
+  expect(res.status()).toEqual(200);
+  expect(res.headers()["content-type"]).toEqual("image/png");
```
vicb, line 49: aws uses [md5 checksum](https://github.com/opennextjs/opennextjs-aws/blob/a3337e04da43e3ea4c421471b318964f985ea729/packages/tests-e2e/tests/appRouter/og.test.ts#L5) in their tests, maybe something to add here

packages/cloudflare/src/cli/build/patches/ast/vercel-og.ts
```diff
@@ -0,0 +1,63 @@
+import { SgNode } from "@ast-grep/napi";
+
+import { applyRule } from "./util.js";
+
+export const vercelOgImportRule = `
+rule:
+  pattern: $NODE
+  kind: string
+  regex: "next/dist/compiled/@vercel/og/index.node.js"
```
vicb, line 9: nit: escape "."

packages/cloudflare/src/cli/build/patches/ast/vercel-og.ts
```diff
@@ -0,0 +1,63 @@
...
+          kind: identifier
+          regex: ^fallbackFont$
+    - has:
+        kind: call_expression
+        pattern: fetch(new URL("$PATH", $$$REST))
+        stopBy: end
+
+fix: |-
+  async function getFallbackFont() {
+    return (await import("$PATH.bin")).default
```
vicb, line 49: thinking more about it, I'm not sure we really need the `.bin`. Would you mind trying with that?

packages/cloudflare/src/cli/build/patches/ast/vercel-og.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export const vercelOgFallbackFontRule = `
+rule:
+  kind: variable_declaration
+  all:
+    - has:
+        kind: variable_declarator
+        has:
+          kind: identifier
+          regex: ^fallbackFont$
```
vicb, line 41: nit: maybe it would be safer to detect the `fetch(font_name)` in case the variable name changes - but let's merge this, we can refine later if needed.

packages/cloudflare/src/cli/build/patches/investigated/patch-vercel-og-library.ts
```diff
@@ -0,0 +1,60 @@
...
+    const outputEdgePath = path.join(outputDir, "index.edge.js");
+
+    if (!existsSync(outputEdgePath)) {
+      const tracedEdgePath = path.join(
+        path.dirname(traceInfoPath),
+        tracedNodePath.replace("index.node.js", "index.edge.js")
+      );
+
+      copyFileSync(tracedEdgePath, outputEdgePath);
+      rmSync(outputEdgePath.replace("index.edge.js", "index.node.js"));
```
vicb, line 44: We don't really need to delete the node file?

packages/cloudflare/src/cli/build/patches/investigated/patch-vercel-og-library.ts
```diff
@@ -0,0 +1,60 @@
...
+
+import { parseFile } from "../ast/util.js";
+import { patchVercelOgFallbackFont, patchVercelOgImport } from "../ast/vercel-og.js";
+
+type TraceInfo = { version: number; files: string[] };
+
+/**
+ * Patches the usage of @vercel/og to be compatible with Cloudflare Workers.
+ *
+ * This involves;
```
vicb, line 16: maybe move those comments inline to identify what part of the code does what?

vicb (approved):
Awesome work, thanks!

.github/workflows/checks.yml
james-elicx: This PR is pointing at your PR's branch as I wanted to do it with the refactored paths so it didn't need changing again. Therefore, had to change this to run everything in my PR

vicb: Oh so when you remove the explicit branches, it will run for all branches?
If that's the case I'm 👍

packages/cloudflare/src/cli/build/patches/ast/vercel-og.ts
james-elicx: You need to configure a custom loader for ttf files if you don't use .bin.

.github/workflows/checks.yml
james-elicx: indeed

packages/cloudflare/src/cli/build/patches/ast/vercel-og.ts
vicb: Would you mind adding a comment in the code?

packages/cloudflare/src/cli/build/patches/ast/vercel-og.ts
```diff
@@ -0,0 +1,63 @@
...
+    - has:
+        kind: call_expression
+        pattern: fetch(new URL("$PATH", $$$REST))
+        stopBy: end
+
+fix: |-
+  async function getFallbackFont() {
+    return (await import("$PATH.bin")).default
```
vicb, line 49: nit: add trailing ";"

packages/cloudflare/src/cli/build/bundle-server.ts
```diff
@@ -33,7 +33,8 @@ export async function bundleServer(buildOpts: BuildOptions): Promise<void> {
   console.log(`\x1b[35m⚙️ Bundling the OpenNext server...\n\x1b[0m`);
 
   patches.patchWranglerDeps(buildOpts);
-  patches.updateWebpackChunksFile(buildOpts);
+  await patches.updateWebpackChunksFile(buildOpts);
+  await patches.patchVercelOgLibrary(buildOpts);
```
vicb, line 37: It would be nice to be able to move the changes to `updateWorkerBundledCode()`
It would probably require adding externals to the ESBuild config.
With that we could `parse()` (ast-grep) the code once and apply multiple edits.

(This again could be handled in a follow-up  PR)

james-elicx: I don't think we can do it there as this is patching a specific route's file whereas that function is patching handler.mjs, and the code for each of the routes doesn't appear to be bundled into the handler.mjs

vicb: > the code for each of the routes doesn't appear to be bundled into the handler.mjs

yet :)

As I mentioned, we should move patches that add dependencies (`require`) from after the bundle to before/during the bundling.

I do have an ast-grep patch solving the route issue. I need to rebase and submit the PR.

PR #287, opened 2025-01-27
refactor: migrate the requirePage patch to ast-grep

vicb:
@james-elicx I'd like to land that but there is one issue with og: it uses relative paths and that messes with the external in `bundle-server`.

Could you add one more rule to convert relative path to absolute (in import function expressions)?

Thanks!


vicb:
@james-elicx maybe actually this should rather be an ESBuild plugin as it would solve the issue for `.bin` and `.wasm?module` across module, this is not really OG specific - hopefully you have not started on this. I'll try and report.

vicb:
Added a commit to resolve external paths to absolute, ready for review!

packages/cloudflare/src/cli/build/bundle-server.ts
```diff
@@ -48,8 +50,18 @@ export async function bundleServer(buildOpts: BuildOptions): Promise<void> {
...
-    plugins: [createFixRequiresESBuildPlugin(buildOpts)],
-    external: ["./middleware/handler.mjs", "caniuse-lite"],
+    plugins: [
+      createFixRequiresESBuildPlugin(buildOpts),
+      inlineRequirePagePlugin(buildOpts),
+      setWranglerExternal(),
+    ],
+    external: [
+      "./middleware/handler.mjs",
+      // Next optional dependencies.
```
vicb, line 60: note: 
I plan to refactor how optional dependencies are handled.
That is put them all in a single array, add comments and use that here and pass them to `patchOptionalDependencies()`

packages/cloudflare/src/cli/build/patches/ast/optional-deps.ts
```diff
@@ -16,7 +16,7 @@ rule:
     pattern: $MOD
     kind: string_fragment
     stopBy: end
-    regex: ^caniuse-lite(/|$)
+    regex: ^caniuse-lite(/|$)|jimp(/|$)|probe-image-size(/|$)
```
dario-piotrowicz, line 19: are you sure about this regex? the `^` only applies to `caniuse-lite` so the regex also matches things just ending with for example `jimp` (e.g. like `hello-jimp`)

maybe this would be more appropriate?
```suggestion
    regex: ^(caniuse-lite|jimp|probe-image-size)(/|$)
```

packages/cloudflare/src/cli/build/patches/plugins/wrangler-external.ts
```diff
@@ -0,0 +1,33 @@
+/**
+ * Makes files handled by wrangler external.
+ *
+ * Paths need to be absolute so that they are valid in the output bundle.
+ */
```
dario-piotrowicz, line 5: should this comment be above the function?

Also "makes files handled by wrangler external." is in my opinion very generic and uninformative, basically without looking at the code this tells me nothing

dario-piotrowicz: regarding "Paths need to be absolute so that they are valid in the output bundle." I think it would be appropriate to briefly mention which paths we're talking about and why they need to be absolute

packages/cloudflare/src/cli/build/patches/to-investigate/inline-next-require.ts
```diff
@@ -1,53 +0,0 @@
-import { existsSync, readFileSync } from "node:fs";
-import { join } from "node:path";
-
-import { type BuildOptions, getPackagePath } from "@opennextjs/aws/build/helper.js";
-
-/**
- * The following avoid various Next.js specific files `require`d at runtime since we can just read
- * and inline their content during build time
- */
```
dario-piotrowicz, line 9: is this comment not worth porting over the new plugin? there I don't see any explanation/context on what the plugin is doing (and I don't think that that is obvious)

packages/cloudflare/src/cli/build/patches/plugins/require-page.ts
```diff
@@ -0,0 +1,88 @@
...
+            return { contents: patchCode(jsCode, getRule(buildOpts)) };
+          }
+        }
+      );
+    },
+  };
+}
+
+function getRule(buildOpts: BuildOptions) {
+  // Load manifests
```
dario-piotrowicz, line 32:
```suggestion
```

packages/cloudflare/src/cli/build/patches/ast/optional-deps.ts
vicb: good catch!

packages/cloudflare/src/cli/build/patches/plugins/wrangler-external.ts
vicb: I expanded the comment to make it clearer.

packages/cloudflare/src/cli/build/patches/to-investigate/inline-next-require.ts
vicb: added a comment in the new patch.

packages/cloudflare/src/cli/build/patches/plugins/wrangler-external.ts
```diff
@@ -0,0 +1,43 @@
+/**
+ * ESBuild plugin to mark files bundled by wrangler as external.
+ *
+ * `.wasm` and `.bin` will ultimately be bundled by wrangler.
+ * We should only mark them as external in the adapter.
+ *
+ * However simply marking them as external would copy the import path to the bundle,
+ * i.e. `import("./file.wasm?module")` and given than the bundle is generated in a
+ * different location than the input files, the relative path would not be valid.
```
dario-piotrowicz, line 9: I think this is very helpful, thanks! 🤩

dario-piotrowicz (approved):
Thanks for the updates @vicb

Looks good to me 😄

PR #289, opened 2025-01-28
test: import app-router e2e tests from aws

vicb:
Run the tests with `pnpm -F app-router e2e-fix`

Note that the script is named `e2e-fix` so that it doesn't run on the CI yet as some tests are failing. 

Some failures are expected (i.e. ISR) and some need to be investigated.

### DIFF
--- a/examples/e2e/app-router/package.json
+++ b/examples/e2e/app-router/package.json
@@ -10,8 +10,9 @@
     "lint": "next lint",
     "clean": "rm -rf .turbo node_modules .next .open-next",
     "build:worker": "pnpm opennextjs-cloudflare",
-    "dev:worker": "wrangler dev --port 8770 --inspector-port 9330",
-    "preview": "pnpm build:worker && pnpm dev:worker"
+    "dev:worker": "wrangler dev --port 8790 --inspector-port 9350",
+    "preview": "pnpm build:worker && pnpm dev:worker",
+    "e2e-fix": "playwright test -c e2e/playwright.config.ts"
   },
   "dependencies": {
     "@opennextjs/cloudflare": "workspace:*",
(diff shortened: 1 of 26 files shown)


examples/e2e/app-router/package.json
```diff
@@ -10,8 +10,9 @@
     "lint": "next lint",
     "clean": "rm -rf .turbo node_modules .next .open-next",
     "build:worker": "pnpm opennextjs-cloudflare",
-    "dev:worker": "wrangler dev --port 8770 --inspector-port 9330",
-    "preview": "pnpm build:worker && pnpm dev:worker"
+    "dev:worker": "wrangler dev --port 8790 --inspector-port 9350",
+    "preview": "pnpm build:worker && pnpm dev:worker",
+    "e2e-fix": "playwright test -c e2e/playwright.config.ts"
```
dario-piotrowicz, line 15: `preview` doesn't seem to work as this app is missing the wrangler dependency, consecutively `e2e-fix` also doesn't work

unless I'm doing something wrong?
Screenshot 2025-01-28 at 14 51 14

vicb: Good point, I missed adding wrangler in all imported apps probably, I'll amend the PRs.
It works locally for me, probably wrangler is installed in a parent folder

vicb:
> Personally I don't see the point in introducing this with the `e2e-fix` script that doesn't run in CI, I think that it should instead have a normal `e2e` script but the incompatible tests get properly skipped (potentially with todo comments and respective gh issues)

Agreed, that is the next step, there is already an issue for that at https://github.com/opennextjs/opennextjs-cloudflare/issues/291. We should create sub-issue for tests that we will skip and investigate them.

Thanks for the review

Issue #119, opened 2024-11-02
Could not resolve "react-dom/server.edge"


mai1015:
Im trying to deploy my nextj app but it is not working

package.json
```json
{
    "name": "awesome-project",
    "version": "0.1.0",
    "private": true,
    "scripts": {
        "build:worker": "cloudflare",
        "dev:worker": "wrangler dev --port 8771",
        "preview:worker": "npm run build:worker && npm run dev:worker",
        "deploy:worker": "npm run build:worker && wrangler deploy"
    },
    "dependencies": {
        "next": "^15.0.2",
        "react": "^18.3.1",
        "react-dom": "^18.3.1"
    },
    "devDependencies": {
        "@opennextjs/cloudflare": "^0.2.1",
        "@types/node": "^22",
        "@types/react": "^18",
        "@types/react-dom": "^18",
        "eslint": "^8",
        "eslint-config-next": "15.0.2",
        "next-translate-plugin": "^2.6.2",
        "postcss": "^8",
        "prettier": "^3.3.3",
        "prettier-plugin-svelte": "^3.2.7",
        "prettier-plugin-tailwindcss": "^0.6.8",
        "tailwindcss": "^3.4.1",
        "typescript": "^5",
        "wrangler": "^3.84.1"
    }
}
```
Im trying to deploy the server with `npm run deploy:worker` and here is the log

log
```
✘ [ERROR] Build failed with 1 error:

  ✘ [ERROR] Could not resolve "react-dom/server.edge"
  
  
  .worker-next/.next/standalone/node_modules/next/dist/compiled/next-server/pages.runtime.prod.js:14:16350:
        14 │ ...";e.exports=require("react-dom/server.edge")},path:e=>{"use str...
           ╵                        ~~~~~~~~~~~~~~~~~~~~~~~
  
    The path "./server.edge" is not exported by package "react-dom":
  
      .worker-next/.next/standalone/node_modules/react-dom/package.json:39:13:
        39 │   "exports": {
           ╵              ^
  
  You can mark the path "react-dom/server.edge" as external to exclude it from the bundle, which
  will remove this error. You can also surround this "require" call with a try/catch block to handle
  this failure at run-time instead of bundle-time.
  ```

Is there anything I did wrong or it is not related to opennextjs-cloudflare

vicb:
Next 15 is [not supported yet](https://opennext.js.org/cloudflare#supported-nextjs-versions)

Also note that we can not debug issues with a repro.

[...]

PR #297, opened 2025-01-29
refactor: Make the list of optional dependencies configurable

vicb:
The second commits adds `critters` and `react-dom/server.edge` to the list of optional dependencies.

Note that I didn't re-use the comment from `wrangler-deps.ts` for `react-dom/server.edge` as it was not fully accurate. `pages.runtime.prod.js` does actually wrap the import in the same way as [Next.js](https://github.com/vercel/next.js/blob/6784575/packages/next/src/server/ReactDOMServerPages.js):

```ts
"./dist/server/ReactDOMServerPages.js":(e,t,r)=>{"use strict";let n;try{n=r("react-dom/server.edge")}catch(e){if("MODULE_NOT_FOUND"!==e.code&&"ERR_PACKAGE_PATH_NOT_EXPORTED"!==e.code)throw e;n=r("react-dom/server.browser")}e.exports=n}
```

The problem is when the package is re-exported:

```ts
e=>{"use strict";e.exports=require("react-dom/server.edge")}
```

This last site will be patched by the optional dependencies patch:

```ts
"react-dom/server.edge": (e2) => {
                    "use strict";
                    try {
  e2.exports = require("react-dom/server.edge");
} catch {
  throw new Error('The optional dependency "react-dom/server.edge" is not installed');
};
```

Verified by modifying the app-router according to [the initial issue](https://github.com/opennextjs/opennextjs-cloudflare/issues/119):

```diff
-    "next": "catalog:e2e",
-    "react": "catalog:e2e",
-    "react-dom": "catalog:e2e"
+    "next": "^15.0.2",
+    "react": "^18.3.1",
+    "react-dom": "^18.3.1"
```

### DIFF
--- a/packages/cloudflare/src/cli/build/bundle-server.ts
+++ b/packages/cloudflare/src/cli/build/bundle-server.ts
@@ -22,7 +22,14 @@ const packageDistDir = path.join(path.dirname(fileURLToPath(import.meta.url)), "
  * They are not required for Next.js to run but only needed to enabled specific features.
  * When one of those dependency is required, it should be installed by the application.
  */
-const optionalDependencies = ["caniuse-lite", "jimp", "probe-image-size"];
+const optionalDependencies = [
+  "caniuse-lite",
+  "critters",
+  "jimp",
+  "probe-image-size",
+  // `server.edge` is not available in react-dom@18
+  "react-dom/server.edge",
+];
 
 /**
  * Bundle the Open Next server.
--- a/packages/cloudflare/src/cli/build/bundle-server.ts
+++ b/packages/cloudflare/src/cli/build/bundle-server.ts
@@ -18,6 +18,13 @@ import { normalizePath, patchCodeWithValidations } from "./utils/index.js";
 const packageDistDir = path.join(path.dirname(fileURLToPath(import.meta.url)), "../..");
 
 /**
+ * List of optional Next.js dependencies.
+ * They are not required for Next.js to run but only needed to enabled specific features.
+ * When one of those dependency is required, it should be installed by the application.
+ */
+const optionalDependencies = ["caniuse-lite", "jimp", "probe-image-size"];
+
+/**
  * Bundle the Open Next server.
  */
 export async function bundleServer(buildOpts: BuildOptions): Promise<void> {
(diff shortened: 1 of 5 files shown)


packages/cloudflare/src/cli/build/bundle-server.ts
```diff
@@ -17,6 +17,20 @@ import { normalizePath, patchCodeWithValidations } from "./utils/index.js";
...
+ * They are not required for Next.js to run but only needed to enabled specific features.
+ * When one of those dependency is required, it should be installed by the application.
+ */
+const optionalDependencies = [
+  "caniuse-lite",
+  "critters",
+  "jimp",
+  "probe-image-size",
+  // `server.edge` is not available in react-dom@18
+  "react-dom/server.edge",
```
james-elicx, line 31: how come we're no longer going to change this require to use `react-dom/server.browser` for <= 18?

james-elicx: I did read the description, but it wasn't clear to me why the re-exported version doesn't need this

vicb: This is what I tried to explain in the PR description: `pages.runtime.prod.js` does that (first snippet), the actual build failure was only caused by the re-export of `react-dom/server.edge` (second snippet) which is patched by the optional deps (third snippet).

I have tested with the versions from the original issue and it works (fails without the patch).

Does this make sense?

james-elicx: Are we saying that the re-exported version is never actually used for the app then?

vicb: I think we don't need to know.

If some code is using the re-export then `require("react-dom/server.edge")` will throw in `pages.runtime.prod.js`.
We are respecting that behavior.

Either nobody is using it or that code have to use a try...catch

james-elicx (approved):

packages/cloudflare/src/cli/build/bundle-server.ts
james-elicx: Understood

PR #298, opened 2025-01-29
refactor: move ast-grep based patches to `patches/ast`

vicb:
Let's use the `patches/ast` folder for all ast-grep based patches

/cc @dario-piotrowicz


james-elicx (approved):

vicb:
Thanks for the review!

PR #300, opened 2025-01-29
Add a test for gh-119

vicb:
Add a test to make sure we have no regression.

The e2e runs on only one browser - do we really need to run the others on 3 different browsers?


examples/bugs/gh-119/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "build:worker": "pnpm opennextjs-cloudflare",
+    "dev:worker": "wrangler dev --port 8750 --inspector-port 9370",
+    "preview": "pnpm build:worker && pnpm dev:worker",
+    "e2e": "playwright test -c e2e/playwright.config.ts",
+    "cf-typegen": "wrangler types --env-interface CloudflareEnv"
+  },
+  "dependencies": {
+    "next": "15.0.4",
+    "react-dom": "^18.3.1",
+    "react": "^18.3.1"
```
dario-piotrowicz, line 19: is this what makes this reproduce the issue? ideally we should document it somewhere as it is very easy to miss (you can add some "json comments" here 🤔)

dario-piotrowicz (commented):
Looks good to me, but I don't understand, why are you bumping the timeouts to 500_000? I really hope that that is not necessary, is it?

dario-piotrowicz:
Also what are you adding is not an "example" I wonder if we should have maybe two directories "examples" and "fixtures"/"e2es" or something like that 🤔

or maybe just rename "examples" and simply use them as fixtures

not a big deal either way

vicb:
> Looks good to me, but I don't understand, why are you bumping the timeouts to 500_000? I really hope that that is not necessary, is it?

100k was still erroring and making me restart the tests multiple times.

examples/bugs/gh-119/package.json
vicb: It's straight from #119 - that's why I make the folder name obvious `gh-119`.

My next PR will follow the same scheme, cloning the repro from the issue.

vicb:
> Also what are you adding is not an "example" I wonder if we should have maybe two directories "examples" and "fixtures"/"e2es" or something like that 🤔
> 
> or maybe just rename "examples" and simply use them as fixtures
> 
> not a big deal either way

Feel free to create an issue with refactoring ideas.
For now I have dependent PRs using the same structure so I don't want to loose too much time for that.
Thanks!

james-elicx (approved):
the timeout problem is a bit depressing. i would have hoped they wouldn't be _that_ slow

dario-piotrowicz (approved):

vicb:
Thanks for the reviews!

PR #303, opened 2025-01-30
ci: upload playwright reports

vicb:
@dario-piotrowicz or @petebacondarwin: do you have any ideas of what the dev tests are? The only difference seems to be the port? Do we still need them? At least I don't think we should both in CI?


examples/api/e2e/playwright.config.ts
```diff
@@ -50,6 +50,6 @@ export default defineConfig({
     command: "pnpm preview:worker",
     url: "http://localhost:8770",
     reuseExistingServer: !process.env.CI,
-    timeout: 70_000,
+    timeout: 100_000,
```
dario-piotrowicz, line 53: was this necessary? or just in case? just curious 🙂

dario-piotrowicz (approved):
Looks good to me 🙂

examples/api/e2e/playwright.config.ts
vicb: Yep, server keeps not starting

vicb:
Thanks for the review @dario-piotrowicz,
We should hopefully see less CI errors.

PR #307, opened 2025-01-31
error-msg

vicb:
Seen on Discord:

image

I also removed mention to "during development" as it can happen while building the app and shortened the message a bit:

image


james-elicx (approved):

vicb:
Thanks James!

Issue #291, opened 2025-01-28
Investigate e2e failures


vicb:
This repo imported the e2e tests from `@opennextjs/aws` into `examples/e2e`.

The unit tests can be executed by `pnpm e2e-fix`

We need to investigate the e2e failures

- Some are expected - i.e. ISR is not supported yet
- Some are not expected

Notes:

The `pages-router` has a build issue to investigate:

```
[WebServer] ▲ [WARNING] Using direct eval with a bundler is not recommended and may cause problems [direct-eval]
[WebServer] 
[WebServer]     .open-next/server-functions/default/examples/e2e/pages-router/handler.mjs:193620:43:
[WebServer]       193620 │ ...         var deprecatedfn = eval("(function (" + args + ') ...
```

Related PRs:
- https://github.com/opennextjs/opennextjs-cloudflare/pull/289
- https://github.com/opennextjs/opennextjs-cloudflare/pull/290

dario-piotrowicz:
This ticket is mostly done, the only thing remaining is the eval build time warning

dario-piotrowicz:
Quick update on the eval warning, the issue seems to be that Next.js in the pages router uses some libraries `raw-body` and `send` that use the `depd` library to deprecate some functions (so that they display some deprecation messages when called).

This is the problematic code that ends up in the Next.js bundles: [depd code](https://github.com/dougwilson/nodejs-depd/blob/7ad9a48dff1e7fe52d4298da7c1235944073bfc9/index.js#L413-L433)

I'm still investigating, my initial feeling is that these `eval`s are probably not necessary and only needed for the deprecation messages and we could patch them out 🤔

PR #314, opened 2025-01-31
enable `e2e/app-router` tests to run in CI

dario-piotrowicz:
this change renames the `e2e-fix` script of the `e2e/app-router` app to just `e2e` so that it is included as part of our CI checks, in order for it not to fail some tests have been skipped (with appropriate comments explain why)

__

Part of #291


vicb (commented):
Thanks!

It looks like not much is failing outside of what I expected (linked to ISR not implemented).

vicb (approved):
Oops looks like I forgot to approve...

PR #315, opened 2025-01-31
enable e2e/app-pages-router tests to run in CI

dario-piotrowicz:
this change renames the `e2e-fix` script of the `e2e/app-pages-router` app to just `e2e` so that it is included as part of our CI checks, in order for it not to fail some tests have been skipped (with appropriate comments explain why)

____

Part of #291


vicb (approved):
Looks great,
Thanks @dario-piotrowicz!

PR #319, opened 2025-02-02
fix: enable using the `direct` queue for isr

james-elicx:
- Allow use of `direct` for the queue.
- Enable ISR tests for app router e2es.
- Fix e2e isr route so it imports the manifest for the revalidation test.

### DIFF
--- a/.changeset/slimy-houses-punch.md
+++ b/.changeset/slimy-houses-punch.md
@@ -3,3 +3,5 @@
 ---
 
 fix: enable using the `direct` queue for isr
+
+The `direct` mode is not recommended for use in production as it does not de-dupe requests.
--- a/.changeset/slimy-houses-punch.md
+++ b/.changeset/slimy-houses-punch.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+fix: enable using the `direct` queue for isr
--- a/examples/e2e/app-router/app/api/isr/route.ts
+++ b/examples/e2e/app-router/app/api/isr/route.ts
@@ -8,11 +8,21 @@ export const dynamic = "force-dynamic";
 // This endpoint simulates an on demand revalidation request
 export async function GET(request: NextRequest) {
   const cwd = process.cwd();
-  const prerenderManifest = await fs.readFile(path.join(cwd, ".next/prerender-manifest.json"), "utf-8");
-  const manifest = JSON.parse(prerenderManifest);
+
+  let manifest;
+  // this fails at build time when next.js tries to evaluate the route
+  try {
+    // @ts-expect-error
+    const prerenderManifest = await import(/* webpackIgnore: true */ "./.next/prerender-manifest.json");
+    manifest = prerenderManifest.default;
+  } catch {
+    return new Response(null, { status: 500 });
+  }
+
   const previewId = manifest.preview.previewModeId;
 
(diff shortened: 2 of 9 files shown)


vicb:
Awesome 🎉

I'll try to review the PR tomorrow afternoon!

conico974 (commented):
While this will work and could be used right now to make ISR work, i think a better solution long term would be to use a queue system for deduplicating the revalidation.

With the `direct` queue, once a page is stale, every request for this page (while it is revalidating and available worldwide) will trigger a new revalidation. 
KV will probably throw a lot of error because of the 1s write per key limitations

james-elicx:
> While this will work and could be used right now to make ISR work, i think a better solution long term would be to use a queue system for deduplicating the revalidation.
> 
> With the `direct` queue, once a page is stale, every request for this page (while it is revalidating and available worldwide) will trigger a new revalidation. KV will probably throw a lot of error because of the 1s write per key limitations

Agreed, I had wrote on the issue that we should do it as an additional piece of work separately to this - both should be available for people to use ideally.

818Nawaf (approved):

818Nawaf (approved):

examples/e2e/app-router/app/api/isr/route.ts
```diff
@@ -8,11 +8,21 @@ export const dynamic = "force-dynamic";
 // This endpoint simulates an on demand revalidation request
 export async function GET(request: NextRequest) {
   const cwd = process.cwd();
-  const prerenderManifest = await fs.readFile(path.join(cwd, ".next/prerender-manifest.json"), "utf-8");
-  const manifest = JSON.parse(prerenderManifest);
+
+  let manifest;
```
vicb, line 12: Can this change be upstreamed?

818Nawaf (approved):

vicb (commented):
Can you rebase and try to upstream the change?

james-elicx:
> Can you rebase and try to upstream the change?

I've opened https://github.com/opennextjs/opennextjs-aws/pull/728 to upstream it.

.changeset/slimy-houses-punch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+fix: enable using the `direct` queue for isr
```
vicb, line 5: Add a note saying that the direct mode is not recommended for production

packages/cloudflare/src/cli/build/utils/ensure-cf-config.ts
```diff
@@ -12,8 +12,9 @@ export function ensureCloudflareConfig(config: OpenNextConfig) {
     dftMaybeUseCache:
       config.default?.override?.incrementalCache === "dummy" ||
       typeof config.default?.override?.incrementalCache === "function",
-    dftUseDummyTagCacheAndQueue:
-      config.default?.override?.tagCache === "dummy" && config.default?.override?.queue === "dummy",
+    dftUseDummyTagCache: config.default?.override?.tagCache === "dummy",
+    dftMaybeUseQueue:
+      config.default?.override?.queue === "dummy" || config.default?.override?.queue === "direct",
```
vicb, line 17: Print a warning saying that the direct mode is not suitable for production

vicb (approved):
Thanks for the update 🚀

PR #320, opened 2025-02-02
feat: d1 adapter for the tag cache

james-elicx:
- Enable tag/path revalidation and next/after e2es for app directory example.
- Enable passing a function for the tag cache.
- D1 adapter for the tag cache, based on the AWS DynamoDB [adapter](https://github.com/opennextjs/opennextjs-aws/blob/main/packages/open-next/src/overrides/tagCache/dynamodb.ts).
- Extracts a manifest of build-time tags/paths and creates an sql file that can be used to stub D1.
- Adds the env.d.ts file to the tsconfig so that TypeScript recognises env vars defined in there.
- Bumps aws to https://github.com/opennextjs/opennextjs-aws/pull/748

Docs PR - https://github.com/opennextjs/docs/pull/66


packages/cloudflare/src/cli/build/utils/extract-cache-assets-manifest.ts
```diff
@@ -0,0 +1,54 @@
...
+  tags: { [tag: string]: string[] };
+  paths: { [path: string]: string[] };
+};
+
+/**
+ * Extracts a map of the cache assets that were generated during the build.
+ *
+ * The mapping creates an index of each tags pointing to its paths, and each path pointing to its tags.
+ */
+export function extractCacheAssetsManifest(buildOpts: BuildOptions) {
```
conico974, line 27: We already compute all the tags and cache in the aws adapter https://github.com/opennextjs/opennextjs-aws/blob/14b81827f9078b98e32115fb5cfe706d03d64537/packages/open-next/src/build/createAssets.ts#L159-L239
BTW you are missing the fetch cache here

packages/cloudflare/src/api/d1-tag-cache.ts
```diff
@@ -0,0 +1,151 @@
+import { debug, error } from "@opennextjs/aws/adapters/logger.js";
+import type { OpenNextConfig } from "@opennextjs/aws/types/open-next.js";
+import type { TagCache } from "@opennextjs/aws/types/overrides.js";
+
+import { getCloudflareContext } from "./cloudflare-context.js";
+
+// inlined during build
+const manifest = process.env.__OPENNEXT_CACHE_TAGS_MANIFEST;
+
+class D1TagCache implements TagCache {
```
conico974, line 10: I'm not sure that D1 is the right fit for this. The tag cache has been designed around DynamoDB and its strength/limitation.
I'm introducing a new mode for the tag cache that would probably better fit an SQL db https://github.com/opennextjs/opennextjs-aws/pull/717

I think that what would make the most sense would be either using D1 with the new mode (and it avoids having to prepopulate anything) or using this mode with KV (But this would probably mean duplicating some data in KV)

Another thing is that if i'm not mistaken D1 is regional, but the tag cache will be reached from all over the world (And the fetch cache could be used in SSR request as well )

james-elicx: I'm not sure KV would really work that well due to its eventually consistent nature - you get up to 60 seconds on entries continuing to be polled, and if you use a <5 min revalidation time, it essentially becomes pointless trying to use a cache IMO.

D1 is regional yeah - means a little more latency for some regions.

james-elicx: I can rework this PR to be based on the new mode if that makes most sense.

packages/cloudflare/src/cli/build/utils/extract-cache-assets-manifest.ts
james-elicx: Thanks, I didn't realise the manifest was already generated. I'll change it to use that

packages/cloudflare/src/api/d1-tag-cache.ts
conico974: For the new mode there is benefits and drawbacks, one option could be to implement both mode so that each user could use the one that is better suited for them. 
@vicb @dario-piotrowicz What do you think about that ? For reference here is the issue explaining the 2 mode and their benefits/drawback https://github.com/opennextjs/opennextjs-aws/issues/706

818Nawaf (approved):

packages/cloudflare/src/api/d1-tag-cache.ts
vicb: +1 to have the 2 modes if it's not too much dev & maintenance vs the benefits.

We can also merge a single mode when it's ready and implement the other mode later.

packages/cloudflare/src/cli/build/utils/extract-cache-assets-manifest.ts
vicb: This is resolved, right?

examples/e2e/app-router/open-next.config.ts
```diff
@@ -1,15 +1,15 @@
 import type { OpenNextConfig } from "@opennextjs/aws/types/open-next.js";
-import cache from "@opennextjs/cloudflare/kvCache";
+import incrementalCache from "@opennextjs/cloudflare/kvCache";
```
vicb, line 2: thought: we mostly use kebab-case, maybe this should be `kv-cache`?

packages/cloudflare/src/api/d1-tag-cache.ts
```diff
@@ -0,0 +1,151 @@
+import { debug, error } from "@opennextjs/aws/adapters/logger.js";
+import type { OpenNextConfig } from "@opennextjs/aws/types/open-next.js";
+import type { TagCache } from "@opennextjs/aws/types/overrides.js";
+
+import { getCloudflareContext } from "./cloudflare-context.js";
+
+// inlined during build
+const manifest = process.env.__OPENNEXT_CACHE_TAGS_MANIFEST;
```
vicb, line 8: I don't think it's right to access `process.env` at top level as it will only be populated after the first request comes in.

packages/cloudflare/src/api/d1-tag-cache.ts
```diff
@@ -0,0 +1,151 @@
...
+
+class D1TagCache implements TagCache {
+  public name = "d1-tag-cache";
```
vicb, line 11: readonly

packages/cloudflare/src/api/d1-tag-cache.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  public async getByPath(rawPath: string): Promise<string[]> {
+    const { isDisabled, db, table } = this.getConfig();
+    if (isDisabled) return [];
+
+    const path = this.getCacheKey(rawPath);
+
+    try {
+      const { success, results } = await db
+        .prepare(`SELECT tag FROM ${table} WHERE path = ?`)
```
vicb, line 21: can't we use bind for the table name? (SQL injection)

packages/cloudflare/src/api/d1-tag-cache.ts
vicb, line 10: Maybe add some comments about the class

packages/cloudflare/src/api/d1-tag-cache.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  protected getBuildId() {
+    return process.env.NEXT_BUILD_ID ?? "no-build-id";
+  }
+
+  protected mergeTagArrays(...arrays: (string[] | undefined)[]) {
+    const set = new Set<string>();
+
+    for (const arr of arrays) {
+      arr?.forEach((v) => set.add(this.removeBuildId(v)));
```
vicb, line 144: nit: maybe remove the id on the dedup values only to avoid duplicating work

packages/cloudflare/src/cli/build/bundle-server.ts
```diff
@@ -112,6 +113,7 @@ export async function bundleServer(buildOpts: BuildOptions): Promise<void> {
       "process.env.NEXT_RUNTIME": '"nodejs"',
       "process.env.NODE_ENV": '"production"',
       "process.env.NEXT_MINIMAL": "true",
+      "process.env.__OPENNEXT_CACHE_TAGS_MANIFEST": `${JSON.stringify(extractCacheAssetsManifest(buildOpts))}`,
```
vicb, line 116: @conico974 do you think this could be added to the config adapter?

conico974: You mean the `open-next.config.ts` ? Not sure how it would work since it is bundled before `next build`

One thing about inlining the manifest here is that this could become pretty big, especially for website with a lot of page.
For example the `__NT_/layout` tag is on every page.
Inlining this will likely become an issue at scale

vicb: I was thinking about `adapter/config` but the size issue would be the same.

So we should preload the DB?

conico974: Would it be possible to generate a SQL file instead and run wrangler execute as part of the deployment ? This seems like a better idea
Or maybe using the rest API with the initialization function from OpenNext

packages/cloudflare/src/api/d1-tag-cache.ts
james-elicx: This is inlined during the build

packages/cloudflare/src/api/d1-tag-cache.ts
james-elicx: Unfortunately not - it gives a syntax error. The table is driven by an env var, so i wouldnt think sql injection will be a problem

examples/e2e/app-router/open-next.config.ts
james-elicx: agreed

packages/cloudflare/src/api/d1-tag-cache.ts
```diff
@@ -0,0 +1,151 @@
...
+
+    try {
+      const { success, results } = await db
+        .prepare(`SELECT tag FROM ${table} WHERE path = ? AND revalidatedAt > ?`)
+        .bind(this.getCacheKey(path), lastModified ?? 0)
+        .all<{ tag: string }>();
+
+      if (!success) throw new Error(`D1 select failed for ${path} - ${lastModified ?? 0}`);
+
+      const tags = results?.map((item) => this.removeBuildId(item.tag)) ?? [];
```
conico974, line 89: This is not needed, we could just use `results.length > 0`

packages/cloudflare/src/api/d1-tag-cache.ts
vicb: It should not... but maybe we can hardcode `tags` here anyway.

james-elicx: I did this with the idea that people might want to use a different table name (multiple projects using same db), so it was using the env var and defaulting to `tags` if not present. I can change to always hardcode to tags if you want, but I feel like it's a useful bit of a configurability to have?

packages/cloudflare/src/cli/build/utils/extract-cache-assets-manifest.ts
```diff
@@ -0,0 +1,52 @@
...
+  paths: { [path: string]: string[] };
+};
+
+/**
+ * Extracts a map of the cache assets that were generated during the build.
+ *
+ * The mapping creates an index of each tags pointing to its paths, and each path pointing to its tags.
+ */
+export function extractCacheAssetsManifest(options: BuildOptions): CacheAssetsManifest {
+  // TODO: Expose the function for getting this data as an adapter-agnostic utility in AWS.
```
conico974, line 23: That's kind of the point of the `initializationFunction` https://opennext.js.org/aws/inner_workings/components/initializer
With a custom wrapper and a custom tag cache it could be turned into a function.
Not entirely sure it's worth it though

packages/cloudflare/src/api/d1-tag-cache.ts
james-elicx: good point

packages/cloudflare/src/cli/build/utils/extract-cache-assets-manifest.ts
james-elicx: Does that run at build time or run time? The docs aren't clear

conico974: It doesn't run at all as part of OpenNext actually. On aws it runs during deployment, either as a CDK custom resource or even as a simple lambda that can get called by IAC.
We just provide the function and it's the task of the IAC to use it.

james-elicx: I see. It doesn't sound like it would quite work at the moment then. I guess this is all pointing at needing a convenient way to interact with bindings at build-time.

conico974: Yeah i think for now the solution would be on OpenNext side to create a sql file on every build (instead of the cache manifest) and ask people to change their deploy command so that they run `wrangler d1 execute NEXT_CACHE_D1 --remote --file generated.sql` before running wrangler deploy.

james-elicx: @vicb thoughts?

packages/cloudflare/src/api/d1-tag-cache.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  public async writeTags(tags: { tag: string; path: string; revalidatedAt?: number }[]): Promise<void> {
+    const { isDisabled, db, table } = this.getConfig();
+    if (isDisabled || tags.length === 0) return;
+
+    try {
+      const results = await db.batch(
+        tags.map(({ tag, path, revalidatedAt }) =>
+          db
+            .prepare(`INSERT INTO ${table} (tag, path, revalidatedAt) VALUES(?, ?, ?)`)
```
conico974, line 94: There is something that feels wrong here.
Revalidating a layout (which is one of the worst case scenario) means that we'll be rewriting the whole table.
Especially during testing this may backfire really quick, write are 1000x more expensive than read on D1.

Just throwing some ideas here, i need to think about this more thoroughly :

We could improve this by having 2 tables, this one where we'll remove the `revalidatedAt` and another one that would only contains `tag` and `revalidatedAt` that we could join when needed.

james-elicx: The tag cache types for the writeTags function led me to believe that the revalidation times are for unique combinations of tags+paths and not solely based on tags. From what you're saying though, it sounds like the revalidated times are solely for tags in the tag cache? Sorry, I'm not too familiar with the cache stuff in AWS. If that's the case, would that mean we can essentially ignore the path here - are all possible path->tag mappings being gathered as part of the step that you have for getting the initial data for dynamo?

If that's the case, what you're suggesting sounds like a pretty good optimisation. One table for the mapping, another for the revalidation times, and just ignore the path in this function?

packages/cloudflare/src/cli/build/open-next/compile-cache-assets-manifest.ts
```diff
@@ -0,0 +1,37 @@
...
+  const stmts = [
+    `CREATE TABLE IF NOT EXISTS ${table} (tag TEXT NOT NULL, path TEXT NOT NULL, revalidatedAt INTEGER NOT NULL, UNIQUE(tag, path) ON CONFLICT REPLACE);`,
+  ];
+
+  if (existsSync(rawManifestPath)) {
+    const rawManifest: RawManifest = JSON.parse(readFileSync(rawManifestPath, "utf-8"));
+
+    rawManifest.forEach(({ tag: { S: tag }, path: { S: path }, revalidatedAt: { N: revalidatedAt } }) => {
+      stmts.push(
+        `INSERT INTO ${table} (tag, path, revalidatedAt) VALUES ('${tag}', '${path}', ${revalidatedAt});`
```
vicb, line 31: Can you rather generate `INSERT INTO ${table} (tag, path, revalidatedAt) VALUES (...), (...), (...);` and replace `'${var}'` with `${JSON.stringify(var)}`

packages/cloudflare/src/cli/build/open-next/compile-cache-assets-manifest.ts
```diff
@@ -0,0 +1,37 @@
...
+  revalidatedAt: { N: string };
+}[];
+
+/**
+ * Generates SQL statements that can be used to initialise the cache assets manifest in an SQL data store.
+ */
+export function compileCacheAssetsManifestSqlFile(options: BuildOptions) {
+  // TODO: Expose the function for getting this data as an adapter-agnostic utility in AWS.
+  const rawManifestPath = path.join(options.outputDir, "dynamodb-provider", "dynamodb-cache.json");
+  const outputPath = path.join(options.outputDir, "cache-assets-manifest.sql");
```
vicb, line 18:
```suggestion
  const outputPath = path.join(options.outputDir, "cloudflare/cache-assets-manifest.sql");
```

I'd like to move the env vars there too and probably some init code later

packages/cloudflare/src/api/d1-tag-cache.ts
conico974: As i previously said, this tag cache was primarily meant to work with DynamoDB, with its strength and weakness.

> would that mean we can essentially ignore the path here

In this function yeah, if we get a `revalidatedAt` equals to `1` it means that it's a new tag/path mapping from `set`, no need to insert the revalidatedAt here.
If we get no value, only `tags`/`revalidatedAt` matters and `revalidatedAt` should be set to `Date.now()` 
Btw we may want to use built in sql for setting the date to avoid issue with `Date.now()` in workers. It's probably not a big deal, but worth mentioning.
I should have added more docs and maybe refactor this a bit as well;

> are all possible path->tag mappings being gathered as part of the step that you have for getting the initial data for dynamo

For the one generated at build time yeah, some other could be generated later during runtime

packages/cloudflare/src/cli/build/open-next/compile-cache-assets-manifest.ts
```diff
@@ -0,0 +1,40 @@
...
+  if (existsSync(rawManifestPath)) {
+    const rawManifest: RawManifest = JSON.parse(readFileSync(rawManifestPath, "utf-8"));
+
+    const values = rawManifest.map(
+      ({ tag, path, revalidatedAt }) =>
+        `(${JSON.stringify(tag.S)}, ${JSON.stringify(path.S)}, ${revalidatedAt.N})`
+    );
+
+    if (values.length) {
+      appendFileSync(outputPath, `INSERT INTO tags (tag, path, revalidatedAt) VALUES ${values.join(", ")};`);
```
vicb, line 37:
```suggestion
      appendFileSync(outputPath, `INSERT INTO ${JSON.stringify(table)} (tag, path, revalidatedAt) VALUES ${values.join(", ")};`);
```

Also replace `${table}` with `${JSON.stringify(table)}` across this PR

packages/cloudflare/src/api/d1-tag-cache.ts
james-elicx: Thanks so much for the explanation. Was super useful info. I've made some changes to use two tables based on that. I think it's a much better approach.

conico974 (commented):
It could be interesting to add a little script to run the `wrangler execute` command, maybe with an option to automatically drop the old table as well

packages/cloudflare/src/cli/build/build.ts
```diff
@@ -86,8 +87,12 @@ export async function build(projectOpts: ProjectOptions): Promise<void> {
   createStaticAssets(options);
 
   if (config.dangerous?.disableIncrementalCache !== true) {
-    createCacheAssets(options);
+    const { useTagCache, metaFiles } = createCacheAssets(options);
     copyCacheAssets(options);
+
+    if (useTagCache) {
```
conico974, line 93: I've just realised that there is an issue in aws with that. We always check the tag cache even for page router (where it's not used), and this might cause an issue here if someone use this tag cache while only using page router. `useTagCache` would be false and it will logs a ton of error because of table not created.
I'll add the fix to this PR https://github.com/opennextjs/opennextjs-aws/pull/717, you should add something in the docs anyway that tag cache is not used for page router and all of this is useless in this case. (i.e. use `dummy`)

conico974 (approved):
LGTM, nice work!!
To be perfectly honest, i'm still not 100% convinced that D1 is the perfect fit for this (at least not alone). The tag cache can be used in SSR for the fetch cache and i'm not sure if D1 can handle heavy workload for these cases.
It's probably fine in most cases but i still wanted to point this out

packages/cloudflare/src/api/d1-tag-cache.ts
```diff
@@ -0,0 +1,173 @@
...
+      .dangerous?.disableTagCache;
+
+    if (!db || isDisabled) {
+      return { isDisabled: true as const };
+    }
+
+    return {
+      isDisabled: false as const,
+      db,
+      table: {
```
vicb, line 153:
```suggestion
      tables: {
```

packages/cloudflare/src/api/d1-tag-cache.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  public async getByPath(rawPath: string): Promise<string[]> {
+    const { isDisabled, db, table } = this.getConfig();
+    if (isDisabled) return [];
+
+    const path = this.getCacheKey(rawPath);
+
+    try {
+      const { success, results } = await db
+        .prepare(`SELECT tag FROM ${table.tags} WHERE path = ?`)
```
vicb, line 36: Can you please update the PR to always use `JSON.stringify(tableName)`

vicb (commented):
I added a couple comments.

IIUC we should first get https://github.com/opennextjs/opennextjs-aws/pull/717 merged and then this PR?

Even if using D1 is not ideal, we can keep iterating before 1.0 but it would be nice to unblock people as early as we can.

WDYT @conico974 @james-elicx ?

conico974:
> IIUC we should first get https://github.com/opennextjs/opennextjs-aws/pull/717 merged and then this PR?

Not necessarily, but it would avoid people getting a bunch of errors if they use this with page router only. And it avoids checking the cache tag for  page router.

>Even if using D1 is not ideal, we can keep iterating before 1.0 but it would be nice to unblock people as early as we can.

I agree, and it will allow to collect some feedback

packages/cloudflare/src/api/d1-tag-cache.ts
james-elicx: I did in each place before, but conico asked me to move all the JSON.stringifies to a single place, so I moved it to the part where the variable is defined. I'll move them all back.

vicb:
@conico974 

> [opennextjs/opennextjs-aws#717](https://github.com/opennextjs/opennextjs-aws/pull/717)

What's missing for this PR to get merged? Do you have an approximate ETA in mind?

conico974:
@vicb it should be ready actually, i just requested a review

james-elicx:
> IIUC we should first get [opennextjs/opennextjs-aws#717](https://github.com/opennextjs/opennextjs-aws/pull/717) merged and then this PR?

I thought we had agreed that we would add both modes with users having the option of deciding which they want to use, so there shouldn't be any dependency on the new mode for this PR. Unless we want to add the other mode in this PR as well and release both modes at the same time?

vicb:
> > IIUC we should first get [opennextjs/opennextjs-aws#717](https://github.com/opennextjs/opennextjs-aws/pull/717) merged and then this PR?
> 
> I thought we had agreed that we would add both modes with users having the option of deciding which they want to use, so there shouldn't be any dependency on the new mode for this PR. Unless we want to add the other mode in this PR as well and release both modes at the same time?

I was referring to "Not necessarily, but it would avoid people getting a bunch of errors if they use this with page router only. And it avoids checking the cache tag for page router."

packages/cloudflare/src/api/d1-tag-cache.ts
```diff
@@ -0,0 +1,175 @@
...
+import type { OpenNextConfig } from "@opennextjs/aws/types/open-next.js";
+import type { TagCache } from "@opennextjs/aws/types/overrides.js";
+import { RecoverableError } from "@opennextjs/aws/utils/error.js";
+
+import { getCloudflareContext } from "./cloudflare-context.js";
+
+/**
+ * An instance of the Tag Cache that uses a D1 binding (`NEXT_CACHE_D1`) as it's underlying data store.
+ *
+ * This table should be populated using an SQL file that is generated during the build process.
```
vicb, line 11: nit: I think this should be moved to the "**Tag/path mappings**" section?

vicb (approved):
Thanks a lot for creating this PR and all of the updates!μ

LGTM, a couple minor comments.

🚀

packages/cloudflare/src/api/d1-tag-cache.ts
```diff
@@ -0,0 +1,175 @@
...
+ *
+ * **Tag/path mappings**
+ *
+ * Information about the relation between tags and paths is stored in a `tags` table that contains
+ * two columns; `tag`, and `path`. The table name can be configured with `NEXT_CACHE_D1_TAGS_TABLE`
+ * environment variable.
+ *
+ * **Tag revalidations**
```
james-elicx, line 19:
```suggestion
 * **Tag revalidations table**
```

packages/cloudflare/src/api/d1-tag-cache.ts
```diff
@@ -0,0 +1,175 @@
...
+/**
+ * An instance of the Tag Cache that uses a D1 binding (`NEXT_CACHE_D1`) as it's underlying data store.
+ *
+ * This table should be populated using an SQL file that is generated during the build process.
+ *
+ * **Tag/path mappings table**
+ *
+ * Information about the relation between tags and paths is stored in a `tags` table that contains
+ * two columns; `tag`, and `path`. The table name can be configured with `NEXT_CACHE_D1_TAGS_TABLE`
+ * environment variable.
```
james-elicx, line 17:
```suggestion
 * **Tag/path mappings table**
 *
 * Information about the relation between tags and paths is stored in a `tags` table that contains
 * two columns; `tag`, and `path`. The table name can be configured with `NEXT_CACHE_D1_TAGS_TABLE`
 * environment variable.
 *
 * This table should be populated using an SQL file that is generated during the build process.
```

PR #323, opened 2025-02-03
fix the encoding of __NEXT_PRIVATE_STANDALONE_CONFIG

vicb:


james-elicx (approved):

james-elicx:
For reference for others, the part where Next.js does a JSON.parse on it - https://github.com/vercel/next.js/blob/58e1bd29a1563032244ee6d3ca689eb255756e0d/packages/next/src/server/config.ts#L1060

PR #334, opened 2025-02-04
fix external rewrites to Cloudflare hosted resources not working

dario-piotrowicz:
rewrites to external resources hosted on Cloudflare don't currently work, this PR addresses that by bumping the aws dependency

For more context see: https://github.com/opennextjs/opennextjs-aws/pull/724


vicb (changes requested):
Could you update the aws package so that we get the test result?

vicb (approved):

PR #342, opened 2025-02-05
refactor: simplify the e2e and script configs

vicb:
- On CI, e2e was rebuilding the apps (via `preview`), drop that as there is a build step before
- Factor the playwright configs
- Automatically assign a different port to each app - no need to hardcode ports in the package.json any more
- Do not run `next dev` tests on multiple browsers
- Remove `dev:worker`
- Rename `preview:worker` to `preview` - to be consistent with c3 templates

This should make the e2e CI job much faster (timeout is reduced)


examples/common/config-e2e.ts
```diff
@@ -0,0 +1,90 @@
...
+import type nodeProcess from "node:process";
+import { getAppPort, getInspectorPort, type AppName } from "./apps";
+
+declare const process: typeof nodeProcess;
+
+export function configurePlaywright(
+  app: AppName,
+  {
+    // Do we run on CI?
+    isCI = false,
```
dario-piotrowicz, line 11: why is this an argument and not just computed here?
you always do pass it as `!!process.env.CI` so it would just make sense to do that directly in this file, wouldn't it?

examples/common/apps.ts
dario-piotrowicz, line 1: very nice idea for not hardcoding the ports! 🫶

examples/common/apps.ts
```diff
@@ -0,0 +1,44 @@
...
+  "vercel-commerce",
+  // // e2e
+  "app-pages-router",
+  "app-router",
+  "pages-router",
+  // // bugs
+  "gh-119",
+  "gh-219",
+  "gh-223",
+] as const;
```
dario-piotrowicz: I don't really like this listing and I don't think it is necessary

I think things could be simplified here by either:
 - (worse) keeping the indexes in the global scope and increment them in `getAppPort` and `getInspectorPort`
 - (better) implementing a generic "getNextAvailablePort(start: number)" that both `getAppPort` and `getInspectorPort` use
 
 I think that with either of the above the solution here would be made more generic and easier to maintain (the `getNextAvailablePort` would also make this more robust)
 
 Anyways this is already an improvement to what we currently have, so if my comment is not blocking

examples/common/apps.ts
```diff
@@ -0,0 +1,44 @@
...
+  "api",
+  "create-next-app",
+  "middleware",
+  "vercel-blog-starter",
+  "vercel-commerce",
+  // // e2e
+  "app-pages-router",
+  "app-router",
+  "pages-router",
+  // // bugs
```
vicb, line 13:
```suggestion
  // e2e
  "app-pages-router",
  "app-router",
  "pages-router",
  // bugs
```

dario-piotrowicz (approved):
Very nice improvement 🫶

dario-piotrowicz (approved):
Very nice improvement 🫶

examples/common/config-e2e.ts
```diff
@@ -0,0 +1,90 @@
...
+    },
+
+    projects,
+
+    /* Run your local dev server before starting the tests */
+    webServer: {
+      command,
+      url: baseURL,
+      reuseExistingServer: !isCI,
+      timeout: 500_000,
```
vicb, line 87:
```suggestion
      timeout,
```

examples/common/package.json
dario-piotrowicz, line 1: is this file really needed?

examples/common/config-e2e.ts
vicb: In general, I don't link that functions have external dependencies.
Injecting the value enable:
- writing tests,
- run on any platform (i.e. outside of GH actions)

dario-piotrowicz:
@vicb this is unrelated to this PR but I wanted to mention that here I notice that we use `npm` instead of `pnpm` for some of the examples, we even have some package-lock files: https://github.com/opennextjs/opennextjs-cloudflare/blob/simplify-configs/examples/bugs/gh-219/package-lock.json, that's (I hope we all agree) clearly wrong and we should just consolidate everything to use the pnpm workspace

examples/common/apps.ts
vicb: > implementing a generic "getNextAvailablePort(start: number)" that both getAppPort and getInspectorPort use

There is no common state across apps, so we would need to sync using i.e. a file. Not sure if worth the burden.

examples/common/package.json
vicb: It solves a build error (I first had the file only).

vicb:
> @vicb this is unrelated to this PR but I wanted to mention that here I notice that we use `npm` instead of `pnpm` for some of the examples, we even have some package-lock files: https://github.com/opennextjs/opennextjs-cloudflare/blob/simplify-configs/examples/bugs/gh-219/package-lock.json, that's (I hope we all agree) clearly wrong and we should just consolidate everything to use the pnpm workspace

Good catch.

Yep, we should use `pnpm` everywhere. I didn't pay attention when importing the apps for e2e.
I'll handle that in a follow up PR.

vicb:
Thanks for the review Dario!

Issue #328, opened 2025-02-03
Error when running the e2e for app-router (Fetch cache)


vicb:
@dario-piotrowicz I am seeing this error when running `pnpm -F app-router e2e` (locally and on the [CI](https://github.com/opennextjs/opennextjs-cloudflare/actions/runs/13108919054/job/36568439435?pr=325)).

Do you have any idea of the root cause?

Thanks!

dario-piotrowicz:
No idea, as you can see that passed in the PR in which I enabled the app-router e2es: https://github.com/opennextjs/opennextjs-cloudflare/pull/314 😕 

I'll have a look

james-elicx:
I've found it to be a bit intermittent

PR #340, opened 2025-02-04
debugging #328

dario-piotrowicz:
I don't seem to be able to debug #328 locally so I am opening this PR just to see the logs in CI when it fails

> [!WARNING]
> This PR is just for debugging, please don't review it

PR #344, opened 2025-02-05
Do not serve stale values from the assets when there is no KV

dario-piotrowicz:
This change makes sure that stale cache values from assets are not served when there is no KV binding.

This makes the issue in https://github.com/opennextjs/opennextjs-cloudflare/issues/328 no longer flaky (it makes the test always fail) since before it was passing because stale cache values were being served from assets (so the fetch cached result would always stay the same, making this [check](https://github.com/opennextjs/opennextjs-cloudflare/blob/420b598ac242ff91768685d312c9940335f3a120/examples/e2e/app-router/e2e/ssr.test.ts#L37) pass). The reason why KV is not being used is still being investigated.

No other tests besides the [328 one](https://github.com/opennextjs/opennextjs-cloudflare/blob/420b598ac242ff91768685d312c9940335f3a120/examples/e2e/app-router/e2e/ssr.test.ts#L31-L38) is affected as far as I know, since the 328 one was the only flaky we had, regarding all the other tests, the ones that were passing before as still passing after my change, and the ones that were failing before (and we skip) are failing after my change.

### DIFF
--- a/.changeset/spotty-baboons-rule.md
+++ b/.changeset/spotty-baboons-rule.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+Fix: make sure `kvCache` discards expired entries
--- a/.changeset/spotty-baboons-rule.md
+++ b/.changeset/spotty-baboons-rule.md
@@ -2,4 +2,4 @@
 "@opennextjs/cloudflare": patch
 ---
 
-Fix: make sure `kvCache` discards expired entries
+Fix: make sure that the kvCache doesn't serve stale cache values from assets when there is no KV binding
--- a/packages/cloudflare/src/api/kvCache.ts
+++ b/packages/cloudflare/src/api/kvCache.ts
@@ -1,9 +1,4 @@
-import type {
-  CachedFetchValue,
-  CacheValue,
-  IncrementalCache,
-  WithLastModified,
-} from "@opennextjs/aws/types/overrides";
+import type { CacheValue, IncrementalCache, WithLastModified } from "@opennextjs/aws/types/overrides";
 import { IgnorableError, RecoverableError } from "@opennextjs/aws/utils/error.js";
 
 import { getCloudflareContext } from "./cloudflare-context.js";
@@ -70,12 +65,15 @@ class Cache implements IncrementalCache {
           // The cache can not be updated when there is no KV
           // As we don't want to keep serving stale data for ever,
           // we pretend the entry is not in cache
-          const entryValue = entry?.value as CachedFetchValue;
-          if (entryValue?.kind === "FETCH") {
-            const expires = entryValue.data.headers?.expires;
-            const expiresTime = new Date(expires as string).getTime();
+          if (
+            entry?.value &&
+            "kind" in entry.value &&
+            entry.value.kind === "FETCH" &&
+            entry.value.data?.headers?.expires
+          ) {
+            const expiresTime = new Date(entry.value.data.headers.expires).getTime();
             if (!isNaN(expiresTime) && expiresTime <= Date.now()) {
-              this.debug(`found expired entry (expire time: ${expires})`);
+              this.debug(`found expired entry (expire time: ${entry.value.data.headers.expires})`);
               return null;
             }
           }
--- a/packages/cloudflare/src/api/kvCache.ts
+++ b/packages/cloudflare/src/api/kvCache.ts
@@ -68,7 +68,7 @@ class Cache implements IncrementalCache {
         }
         if (!kv) {
           // The cache can not be updated when there is no KV
-          // As we don't want to keep serving stale data for ever, 
+          // As we don't want to keep serving stale data for ever,
           // we pretend the entry is not in cache
           const entryValue = entry?.value as CachedFetchValue;
           if (entryValue?.kind === "FETCH") {
--- a/packages/cloudflare/src/api/kvCache.ts
+++ b/packages/cloudflare/src/api/kvCache.ts
@@ -1,4 +1,9 @@
-import type { CacheValue, IncrementalCache, WithLastModified } from "@opennextjs/aws/types/overrides";
+import type {
+  CachedFetchValue,
+  CacheValue,
+  IncrementalCache,
+  WithLastModified,
+} from "@opennextjs/aws/types/overrides";
 import { IgnorableError, RecoverableError } from "@opennextjs/aws/utils/error.js";
 
 import { getCloudflareContext } from "./cloudflare-context.js";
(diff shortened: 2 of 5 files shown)


packages/cloudflare/src/api/kvCache.ts
```diff
@@ -62,6 +79,17 @@ class Cache implements IncrementalCache {
           };
         }
       }
+
+      const entryValue = entry?.value as CachedFetchValue | undefined;
```
vicb, line 83: Why do we need `as CachedFetchValue` here? 
Couldn't this be changed upstream?

packages/cloudflare/src/api/kvCache.ts
```diff
@@ -62,6 +79,17 @@ class Cache implements IncrementalCache {
...
+      if (entryValue?.kind === "FETCH") {
```
vicb, line 84: Don't we only want to do this when the value comes from the assets?

vicb (commented):
Could you add a comment in the file about what the code is doing an why?

> no longer flaky (it makes the test always fail)

Is that an improvement?

packages/cloudflare/src/api/kvCache.ts
dario-piotrowicz: why wouldn't we want to check the expiration date also for entries from the KV?

vicb: My understanding is that this is done by the caller.

My understanding of the issue: when an entry is expired, Next will use the [stalled entry](https://nextjs.org/docs/app/building-your-application/caching#time-based-revalidation) and update the cache in the background.

If there is no KV then we can not update the cache and we will keep serving the stalled entry.

Then I think that we don't want to use the stalled entry in this case only.

So I think my previous comment should be

```diff
-Don't we only want to do this when the value comes from the assets?
+Don't we only want to do this when there is no KV?
```

dario-piotrowicz:
> Could you add a comment in the file about what the code is doing an why?

ok

> Is that an improvement?

Yes, this is fixing a false positive we currently have, locally things are not working as intended, we are serving state cached data which make the test app-router fetch cache test pass, with this change the state data is no longer served so the underlying broken catching logic gets exposed

packages/cloudflare/src/api/kvCache.ts
dario-piotrowicz: I've upstreamed the type, however I still need the `as CachedFetchValue` since typescript is unable to narrow down the type (because of the generic `IsFetch`) 😖

dario-piotrowicz:
> Is that an improvement?

I thought we discussed this and agreed that not keep always serving the same stale build time cache data was the correct behavior, if you disagree I'm fine closing the PR

vicb:
> not keep always serving the same stale build time cache data was the correct behavior, if you disagree I'm fine closing the PR

I agree with that 100%.

My question was if going from flaky to failing is an improvement.

I think the outcome of that PR should be that the tests are passing?

packages/cloudflare/src/api/kvCache.ts
```diff
@@ -61,7 +66,20 @@ class Cache implements IncrementalCache {
...
+        // is missing or not working properly we end up always serving the same stale build time generated cache entries
+        const entryValue = entry?.value as CachedFetchValue;
+        if (entryValue?.kind === "FETCH") {
+          const expires = entryValue.data.headers?.expires;
+          const expiresTime = new Date(expires as string).getTime();
+          if (!isNaN(expiresTime) && expiresTime <= Date.now()) {
+            this.debug(`found expired entry (expire time: ${expires})`);
+            return null;
+          }
+        }
```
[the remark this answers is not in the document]
dario-piotrowicz: > wondering if
> ...
> Can not be rewritten without the cast as
> ...

No, that still requires some type casting or a type guard

818Nawaf (approved):

packages/cloudflare/src/api/kvCache.ts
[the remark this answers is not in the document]
james-elicx: I think a stricter check instead of the type cast would probably be preferable. Something like follows would work:
```ts
if (entryValue && "kind" in entryValue && entryValue.kind === "FETCH") { ... }
```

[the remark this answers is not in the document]
james-elicx: it would also be nice to get rid of the type cast for `expires as string` too. The existence of the `expires` value could be incorporated into the `if` statement's condition as well.

.changeset/spotty-baboons-rule.md
```diff
@@ -0,0 +1,5 @@
+---
```
vicb, line 1: maybe merge the 2 changesets, they are related

vicb:
> > My question was if going from flaky to failing is an improvement.
> 
> Well I see it as a temporary improvement, my plan here is to go from flaky-difficult-to-test bug -> consistently failing (easier to work on) bug -> fixed bug
> 
> > I think the outcome of that PR should be that the tests are passing?
> 
> The outcome of this PR is to fix the fact that now we can serve static stale assets forever, as a side effect this makes working on the fetch cache bug easier, that will be addressed in a separate PR (probably #340)

The PR is about not serving stale cache values from assets when there is no KV binding. So I think the title and description of the PR must be updated as well as the changeset.

You could also expand the description of the PR to tell more about why/which of the (skipped) tests always fail after this PR:

- maybe we don't really understand and it's fine to say that we will investigate
- if we have some clue, it would be good to update the linked issue

The kind of information that I am looking for is why this PR would make the tests always fail - it is supposed to change the behavior only when there is no kv binding but we have [a kv binding in the tests](https://github.com/opennextjs/opennextjs-cloudflare/blob/420b598ac242ff91768685d312c9940335f3a120/examples/e2e/app-router/wrangler.json#L11-L16).

I hope this helps.

dario-piotrowicz:
> The PR is about not serving stale cache values from assets when there is no KV binding. So I think the title and description of the PR must be updated as well as the changeset.

Updated 👍 

> You could also expand the description of the PR to tell more about why/which of the (skipped) tests always fail after this PR:
> 
> * maybe we don't really understand and it's fine to say that we will investigate
> * if we have some clue, it would be good to update the linked issue
> 
> The kind of information that I am looking for is why this PR would make the tests always fail - it is supposed to change the behavior only when there is no kv binding but we have [a kv binding in the tests](https://github.com/opennextjs/opennextjs-cloudflare/blob/420b598ac242ff91768685d312c9940335f3a120/examples/e2e/app-router/wrangler.json#L11-L16).

I've added the information, I hope it helps

vicb (approved):
Thanks a lot for the updates, super clear 🙏

.changeset/spotty-baboons-rule.md
dario-piotrowicz: would having the 2 changesets merged have some benefit? or put differently, does having 2 separate changesets have downsides?

dario-piotrowicz:
Thanks for the approval @vicb 👍 

We can merge this PR, or if you think it would be safer we can be merged when issue https://github.com/opennextjs/opennextjs-cloudflare/issues/328 is fully understood, either is totally fine by me

.changeset/spotty-baboons-rule.md
vicb: Up to you

vicb:
Let's merge it

PR #349, opened 2025-02-06
fix: use intended asset cache dir

DaniFoldi:
Hi :wave:,

I noticed the typo during a build output, and I think this wanted to be the `cdn-cgi` special cf path.


dario-piotrowicz (approved):
looks good to me, thanks a lot for the fix @DaniFoldi 🫶

vicb (approved):

PR #352, opened 2025-02-07
refactor: make inlineRequirePagePlugin async

vicb:
ESBuild plugin should be async: https://esbuild.github.io/plugins/#on-load

### DIFF
--- a/packages/cloudflare/src/cli/build/patches/plugins/require-page.ts
+++ b/packages/cloudflare/src/cli/build/patches/plugins/require-page.ts
@@ -1,4 +1,3 @@
-import { existsSync, readFileSync } from "node:fs";
 import { readFile } from "node:fs/promises";
 import { join } from "node:path";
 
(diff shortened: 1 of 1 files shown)


packages/cloudflare/src/cli/build/patches/plugins/require-page.ts
```diff
@@ -20,27 +19,37 @@ export function inlineRequirePagePlugin(buildOpts: BuildOptions) {
...
-  const appPathsManifests: string[] = existsSync(appPathsManifestFile)
-    ? Object.values(JSON.parse(readFileSync(appPathsManifestFile, "utf-8")))
-    : [];
+  let pagesManifests: string[] = [];
+  try {
+    pagesManifests = Object.values(JSON.parse(await readFile(pagesManifestFile, "utf-8")));
+  } catch {
+    // The file does not exists
+    pagesManifests = [];
+  }
```
dario-piotrowicz, line 43: just wanted to point out that this (and the one below) `catch` block catches any exception, so it's not just if the file doesn't exist but for example if the file is malformed (and `JSON.parse` throws), I think that's fine here, but the `// The file does not exists` might be a bit off/misleading in my opinion

dario-piotrowicz (approved):
Looks good to me 😄

packages/cloudflare/src/cli/build/patches/plugins/require-page.ts
vicb: That's right but if this file is broken, the app would be broken anyway

vicb:
Thanks for the review!

PR #354, opened 2025-02-07
perf: reduce the amount of code converted to AST

vicb:
We only patch files (vs the whole bundle) to patch instrumentation

Changes in this PR:
- move the instrumentation patch to the build and on files vs the whole bundle
- create an ESBuild plugin to apply multiple content modifications
- move one plugin to the plugins folder
- drop ts-morph

I did that because users are reporting OOM on DIscord while patching - the PR should reduces CPU and memory usage by minimizing parsing

@wesbos said on Discord: 

- "locally the memory usage went from 3.5gb to 700mb. Trying a deploy now"
- "That worked! And it seems shaved off ~60s from my build too. Thank you!" (fixed a broken build on Workers build)


wesbos:
This PR fixed my memory issues - thank you!

vicb:
> This PR fixed my memory issues - thank you!

I have a lot of memory issues, I should try :)

james-elicx (approved):
The content updater for multiple plugins is a pretty neat idea. Nice, looks good to me.

PR #356, opened 2025-02-09
fix:Crypto works under node18

nojiritakeshi:
When using opennextjs-cloudflare under node18, I get the following error in “crypto”.
```
file:///home/nojiri/yourself-score/node_modules/@opennextjs/cloudflare/dist/cli/utils/ask-confirmation.js:3
    const questionName = crypto.randomUUUID();

ReferenceError: crypto is not defined
```
Until node@v18 is deprecated, we think it's ok to import.

### DIFF
--- a/.changeset/fifty-parrots-retire.md
+++ b/.changeset/fifty-parrots-retire.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+Crypto does not generate errors even when node version is less than 18.
--- a/.changeset/fifty-parrots-retire.md
+++ b/.changeset/fifty-parrots-retire.md
@@ -2,4 +2,4 @@
 "@opennextjs/cloudflare": patch
 ---
 
-Crypto does not generate errors even when node version is less than 18.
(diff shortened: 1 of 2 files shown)


.changeset/fifty-parrots-retire.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+Crypto does not generate errors even when node version is less than 18.
```
vicb, line 5:
```suggestion
import `randomUUID` from `node:crypto` to support NodeJS 18
```

packages/cloudflare/src/cli/utils/ask-confirmation.ts
```diff
@@ -1,7 +1,10 @@
+// TODO: Drop this file when Node.js 18 is deprecated
```
vicb, line 1:
```suggestion
```

vicb (approved):

vicb:
Thanks for your PR @nojiritakeshi

PR #360, opened 2025-02-09
feat: basic in-memory de-duping revalidation queue

james-elicx:
Super basic de-duping on a per-isolate basis for the queue.

This approach uses the MessageGroupId for the de-duping key as it is based on just the path, and the key for ISR pages in the incremental cache is also the path, so we can compute the correct key to free from the list of revalidated paths.

### DIFF
--- a/packages/cloudflare/src/api/kv-cache.ts
+++ b/packages/cloudflare/src/api/kv-cache.ts
@@ -2,7 +2,6 @@ import type { CacheValue, IncrementalCache, WithLastModified } from "@opennextjs
 import { IgnorableError, RecoverableError } from "@opennextjs/aws/utils/error.js";
 
 import { getCloudflareContext } from "./cloudflare-context.js";
-import memoryQueue from "./memory-queue.js";
 
 export const CACHE_ASSET_DIR = "cdn-cgi/_next_cache";
 
@@ -117,8 +116,6 @@ class Cache implements IncrementalCache {
       );
     } catch {
       throw new RecoverableError(`Failed to set cache [${key}]`);
-    } finally {
-      memoryQueue.remove(key);
     }
   }
 
--- a/packages/cloudflare/src/api/kv-cache.ts
+++ b/packages/cloudflare/src/api/kv-cache.ts
@@ -2,6 +2,7 @@ import type { CacheValue, IncrementalCache, WithLastModified } from "@opennextjs
 import { IgnorableError, RecoverableError } from "@opennextjs/aws/utils/error.js";
 
 import { getCloudflareContext } from "./cloudflare-context.js";
+import memoryQueue from "./memory-queue.js";
 
 export const CACHE_ASSET_DIR = "cdn-cgi/_next_cache";
 
@@ -116,6 +117,8 @@ class Cache implements IncrementalCache {
       );
     } catch {
       throw new RecoverableError(`Failed to set cache [${key}]`);
+    } finally {
+      memoryQueue.remove(key);
     }
   }
 
--- a/packages/cloudflare/src/api/memory-queue.spec.ts
+++ b/packages/cloudflare/src/api/memory-queue.spec.ts
@@ -1,7 +1,7 @@
 import { generateMessageGroupId } from "@opennextjs/aws/core/routing/queue.js";
-import { beforeAll, describe, expect, it, vi } from "vitest";
+import { afterEach, beforeAll, describe, expect, it, vi } from "vitest";
 
-import cache from "./memory-queue";
+import cache, { DEFAULT_REVALIDATION_TIMEOUT } from "./memory-queue";
 
 vi.mock("./.next/prerender-manifest.json", () => Promise.resolve({ preview: { previewModeId: "id" } }));
 
@@ -17,20 +17,36 @@ describe("MemoryQueue", () => {
     globalThis.internalFetch = vi.fn().mockReturnValue(new Promise((res) => setTimeout(() => res(true), 1)));
   });
 
-  it("should de-dupe revalidations", async () => {
-    const firstBatch = [cache.send(defaultOpts), cache.send(defaultOpts)];
-    vi.advanceTimersByTime(1);
-    await Promise.all(firstBatch);
+  afterEach(() => vi.clearAllMocks());
+
+  it("should process revalidations for a path", async () => {
+    const firstRequest = cache.send(defaultOpts);
+    vi.advanceTimersByTime(DEFAULT_REVALIDATION_TIMEOUT);
+    await firstRequest;
     expect(globalThis.internalFetch).toHaveBeenCalledTimes(1);
 
-    const secondBatch = [cache.send(defaultOpts)];
-    vi.advanceTimersByTime(10_000);
-    await Promise.all(secondBatch);
-    expect(globalThis.internalFetch).toHaveBeenCalledTimes(2);
-
-    const thirdBatch = [cache.send({ ...defaultOpts, MessageGroupId: generateMessageGroupId("/other") })];
+    const secondRequest = cache.send(defaultOpts);
     vi.advanceTimersByTime(1);
-    await Promise.all(thirdBatch);
-    expect(globalThis.internalFetch).toHaveBeenCalledTimes(3);
+    await secondRequest;
+    expect(globalThis.internalFetch).toHaveBeenCalledTimes(2);
+  });
+
+  it("should process revalidations for multiple paths", async () => {
+    const firstRequest = cache.send(defaultOpts);
+    vi.advanceTimersByTime(1);
+    await firstRequest;
+    expect(globalThis.internalFetch).toHaveBeenCalledTimes(1);
+
+    const secondRequest = cache.send({ ...defaultOpts, MessageGroupId: generateMessageGroupId("/other") });
+    vi.advanceTimersByTime(1);
+    await secondRequest;
+    expect(globalThis.internalFetch).toHaveBeenCalledTimes(2);
+  });
+
+  it("should de-dupe revalidations", async () => {
+    const requests = [cache.send(defaultOpts), cache.send(defaultOpts)];
+    vi.advanceTimersByTime(1);
+    await Promise.all(requests);
+    expect(globalThis.internalFetch).toHaveBeenCalledTimes(1);
   });
 });
--- a/packages/cloudflare/src/api/memory-queue.spec.ts
+++ b/packages/cloudflare/src/api/memory-queue.spec.ts
@@ -3,9 +3,7 @@ import { beforeAll, describe, expect, it, vi } from "vitest";
 
 import cache from "./memory-queue";
 
-vi.mock("./internal/manifest.js", () => ({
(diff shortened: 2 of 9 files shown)


packages/cloudflare/src/api/kv-cache.ts
```diff
@@ -116,6 +117,8 @@ class Cache implements IncrementalCache {
       );
     } catch {
       throw new RecoverableError(`Failed to set cache [${key}]`);
+    } finally {
```
vicb, line 120: looks weird to have a dependency here!

packages/cloudflare/src/api/memory-queue.ts
```diff
@@ -0,0 +1,52 @@
...
+  public async send({ MessageBody: { host, url }, MessageGroupId }: QueueMessage): Promise<void> {
+    this.revalidatedPaths.set(
+      MessageGroupId,
+      // force remove to allow new revalidations incase something went wrong
+      setTimeout(() => this.removeId(MessageGroupId), 10_000)
+    );
+
+    try {
+      // @ts-expect-error
+      const manifest = await import("./.next/prerender-manifest.json");
```
vicb, line 24: Can you add a TODO to drop this import.
See https://github.com/opennextjs/opennextjs-cloudflare/issues/361

vicb (commented):
I took a quick look but I'm not sure to understand how dedup works here.

Can you add  some tests?

Thanks

packages/cloudflare/package.json
```diff
@@ -73,7 +73,7 @@
   "dependencies": {
     "@ast-grep/napi": "^0.34.1",
     "@dotenvx/dotenvx": "catalog:",
-    "@opennextjs/aws": "https://pkg.pr.new/@opennextjs/aws@727",
+    "@opennextjs/aws": "https://pkg.pr.new/@opennextjs/aws@0b8705a",
```
vicb, line 76: FYI according to the description, this could be `@731` (each PR is released with the PR number, see the comment on the PR)

james-elicx: yeah sorry this was because i pushed different changes to the pr so i just used the commit instead of removing the old download from my pnpm cache. sorted now

packages/cloudflare/src/api/kv-cache.ts
james-elicx: from what i can tell, there isn't a way to know when a revalidation is complete from within the queue itself, so i put a call to it from the incremental cache instead.

two things come to mind:
1. remove the entry after the fetch call instead of after the incremental cache set
2. change the cache stuff in aws to invoke a new method on a queue after an incremental cache set

i can take a look at the second option. it sounds reasonable. just would be nice if we had a monorepo and i could do it in this pr.

james-elicx: done the second option using https://github.com/opennextjs/opennextjs-aws/pull/734

james-elicx: conico made a good point about chance of separate isolates handling requests... let me change to the first option.

packages/cloudflare/src/api/memory-queue.spec.ts
```diff
@@ -3,8 +3,8 @@ import { beforeAll, describe, expect, it, vi } from "vitest";
 
 import cache from "./memory-queue";
 
-vi.mock("./.next/prerender-manifest.json", () => ({
```
james-elicx: This mock would fail to resolve a value for the dynamic import intermittently in the tests, so I moved the function to its own file so that the tests can be stable.

image

james-elicx: it appears to be when two calls to the dynamic import happen at the exact same time that vitest fails to resolve them, so managed to fix this by changing the test so we're not calling it twice at the same time...

packages/cloudflare/src/api/memory-queue.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  public revalidatedPaths = new Map<string, ReturnType<typeof setTimeout>>();
+
+  public async send({ MessageBody: { host, url }, MessageGroupId }: QueueMessage): Promise<void> {
+    if (this.revalidatedPaths.has(MessageGroupId)) return;
+
+    this.revalidatedPaths.set(
+      MessageGroupId,
+      // force remove to allow new revalidations incase something went wrong
+      setTimeout(() => this.revalidatedPaths.delete(MessageGroupId), 10_000)
```
vicb, line 20: Should `10_000` be an exported constant (re-used in tests) or a config option?

packages/cloudflare/src/api/memory-queue.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+  it("should de-dupe revalidations", async () => {
+    const firstBatch = [cache.send(defaultOpts), cache.send(defaultOpts)];
+    vi.advanceTimersByTime(1);
+    await Promise.all(firstBatch);
+    expect(globalThis.internalFetch).toHaveBeenCalledTimes(1);
+
+    const secondBatch = [cache.send(defaultOpts)];
+    vi.advanceTimersByTime(10_000);
+    await Promise.all(secondBatch);
+    expect(globalThis.internalFetch).toHaveBeenCalledTimes(2);
```
vicb, line 29: Should this be split in multiple tests so that `.toHaveBeenCalledTimes()` doesn't accumulate?

packages/cloudflare/src/api/memory-queue.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+    vi.useFakeTimers();
+    globalThis.internalFetch = vi.fn().mockReturnValue(new Promise((res) => setTimeout(() => res(true), 1)));
+  });
+
+  afterEach(() => vi.clearAllMocks());
+
+  it("should process revalidations for a path", async () => {
+    const firstRequest = cache.send(defaultOpts);
+    vi.advanceTimersByTime(DEFAULT_REVALIDATION_TIMEOUT);
+    await firstRequest;
```
vicb, line 25: Wondering if that would make more sense?

```suggestion
    await cache.send(defaultOpts);
    vi.advanceTimersByTime(DEFAULT_REVALIDATION_TIMEOUT);
```

packages/cloudflare/src/api/memory-queue.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+describe("MemoryQueue", () => {
+  beforeAll(() => {
+    vi.useFakeTimers();
+    globalThis.internalFetch = vi.fn().mockReturnValue(new Promise((res) => setTimeout(() => res(true), 1)));
+  });
+
+  afterEach(() => vi.clearAllMocks());
+
+  it("should process revalidations for a path", async () => {
+    const firstRequest = cache.send(defaultOpts);
```
vicb, line 23: I would inline defaultOpts in all tests

See DRY vs DAMP, i.e. https://enterprisecraftsmanship.com/posts/dry-damp-unit-tests/

packages/cloudflare/src/api/memory-queue.ts
```diff
@@ -0,0 +1,51 @@
...
+ * The Memory Queue offers basic ISR revalidation by directly requesting a revalidation of a route.
+ *
+ * It offers basic support for in-memory de-duping per isolate.
+ */
+export class MemoryQueue implements Queue {
+  readonly name = "memory-queue";
+
+  public revalidatedPaths = new Map<string, ReturnType<typeof setTimeout>>();
+
+  constructor(private opts: Options = { revalidationTimeoutMs: DEFAULT_REVALIDATION_TIMEOUT }) {}
```
vicb, line 18: Couldn't the type be inferred here?

```suggestion
  constructor(private opts = { revalidationTimeoutMs: DEFAULT_REVALIDATION_TIMEOUT }) {}
```

packages/cloudflare/src/api/memory-queue.ts
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * The Memory Queue offers basic ISR revalidation by directly requesting a revalidation of a route.
+ *
+ * It offers basic support for in-memory de-duping per isolate.
+ */
+export class MemoryQueue implements Queue {
+  readonly name = "memory-queue";
+
+  public revalidatedPaths = new Map<string, ReturnType<typeof setTimeout>>();
```
vicb, line 16: Should `public` be `private` ?
If not you can drop `public` for consistency (with i.e. name)

packages/cloudflare/src/cli/build/utils/ensure-cf-config.ts
```diff
@@ -15,7 +15,9 @@ export function ensureCloudflareConfig(config: OpenNextConfig) {
       typeof config.default?.override?.incrementalCache === "function",
     dftUseDummyTagCache: config.default?.override?.tagCache === "dummy",
     dftMaybeUseQueue:
-      config.default?.override?.queue === "dummy" || config.default?.override?.queue === "direct",
+      config.default?.override?.queue === "dummy" ||
+      config.default?.override?.queue === "direct" ||
+      typeof config.default?.override?.incrementalCache === "function",
```
vicb, line 20:
```suggestion
      typeof config.default?.override?.queue === "function",
```

packages/cloudflare/src/api/memory-queue.ts
```diff
@@ -0,0 +1,51 @@
+import logger from "@opennextjs/aws/logger.js";
+import type { Queue, QueueMessage } from "@opennextjs/aws/types/overrides.js";
+
+type Options = { revalidationTimeoutMs: number };
+
+export const DEFAULT_REVALIDATION_TIMEOUT = 10_000;
```
vicb, line 6:
```suggestion
export const DEFAULT_REVALIDATION_TIMEOUT_MS = 10_000;
```

packages/cloudflare/src/api/memory-queue.ts
```diff
@@ -0,0 +1,51 @@
...
+    if (this.revalidatedPaths.has(MessageGroupId)) return;
+
+    this.revalidatedPaths.set(
+      MessageGroupId,
+      // force remove to allow new revalidations incase something went wrong
+      setTimeout(() => this.revalidatedPaths.delete(MessageGroupId), this.opts.revalidationTimeoutMs)
+    );
+
+    try {
+      const protocol = host.includes("localhost") ? "http" : "https";
```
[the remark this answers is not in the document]
vicb: I guess there are multiple instances of that already, but would be nice to think about the best way to handle that at some point

vicb (approved):
LGTM.

I added some comments but this looks good to go.

Thanks for your work on this and sorry for the review delay.

packages/cloudflare/src/api/memory-queue.spec.ts
james-elicx: this would cause the test to time out as it would be awaiting on the fetch, and neither the fetch or the settimeout would resolve because we haven't advanced the timers.

vicb (approved):

james-elicx:
Docs PR - https://github.com/opennextjs/docs/pull/74

PR #365, opened 2025-02-10
refactor(ContentUpdater): add a filterContent argument

vicb:
It is used as a fast heuristics to trigger the update

### DIFF
--- a/packages/cloudflare/src/cli/build/patches/plugins/require.ts
+++ b/packages/cloudflare/src/cli/build/patches/plugins/require.ts
@@ -1,44 +1,48 @@
 import type { ContentUpdater } from "./content-updater";
 
 export function fixRequire(updater: ContentUpdater) {
-  return updater.updateContent("fix-require", { filter: /\.(js|mjs|cjs|jsx|ts|tsx)$/ }, ({ contents }) => {
-    // `eval(...)` is not supported by workerd.
-    contents = contents.replaceAll(`eval("require")`, "require");
+  return updater.updateContent(
+    "fix-require",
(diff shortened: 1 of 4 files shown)


packages/cloudflare/src/cli/build/patches/plugins/require.ts
```diff
@@ -1,44 +1,48 @@
 import type { ContentUpdater } from "./content-updater";
 
 export function fixRequire(updater: ContentUpdater) {
-  return updater.updateContent("fix-require", { filter: /\.(js|mjs|cjs|jsx|ts|tsx)$/ }, ({ contents }) => {
-    // `eval(...)` is not supported by workerd.
-    contents = contents.replaceAll(`eval("require")`, "require");
+  return updater.updateContent(
+    "fix-require",
+    { filter: /\.(js|mjs|cjs|jsx|ts|tsx)$/, contentFilter: /.*/ },
```
conico974, line 6: Shouldn't `contentFilter` be something like `eval\("require"\)|require\(.@opentelemetry\/api.\)` here ?

conico974 (approved):
Just a little comment, otherwise LGTM

packages/cloudflare/src/cli/build/patches/plugins/require.ts
vicb: It wouldn't really have an impact. 
It's used mostly to prevent expensive computation like parsing code to ast.
This patch is based on string replace adding a filtering will have little effect (might even be worse?)

vicb:
Thanks for the review @conico974

PR #366, opened 2025-02-11
refactor: rename the api example to playground

vicb:
I feel like "api'" is no a good name any more as we keep adding more misc tests there.

I also have plan to add a few other pages to test cache features.


conico974 (approved):

vicb:
Thanks for the reviews!

Issue #331, opened 2025-02-03
Add e2e tests that test just the build process


dario-piotrowicz:
When applications are build using the adapter there are a few things that can happen in the terminal:
 - some prompts (e.g. for creating missing open-next and wrangler config files)
 - logs (e.g. workers saved at: <path>)
 - build errors occur (e.g. `getCloudflareContext` being used in an SSG'd route)

These things can't be tested in our playwright e2es since they simply rely on a preview command and test the application's functionalities, so ideally we should create a new set of "lower level" e2es that can test the build aspect of the adapter itself.

Additionally with this we can also test that files have been correctly saved onto the filesystem (e.g. `handler.mjs.meta.json`).

Issue #229, opened 2025-01-07
[FEATURE] improve `getCloudflareContext()`


vicb:
`getClouflareContext` retrieves the context from either a worker request (through ALS) or by starting a `getPlatformProxy` in `next dev`.

There are a few things we could improve:

- throw in a worker context when `getClouflareContext` is called outside of a request context (i.e. top level)
- remove the node specific code in the worker context (we could define `WORKER_CONTEXT = true` in esbuild to strip the code at build time).

See: 
- https://github.com/cloudflare/next-on-pages/blob/2cd4c3c704a00e6b693229f1f14102abc6318d11/packages/next-on-pages/src/api/getRequestContext.ts#L58
- https://github.com/cloudflare/next-on-pages/blob/2cd4c3c704a00e6b693229f1f14102abc6318d11/packages/next-on-pages/src/api/getRequestContext.ts#L62

/cc @dario-piotrowicz @james-elicx

PR #372, opened 2025-02-12
add "async mode" to `getCloudflareContext`

### LINKED ISSUE #371: getCloudflareContext async
That would be helpful for SSG

But might get local data (D1) at build time.

Needs to investigate

dario-piotrowicz:
add a new option to `getCloudflareContext` that makes it run in "async mode", the difference being that the returned value is a promise of the Cloudflare context instead of the context itself

The main of this is that it allows the function to also run during SSG (since the missing context can be created on demand).

___

Closes #371 

Docs PR: https://github.com/opennextjs/docs/pull/75


.changeset/tough-tables-talk.md
```diff
@@ -0,0 +1,10 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+add "async mode" to `getCloudflareContext`
+
+add a new option to `getCloudflareContext` that makes it run in "async mode", the difference being that the returned value is a
```
vicb, line 7:
```suggestion
Add an `async` option to `getCloudflareContext({async})` to run it in "async mode", the difference being that the returned value is a
```

examples/common/apps.ts
```diff
@@ -6,6 +6,7 @@ const apps = [
   "middleware",
   "vercel-blog-starter",
   "vercel-commerce",
+  "ssg-app",
```
vicb, line 9: I renamed api to playground (and now I realize I missed l.4), maybe you can re-use that playground to add your stuff instead of creating yer another app?

packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -45,6 +45,14 @@ type InternalGlobalThis<
   __NEXT_DATA__: Record<string, unknown>;
 };
 
+type GetCloudflareContextOptions = Partial<{
```
vicb, line 48: Do we need the Partial here?

packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -53,26 +61,46 @@ type InternalGlobalThis<
...
-    // can do about this, so the only solution is to error asking the developer to opt-out of SSG
-    // Next.js sets globalThis.__NEXT_DATA__.nextExport to true for the worker, so we can use that to detect
-    // that the route is being SSG'd (source: https://github.com/vercel/next.js/blob/4e394608423/packages/next/src/export/worker.ts#L55-L57)
-    if (global.__NEXT_DATA__?.nextExport === true) {
+    // The non-async mode of `getCloudflareContext`, relies on the context set on the global state
+    // by either the worker entrypoint (in prod) or by `initOpenNextCloudflareForDev` (in dev), neither
+    // can work during SSG since for SSG Next.js creates (jest) workers that don't get access to the
+    // normal global state. There isn't much we can do about this so we can only throw with a helpful
+    // error message for the user.
+    // Note: Next.js sets globalThis.__NEXT_DATA__.nextExport to true for the worker, so we can use that to detect
```
vicb, line 88: "for the worker" sounds confusing here.

You could either expand or shorten: "Next.js sets globalThis.__NEXT_DATA__.nextExport to true for SSG routes"

packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -53,26 +61,46 @@ type InternalGlobalThis<
...
+    // wether the route is being SSG'd (source: https://github.com/vercel/next.js/blob/4e394608423/packages/next/src/export/worker.ts#L55-L57)
+    if (!asyncMode && global.__NEXT_DATA__?.nextExport === true) {
       throw new Error(
         `\n\nERROR: \`getCloudflareContext\` has been called in a static route` +
-          ` that is not allowed, please either avoid calling \`getCloudflareContext\`` +
-          ` in the route or make the route non static (for example by exporting the` +
-          ` \`dynamic\` route segment config set to \`'force-dynamic'\`.\n`
+          ` that is not allowed, this can be solved in different ways:\n\n` +
+          ` - call \`getCloudflareContext\` in \`async\` mode\n` +
```
vicb, line 94:
```suggestion
          ` - call \`getCloudflareContext({async: true})\` to use the \`async\` mode\n` +
```

packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -53,26 +61,46 @@ type InternalGlobalThis<
 export function getCloudflareContext<
   CfProperties extends Record<string, unknown> = IncomingRequestCfProperties,
   Context = ExecutionContext,
->(): CloudflareContext<CfProperties, Context> {
+>(options: { async: true }): Promise<CloudflareContext<CfProperties, Context>>;
+export function getCloudflareContext<
+  CfProperties extends Record<string, unknown> = IncomingRequestCfProperties,
+  Context = ExecutionContext,
+>(options?: { async?: false }): CloudflareContext<CfProperties, Context>;
+export function getCloudflareContext<
```
vicb, line 69: I'm not really sure to like `getCloudflareContext({async:true})` over `getClouflfareContextAsync()` - I need to think more

examples/common/apps.ts
dario-piotrowicz: playground is not fully SSG:
Screenshot 2025-02-12 at 20 42 29

I intentionally added a new app that is fully SSG so that we can make sure that such an application can also be successfully run (if I remember correctly there were edge cases in next-on-pages in which an app with only SSG routes could not run)

I do think that this is a valuable addition, but if you strongly disagree I can try to reuse playground

packages/cloudflare/src/api/cloudflare-context.ts
dario-piotrowicz: No we don't, I was just future proofing this object in case in the future there might be more options, I can totally remove the `Partial` if you prefer (it does make little sense when there is only a single option)

examples/common/apps.ts
vicb: > playground is not fully SSG:

I think the important thing is that the layout/page you add are SSG?
Does whatever other existing routes are have any impact?

vicb: > if I remember correctly there were edge cases in next-on-pages in which an app with only SSG routes could not run

Oh maybe I should have read your comment fully before answering :)

packages/cloudflare/src/api/cloudflare-context.ts
vicb: Yeah, we can add it later if needed

examples/common/apps.ts
dario-piotrowicz: > Does whatever other existing routes are have any impact?

yeah probably not 🙂

packages/cloudflare/src/api/cloudflare-context.ts
dario-piotrowicz: Yes I agree, I might also prefer to have a distinct function for the async version (it would certainly simplify the function's type) I went with this version since that's what was discussed and agreed on in our weekly meeting

I'm happy to have two separate functions if you prefer that, it does feel clearer to me

although there are also some downsides:
 - more overhead to developers as they need to be aware of two functions
 - more overhead on us to document both functions
 
 up to you, I am fine with either 👍

examples/common/apps.ts
dario-piotrowicz: anyways having an only-SSG app might be not needed, personally I think that it is a worthwhile addition that doesn't really hurt, but again if you strongly disagree I can just reuse the playground one

packages/cloudflare/src/api/cloudflare-context.ts
dario-piotrowicz: removed 👍

examples/common/apps.ts
vicb: No pb if you feel like it's a worthwhile addition.

packages/cloudflare/src/api/cloudflare-context.ts
vicb: I'm not sure, it's more a gut feeling but changing the return type according to the what parameter is passed does not look super clean to me.

> more overhead on us to document both functions

IDK, I feel like documenting `getCloudflareContext()` and `getCloudflareContext({async: true})` or `getCloudflareContext()` and `getCloudflareContextAsync()` is not really different?

dario-piotrowicz: > I'm not sure, it's more a gut feeling but changing the return type according to the what parameter is passed does not look super clean to me.

yes, I agree

> IDK, I feel like documenting getCloudflareContext() and getCloudflareContext({async: true}) or getCloudflareContext() and getCloudflareContextAsync() is not really different?

the way I see it is that if you only have a single function you only need to document that and can as part of that mention that it has an "async mode", if you have two functions it probably can get a bit more verbose (not a huge deal either way)

examples/ssg-app/app/page.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export default async function Home() {
+  const cloudflareContext = await getCloudflareContext({
+    async: true,
+  });
+
+  return (
+    <div className={styles.page}>
+      <main className={styles.main}>
+        <h1>Hello from a Statically generated app</h1>
```
vicb, line 12:
```suggestion
        <h1>Hello from a Statically generated page</h1>
```

packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -53,26 +61,46 @@ type InternalGlobalThis<
...
+  const asyncMode = options.async;
+
   if (!cloudflareContext) {
-    // For SSG Next.js creates (jest) workers that run in parallel, those don't get the current global
-    // state so they can't get access to the cloudflare context, unfortunately there isn't anything we
-    // can do about this, so the only solution is to error asking the developer to opt-out of SSG
-    // Next.js sets globalThis.__NEXT_DATA__.nextExport to true for the worker, so we can use that to detect
-    // that the route is being SSG'd (source: https://github.com/vercel/next.js/blob/4e394608423/packages/next/src/export/worker.ts#L55-L57)
-    if (global.__NEXT_DATA__?.nextExport === true) {
+    // The non-async mode of `getCloudflareContext`, relies on the context set on the global state
```
vicb, line 83:
```suggestion
    // The sync mode of `getCloudflareContext`, relies on the context set on the global state
```

packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -53,26 +61,46 @@ type InternalGlobalThis<
...
+    // by either the worker entrypoint (in prod) or by `initOpenNextCloudflareForDev` (in dev), neither
+    // can work during SSG since for SSG Next.js creates (jest) workers that don't get access to the
+    // normal global state. There isn't much we can do about this so we can only throw with a helpful
+    // error message for the user.
+    // Note: Next.js sets globalThis.__NEXT_DATA__.nextExport to true for SSG routes, so we can use that to detect
+    // wether the route is being SSG'd (source: https://github.com/vercel/next.js/blob/4e394608423/packages/next/src/export/worker.ts#L55-L57)
```
vicb, line 89:
```suggestion
    // Note: Next.js sets globalThis.__NEXT_DATA__.nextExport to true for SSG routes
    // source: https://github.com/vercel/next.js/blob/4e394608423/packages/next/src/export/worker.ts#L55-L57)
```

packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -192,3 +220,19 @@ async function getCloudflareContextFromWrangler<
...
+ * globalThis so that it can be accessed later.
+ *
+ * @returns the local proxy version of the cloudflare context
+ */
+async function getAndSetCloudflareContextInNextDev<
+  CfProperties extends Record<string, unknown> = IncomingRequestCfProperties,
+  Context = ExecutionContext,
+>(): Promise<CloudflareContext<CfProperties, Context>> {
+  const context = await getCloudflareContextFromWrangler();
+  addCloudflareContextToNodejsGlobal(context);
```
vicb, line 236: This is not what the comment says:

```
when
 * running in the standard next dev server (via `next dev`), is also sets this value on the
 * globalThis so that it can be accessed later
 ```

So 
- maybe `addCloudflareContextToNodejsGlobal` should be `addCloudflareContextToGlobalThis`
- Do we want to do that in SSG or not?

File: examples/ssg-app/wrangler.json [hunk 11]
```diff
@@ -0,0 +1,22 @@
+{
```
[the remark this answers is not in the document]
dario-piotrowicz: what do you mean? why?

packages/cloudflare/src/api/cloudflare-context.ts
dario-piotrowicz: > This is not what the comment says:

You're right! thank you for noticing the outdated/invalid comment 🙏 

I've updated it: https://github.com/opennextjs/opennextjs-cloudflare/pull/372/commits/3c2df2723825985addff4111fdf0511180993b74

> maybe addCloudflareContextToNodejsGlobal should be addCloudflareContextToGlobalThis

I've been adding the `Nodejs` word in the functions to make it clearer that these only effect the globalThis of the nodejs process, technically this is incorrect, they effect the current globalThis, which in our case just so happen to always be node.js (the edge runtime globalThis is not effected by these (that [is patched differently](https://github.com/opennextjs/opennextjs-cloudflare/blob/76bc17da001705508730e6b6e20c75e1f15b2ba6/packages/cloudflare/src/api/cloudflare-context.ts#L150))).

I am happy keeping the `Nodejs` word if it makes things any clearer, otherwise if it doesn't, I'm happy to remove it, either way is fine, up to you I'm equally fine with either

> Do we want to do that in SSG or not?

Yes we want to do this for SSG (it's not strictly necessary but we don't want to call `getPlatformProxy` more times than its needed even in SSG)

packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -53,26 +61,45 @@ type InternalGlobalThis<
...
-          ` \`dynamic\` route segment config set to \`'force-dynamic'\`.\n`
+          ` that is not allowed, this can be solved in different ways:\n\n` +
+          ` - call \`getCloudflareContext({async: true})\` to use the \`async\` mode\n` +
+          ` - avoid calling \`getCloudflareContext\` in the route\n` +
+          ` - make the route non static\n`
       );
     }
 
+    if (options.async) {
+      return getAndSetCloudflareContextInNodejs();
```
vicb, line 100: Do we really need a separate function here, vs

```
return getCloudflareContextFromWrangler().then(context => {
  addCloudflareContextToNodejsGlobal(context);
  return context;
  });
```

vicb (approved):

packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -53,26 +61,48 @@ type InternalGlobalThis<
...
-          ` in the route or make the route non static (for example by exporting the` +
-          ` \`dynamic\` route segment config set to \`'force-dynamic'\`.\n`
+          ` that is not allowed, this can be solved in different ways:\n\n` +
+          ` - call \`getCloudflareContext({async: true})\` to use the \`async\` mode\n` +
+          ` - avoid calling \`getCloudflareContext\` in the route\n` +
+          ` - make the route non static\n`
       );
     }
 
+    if (options.async) {
```
vicb, line 99: Maybe 
- you can move this if as the first nested if and you can drop !asyncMode l. 89.
- be consistent in using options.async or asyncMode
- do a release once this gets merged

dario-piotrowicz: > be consistent in using options.async or asyncMode

oops, not sure how I missed using `asyncMode` here when I refactored this, my bad, thanks for noticing it 👍

dario-piotrowicz:
@vicb sorry I had to update the logic in `getCloudflareContext` as I realized that in some cases users could get the wrong behavior (a wrong error being thrown), sorry for the inconvenience please have another look 🙏 

___

PS: ideally all the different error cases of this (and their error messages) should be tested, for that we would need an extra app that doesn't set `initOpenNextCloudflareForDev`, also the build SSG aspects can't be properly tested anyways: https://github.com/opennextjs/opennextjs-cloudflare/issues/331. So it doesn't feel worth it to me (at least not right now, maybe after 331), especially since if things would go wrong at worst users could simply get the wrong error message.
I'm happy to add some tests here or in a followup if you disagree and thing that they'd be worth it.

packages/cloudflare/src/api/cloudflare-context.ts
vicb: No strong pref, pick what your prefer

examples/ssg-app/e2e/playwright.config.ts
```diff
@@ -0,0 +1,3 @@
+import { configurePlaywright } from "../../common/config-e2e";
+
+export default configurePlaywright("ssg-app", { isCI: !!process.env.CI });
```
vicb, line 3: I don't think testing multiple browsers is often needed.
I mainly kept it when refactoring to have concurrent requests - that's how I figured out the waitUntil bug.
Maybe the default should be a single browser and we only enable multiple browsers for a few apps?

```suggestion
export default configurePlaywright("ssg-app", { isCI: !!process.env.CI, multipleBrowsers: false });
```

examples/ssg-app/wrangler.json
```diff
@@ -0,0 +1,22 @@
...
+  "vars": {
+    "APP_VERSION": "1.2.345"
+  },
+  "kv_namespaces": [
+    // Create a KV binding with the binding name "NEXT_CACHE_WORKERS_KV"
+    // to enable the KV based caching:
+    // {
+    //   "binding": "NEXT_CACHE_WORKERS_KV",
+    //   "id": "<BINDING_ID>"
+    // }
```
vicb, line 20:
```suggestion
```

examples/ssg-app/wrangler.json
```diff
@@ -0,0 +1,22 @@
...
+  ]
```
vicb, line 21:
```suggestion
```

vicb (approved):
LGTM

packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -53,44 +61,71 @@ type InternalGlobalThis<
...
+    return asyncMode ? Promise.resolve(cloudflareContext) : cloudflareContext;
+  }
+
+  const inNodejsRuntime = process.env.NEXT_RUNTIME === "nodejs";
+
+  // Note: Next.js sets globalThis.__NEXT_DATA__.nextExport to true for SSG routes
+  // source: https://github.com/vercel/next.js/blob/4e394608423/packages/next/src/export/worker.ts#L55-L57)
+  const inSSG = global.__NEXT_DATA__?.nextExport === true;
+
+  if ((inNodejsRuntime || inSSG) && asyncMode) {
```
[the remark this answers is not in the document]
vicb: I find the logic confusing...

Could something like

```ts
if (cloudflareContext) { ... }

if (isAsync) {
  if (isEdgeRuntime) {
    throw "should be sync in edge"
  }
  return getContextAsync(); // local function
}

// sync

if (inSSG) {
  throw "should be async in SSG"
}

throw "should call init";
```

I needed to write that down to (hopefully) unconfuse me :)

It looks similar to what you have but shouldn't you have `throw "should be sync in edge"`?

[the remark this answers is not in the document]
vicb: ^still not very happy with that,

next step is:

```ts
if (cloudflareContext) { 
  return it or Promise<it>;
}

if (isSync) {
 getContextSync(...); //local fn
}
 
return getContextAsync(...); // local fn
```

next step is:

```ts
export function getCloudflareContext(...) {

// ...

}

export function getCloudflareContextAsync(...) {

// ...

}
```

Not sure if I've been helpful ;)

examples/ssg-app/wrangler.json
dario-piotrowicz: this feels pretty out of scope here, can I do it in a follow up PR?

packages/cloudflare/src/api/cloudflare-context.ts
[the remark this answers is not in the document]
dario-piotrowicz: > I'm not sure about `(inNodejsRuntime || inSSG)`
> 
> * it is always [true](https://github.com/opennextjs/opennextjs-cloudflare/blob/76bc17da001705508730e6b6e20c75e1f15b2ba6/packages/cloudflare/src/cli/build/bundle-server.ts#L122) on workers
> * the only case where I think this can be false is the edge runtime
> 
> If that's true, could the test be simplified as `!isEdgeRuntime && asyncMode`

This code is targeting `next dev` (where `process.env.NEXT_RUNTIME === "nodejs"` is not always `true`), in a built worker the context should always be available, so `getCloudflareContext` should always return here: https://github.com/opennextjs/opennextjs-cloudflare/blob/3cc360b7795ec4069c2b6081f0f0efe8f8ba609d/packages/cloudflare/src/api/cloudflare-context.ts#L82-L84 (unless maybe when the function is called in the top level? something likely to address as part of https://github.com/opennextjs/opennextjs-cloudflare/issues/229)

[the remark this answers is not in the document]
dario-piotrowicz: > It looks similar to what you have but shouldn't you have throw "should be sync in edge"?

In the edge runtime the async version works just fine, are you saying that we should arbitrarily prevent developers from using the async version inside the edge runtime (since it's not necessary there)? that feels arbitrary and unnecessary to me (is there any harm in allowing developers to use the async version there if they want to?)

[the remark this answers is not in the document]
dario-piotrowicz: I've added an if-else statement splitting the branches based on `asyncMode` I hope that makes it clearer please let me know

also please keep in mind that the error thrown at the end is applicable for both sync and async mode

if we keep the two modes in a single function splitting them feels to me to unnecessarily add more noise and potential code duplication down the line (and new option could potentially have to be passed to each function and handled there etc..)

packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -53,44 +61,71 @@ type InternalGlobalThis<
...
-    // during local development (`next dev`) it might be missing only if the developers hasn't called
-    // the `initOpenNextCloudflareForDev` function in their Next.js config file
+  // whether `getCloudflareContext` is run in "async mode"
+  const asyncMode = options.async;
+
+  if (cloudflareContext) {
+    return asyncMode ? Promise.resolve(cloudflareContext) : cloudflareContext;
+  }
+
+  const inNodejsRuntime = process.env.NEXT_RUNTIME === "nodejs";
```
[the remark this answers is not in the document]
dario-piotrowicz: added 👍

examples/ssg-app/e2e/playwright.config.ts
vicb: The suggested change should be done here.

> Maybe the default should be a single browser and we only enable multiple browsers for a few apps?

That should be discussed en done in a different PR if we decide it's a good thing

packages/cloudflare/src/api/cloudflare-context.ts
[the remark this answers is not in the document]
vicb: > In the edge runtime the async version works just fine

Wasn't that a reason why we introduce sync? (That is wrangler can not be called because it relies on Node API that are not available in the edge version)

examples/ssg-app/wrangler.json
vicb: why out of scope, this is not needed?

vicb: Oh actually that one should be close for the one below

packages/cloudflare/src/api/cloudflare-context.ts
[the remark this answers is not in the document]
dario-piotrowicz: > Wasn't that a reason why we introduce sync?

Not really, introducing sync wasn't a reason to make the change, it was an extra side effect/benefit that we got from making the switch, we could have kept the function async if we wanted to (but you argued that that's an API that developers are having issues with)

> That is wrangler can not be called because it relies on Node API that are not available in the edge version

Yes, in the edge runtime we can't import and call wrangler directly, but that is not really relevant

if you have called `initOpenNextCloudflareForDev` then async mode will just find the patched context and return it here: https://github.com/opennextjs/opennextjs-cloudflare/blob/30683e7c8462f3cfd17330cace512da582048fb6/packages/cloudflare/src/api/cloudflare-context.ts#L83

and If you haven't called `initOpenNextCloudflareForDev` the function will simply eventually throw this error: https://github.com/opennextjs/opennextjs-cloudflare/blob/30683e7c8462f3cfd17330cace512da582048fb6/packages/cloudflare/src/api/cloudflare-context.ts#L118-L132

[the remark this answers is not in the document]
vicb: Oh I see so `async` will retrieve the value set by initBlabla in the edge runtime.

Maybe this should be explained in the code here (maybe we can drop the ref in Next?)

```ts
  // Note: Next.js sets process.env.NEXT_RUNTIME to 'nodejs' when the runtime in use is the node.js one
  // source: https://github.com/vercel/next.js/blob/1e8603113/packages/next/src/build/webpack/plugins/define-env-plugin.ts#L179-L183
  const inNodejsRuntime = process.env.NEXT_RUNTIME === "nodejs";
  ```
  
I think I finally got the logic now... and I'm pretty I will not get in when I look back at the code in a few weeks.

So maybe splitting that function in local `getContextSync` and `getContextAsync` could help make the code clearer?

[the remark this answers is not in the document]
Comment (dario-piotrowicz) [reply packages/cloudflare/src/api/cloudflare-context.ts]: > maybe we can drop the ref in Next?

I don't follow, what are you suggesting?

> So maybe splitting that function in local getContextSync and getContextAsync could help make the code clearer?

I personally don't think so, as I already commented above:
> if we keep the two modes in a single function splitting them feels to me to unnecessarily add more noise and potential code duplication down the line (and new option could potentially have to be passed to each function and handled there etc..)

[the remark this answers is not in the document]
Comment (vicb) [reply packages/cloudflare/src/api/cloudflare-context.ts]: > I don't follow, what are you suggesting?

```
  // Note: Next.js sets process.env.NEXT_RUNTIME to 'nodejs' when the runtime in use is the node.js one
  // HERE: explain why we need that
  const inNodejsRuntime = process.env.NEXT_RUNTIME === "nodejs";
  ```
  
  > I personally don't think so
  
My feedback is that the code is hard to follow.

vicb (approved):
Maybe:

```ts
const InitOpenNextCloudflareForDevErrorMsg =`...`;
...
throw new Error(InitOpenNextCloudflareForDevErrorMsg);
```

Issue #321, opened 2025-02-02
[BUG] handler.mjs: Comparison with -0 using the "===" operator will also match 0 [equals-negative-zero]


LubomirGeorgiev:
### Describe the bug

When I build the app I get this warning:
```
▲ [WARNING] Comparison with -0 using the "===" operator will also match 0 [equals-negative-zero]

    .open-next/server-functions/default/handler.mjs:95787:47:
      95787 │                                 else if (t === -0) {
```

### Steps to reproduce

Go to https://github.com/LubomirGeorgiev/cloudflare-workers-nextjs-saas-template/actions/runs/13103167050/job/36553955672#step:8:137

### Expected behavior

Build without warnings

### @opennextjs/cloudflare version

0.4.3

### Wrangler version

3.107.2

### next info output

```shell
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:24 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T6030
  Available memory (MB): 36864
  Available CPU cores: 11
Binaries:
  Node: 22.6.0
  npm: 10.8.2
  Yarn: N/A
  pnpm: 9.12.1
Relevant Packages:
  next: 14.2.23 // An outdated version detected (latest is 15.1.6), upgrade is highly recommended!
  eslint-config-next: 14.2.23
  react: 18.3.1
  react-dom: 18.3.1
  typescript: 5.7.3
Next.js Config:
  output: N/A
 ⚠ An outdated version detected (latest is 15.1.6), upgrade is highly recommended!
   Please try the latest canary version (`npm install next@canary`) to confirm the issue still exists before creating a new issue.
   Read more - https://nextjs.org/docs/messages/opening-an-issue
```

vicb:
The faulty code is in `next@14.2.11_@opentelemetry+api@1.9.0_@playwright+test@1.47.0_react-dom@18.3.1_react@18.3.1__react@18.3.1/node_modules/next/dist/compiled/node-fetch/index.js`.

While we could silence it, it doesn't sound right.

The better fix would be to investigate and submit a fix upstream

nrjrajpal:
Can someone please tell me how this issue can be resolved? I am new to Cloudflare, and I am unable to build my application because of this. I really appreciate any help you can provide.

vicb:
[...]

PR #394, opened 2025-02-18
fix: improve windows support

vicb:
The PR title/changeset is a bit vague because I didn't actually tested on windows.

But I believe this PR will improve, i.e.:
- https://github.com/cloudflare/workers-sdk/issues/8159
- https://github.com/opennextjs/opennextjs-cloudflare/issues/321#issuecomment-2646200860

What this PR does is to normalize the path variables vs only the constants those are compared to.


dario-piotrowicz (approved):

vicb:
Thanks Dario!

PR #397, opened 2025-02-19
feat: custom worker entrypoint

imconnorngl:
We've recently attempted to adopt OpenNext to run some of our applications but have struggled to do so due to it lacking a feature which exists on [Next On Pages](https://github.com/cloudflare/next-on-pages).

This PR attempts to implement this through the use of a option which allows the user to provide an entrypoint to use in favour of the templated entrypoint (`opennextjs-cloudflare --customWorkerEntrypoint=./entrypoint.ts`).

We're currently using this through a fork for our application.


vicb:
> We've recently attempted to adopt OpenNext to run some of our applications but have struggled to do so due to it lacking a feature which exists on [Next On Pages](https://github.com/cloudflare/next-on-pages).
> 
> This PR attempts to implement this through the use of a option which allows the user to provide an entrypoint to use in favour of the templated entrypoint (`opennextjs-cloudflare --customWorkerEntrypoint=./entrypoint.ts`).
> 
> We're currently using this through a fork for our application.

Thanks for the PR!

It's probably a bit early in the lifecycle of the project to consider merging this now.

Few comments:
- I'm not sure if ON should compile user code (vs wrangler) - maybe it should only expose a documented request handler
- We probably don't want to add more CLI options but rather remove them as we plan to merge with the aws adapter - additional options should go to the open next config file.

Maybe the simplest for now is to start from the generated `.open-next/worker.js` (or source [`worker.ts`](https://github.com/opennextjs/opennextjs-cloudflare/blob/main/packages/cloudflare/src/cli/templates/worker.ts)), copy edit the file somewhere in your project - you will need to adjust the relative paths. Then you can point your wrangler config (`wrangler.json') to that file.

You should think about checking the diffs in `worker.js/ts` when updating to a new version. Things will probably change there before 1.0.

I hope this helps

vicb:
See https://github.com/opennextjs/opennextjs-cloudflare/pull/591

PR #400, opened 2025-02-20
refactor: drop `patchExceptionBubbling`

vicb:
This does not seem to be required.


dario-piotrowicz (approved):

vicb:
Thanks Dario - as discussed offline, we'll re-open with a proper test if this is ever an issue.

PR #401, opened 2025-02-20
refactor: patch getBuildId using ast-grep

vicb:


packages/cloudflare/src/cli/build/patches/plugins/build-id.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * Inline `getBuildId` as it relies on `readFileSync`that is not supported by workerd.
```
dario-piotrowicz, line 2:
```suggestion
 * Inline `getBuildId` as it relies on `readFileSync` that is not supported by workerd.
```

packages/cloudflare/src/cli/build/patches/plugins/build-id.ts
```diff
@@ -0,0 +1,31 @@
...
+    "inline-build-id",
+    {
+      filter: getCrossPlatformPathRegex(String.raw`/next/dist/server/next-server\.js$`, { escape: false }),
+      contentFilter: /getBuildId\(/,
+    },
+    async ({ contents }) => patchCode(contents, rule)
+  );
+}
+
+const rule = `
```
dario-piotrowicz, line 21: should this rule be tested?

dario-piotrowicz (approved):

packages/cloudflare/src/cli/build/patches/plugins/build-id.ts
vicb: fixed (and the source as it was a copy/paste)

packages/cloudflare/src/cli/build/patches/plugins/build-id.ts
vicb: good point, added a commit with a test

vicb:
Thanks for the review Dario, PTAL

PR #404, opened 2025-02-21
fix incorrect (sync) `getCloudflareContext` error message

### LINKED ISSUE #382: [Feature] can `getCloudflareContext({async: false})` work at top level with `next dev`
If you update `examples/playground/api/env` as follow:

```ts
import { getCloudflareContext } from "@opennextjs/cloudflare";

// the output.
export const dynamic = "force-dynamic";

console.log(getCloudflareContext());

export async function GET() {
  return new Response(JSON.stringify(process.env));
}
```

Note: calling `getCloudflareContext` is not actually supported at top level but I'm working on adding support for that.

It would log an incorrect message (the route is dynamic):

```
ERROR: `getCloudflareContext` has been called in a static route, that is not allowed, this can be solved in different ways:

 - call `getCloudflareContext({async: true})` to use the `async` mode
 - avoid calling `getCloudflareContext` in the route
 - make the route non static
```

- This is not a static route
- `getCloudflareContext({async: true})` seem to work

We should first fix the error message.

It would also be nice that we could make `{async: false}` work here if possible

dario-piotrowicz:
currently `getCloudflareContext` run in sync mode at the top level of a not static route
gives a misleading error message saying that the function needs to be run in a not static
route, the changes here correct this error message clarifying that the problem actually is

fixes  #382

### DIFF
--- a/.changeset/warm-lizards-lick.md
+++ b/.changeset/warm-lizards-lick.md
@@ -1,0 +1,9 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+fix incorrect (sync) `getCloudflareContext` error message
+
+currently `getCloudflareContext` run in sync mode at the top level of a not static route
+gives a misleading error message saying that the function needs to be run in a not static
+route, the changes here correct this error message clarifying that the problem actually is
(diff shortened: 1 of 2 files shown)


dario-piotrowicz:
PS: as mentioned in this comment: https://github.com/opennextjs/opennextjs-cloudflare/issues/382#issuecomment-2674999814, unfortunately I think that updating the error message to be more generic is the best we can do here 😓 (since we can't make the sync version of `getCloudflareContext` work at the top level and we can't detect whether a route is being server side rendered or not)

vicb (approved):

PR #409, opened 2025-02-24
make sure that instrumentation files work

### LINKED ISSUE #171: Investigate and support Next.js instrumentation
Next.js 15 introduced a `loadInstrumentationModule` methods that appears in our bundled worker which caused our workers to fail at runtime.

We're solved such issue by replacing the content of the `loadInstrumentationModule` methods with a no-op implementation (see: https://github.com/opennextjs/opennextjs-cloudflare/pull/169), this was the simplest and quickest way to work around the issue and unblock Next.js 15 for us, it is however most likely not the correct solution and almost certainly breaks [Next.js' instrumentation](https://nextjs.org/docs/app/building-your-application/optimizing/instrumentation).
 
So this issue is for investigating and understanding if and how we support instrumentation, and if we don't support it (which most likely we don't) find a way to do so.

dario-piotrowicz:
fixes #171


examples/common/apps.ts
```diff
@@ -7,6 +7,7 @@ const apps = [
   "vercel-blog-starter",
   "vercel-commerce",
   "ssg-app",
+  "instrumentation-app",
```
vicb, line 10: Is there a particular reason to add an app instead of re-using i.e. the playground?

examples/middleware/tsconfig.json
```diff
@@ -21,5 +21,5 @@
     ]
   },
   "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", ".next/types/**/*.ts"],
-  "exclude": ["node_modules", "open-next.config.ts", "worker.ts"]
+  "exclude": ["node_modules"]
```
vicb, line 24: question: why is this file and its permission updated?

packages/cloudflare/src/cli/build/patches/plugins/load-instrumentation.ts
```diff
@@ -2,23 +2,46 @@
...
+export function patchLoadInstrumentation(updater: ContentUpdater, buildOpts: BuildOptions) {
   return updater.updateContent(
     "patch-load-instrumentation",
     { filter: /\.(js|mjs|cjs|jsx|ts|tsx)$/, contentFilter: /async loadInstrumentationModule\(/ },
-    ({ contents }) => patchCode(contents, instrumentationRule)
+    async ({ contents }) => patchCode(contents, await getRule(buildOpts))
   );
 }
+
+export async function getRule(buildOpts: BuildOptions) {
```
vicb, line 21: Inject `instrumentationFilePath: string | null` to simplify the testing

vicb (changes requested):
Thanks!

I have added a few comments

examples/common/apps.ts
vicb: Oh probably because this is in Next15...

Does instrumentation work on Next14?

Maybe we should have a playground14 and a playground15?

examples/middleware/tsconfig.json
dario-piotrowicz: I made this change by mistake, I was sure I reverted it but I forgot to push the change up, done, thanks for noticing it 👍

dario-piotrowicz:
converted to draft as I just noticed that the patch does not work on Next 14

examples/common/apps.ts
dario-piotrowicz: done

examples/common/apps.ts
```diff
@@ -3,7 +3,8 @@ const apps = [
   // examples
   "create-next-app",
   "middleware",
-  "playground",
+  "playground14",
+  "playground15",
```
vicb, line 7: Thanks 🙏

examples/middleware/tsconfig.json
vicb, line 1: Can you revert the permission change?

examples/playground14/app/api/instrumentation/route.ts
```diff
@@ -0,0 +1,10 @@
+import { NextResponse } from "next/server";
+
+export const dynamic = "force-dynamic";
+
+export function GET() {
+  return NextResponse.json({
+    "nodejs-instrumentation-setup": globalThis["__NODEJS_INSTRUMENTATION_SETUP"] ?? "undefined",
```
vicb, line 7: is `?? "undefined"` really needed?

examples/playground15/README.md
vicb, line 1: Remove this file?

packages/cloudflare/src/cli/build/patches/plugins/load-instrumentation.spec.ts
```diff
@@ -1,48 +1,61 @@
-import { describe, expect, test } from "vitest";
+import { describe, expect, test, vi } from "vitest";
 
 import { patchCode } from "../ast/util.js";
-import { instrumentationRule } from "./load-instrumentation.js";
+import { getRule } from "./load-instrumentation.js";
+
+vi.mock(import("node:fs"), async (importOriginal) => {
```
vicb, line 6: remove?

packages/cloudflare/src/cli/build/patches/plugins/load-instrumentation.ts
```diff
@@ -2,23 +2,34 @@
...
-export function patchLoadInstrumentation(updater: ContentUpdater) {
   return updater.updateContent(
     "patch-load-instrumentation",
     { filter: /\.(js|mjs|cjs|jsx|ts|tsx)$/, contentFilter: /async loadInstrumentationModule\(/ },
-    ({ contents }) => patchCode(contents, instrumentationRule)
+    async ({ contents }) => patchCode(contents, await getRule(builtInstrumentationPath))
   );
 }
+
+export async function getRule(builtInstrumentationPath: string | null) {
```
vicb, line 21: remove async + corresponding awaits (also in test)

examples/playground14/app/api/instrumentation/route.ts
dario-piotrowicz: no, but without it you would not get the field at all, for making this more self-explanatory I find it nice to clearly include "undefined" instead, but if you prefer the other version I'll switch to it

packages/cloudflare/src/cli/build/patches/plugins/instrumentation.ts
```diff
@@ -0,0 +1,84 @@
...
+      "patch-load-instrumentation",
+      { filter: /\.(js|mjs|cjs|jsx|ts|tsx)$/, contentFilter: /async loadInstrumentationModule\(/ },
+      async ({ contents }) => patchCode(contents, getNext15Rule(builtInstrumentationPath))
+    ),
+    updater.updateContent(
+      "patch-prepareImpl",
+      { filter: /\.(js|mjs|cjs|jsx|ts|tsx)$/, contentFilter: /async prepareImpl\(/ },
+      async ({ contents }) => patchCode(contents, getNext14Rule(builtInstrumentationPath))
+    ),
+  ];
```
vicb, line 23: We do not need to return 2 plugins, they are just "fake" plugins anyway

```suggestion
  updater.updateContent(
    "patch-instrumentation-next15",
    { filter: /\.(js|mjs|cjs|jsx|ts|tsx)$/, contentFilter: /async loadInstrumentationModule\(/ },
    async ({ contents }) => patchCode(contents, getNext15Rule(builtInstrumentationPath))
  );
  
  updater.updateContent(
    "patch-instrumentation-next14",
    { filter: /\.(js|mjs|cjs|jsx|ts|tsx)$/, contentFilter: /async prepareImpl\(/ },
    async ({ contents }) => patchCode(contents, getNext14Rule(builtInstrumentationPath))
  );

  return {
    "patch implementation",
    setup() {},
  };
```

packages/cloudflare/src/cli/build/patches/plugins/instrumentation.ts
```diff
@@ -0,0 +1,84 @@
...
+        }
+  `;
+}
+
+/**
+ * Gets the instrumentation.js file that the Next.js build process generates when an
+ * instrumentation hook is provided in the app's source
+ *
+ * @param buildOpts the open-next build options
+ * @returns a string pointing to the instrumentation.js file location, or null if such file is not found
```
vicb, line 65: maybe shorten this as the comment is already quite detailed?
```suggestion
 * @returns the path to the instrumentation.js, or null if it doesn't exist
```

packages/cloudflare/src/cli/build/bundle-server.ts
```diff
@@ -91,7 +91,7 @@ export async function bundleServer(buildOpts: BuildOptions): Promise<void> {
       setWranglerExternal(),
       fixRequire(updater),
       handleOptionalDependencies(optionalDependencies),
-      patchLoadInstrumentation(updater),
+      ...patchInstrumentation(updater, buildOpts),
```
vicb, line 94: With my suggestion in the patch file
```suggestion
      patchInstrumentation(updater, buildOpts),
```

packages/cloudflare/src/cli/build/patches/plugins/instrumentation.ts
```diff
@@ -0,0 +1,84 @@
...
+ */
+function getBuiltInstrumentationPath(buildOpts: BuildOptions): string | null {
+  const { outputDir } = buildOpts;
+
+  const baseDir = join(outputDir, "server-functions/default", getPackagePath(buildOpts));
+  const dotNextDir = join(baseDir, ".next");
+  const maybeBuiltInstrumentationPath = join(dotNextDir, "server", `${INSTRUMENTATION_HOOK_FILENAME}.js`);
```
vicb, line 72: maybe?
```suggestion
  const maybeBuiltInstrumentationPath = join(outputDir, "server-functions/default", getPackagePath(buildOpts), `.next/server/${INSTRUMENTATION_HOOK_FILENAME}.js` );
```

packages/cloudflare/src/cli/build/patches/plugins/instrumentation.ts
```diff
@@ -0,0 +1,84 @@
...
+  const builtInstrumentationPath = existsSync(maybeBuiltInstrumentationPath)
+    ? maybeBuiltInstrumentationPath
+    : null;
+
+  return builtInstrumentationPath;
```
vicb, line 77:
```suggestion
  return existsSync(maybeBuiltInstrumentationPath)
    ? maybeBuiltInstrumentationPath
    : null;
```

examples/playground14/e2e/playwright.dev.config.ts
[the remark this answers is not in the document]
dario-piotrowicz: ah I see, the value here is that this is testing that `getCloudflareContext` also works in dev, do you think that that isn't worth testing?

vicb (approved):
Awesome, thanks!

examples/playground14/e2e/playwright.dev.config.ts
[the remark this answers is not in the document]
vicb: Oh yes, you're right, it makes complete sense now that the example is the full playground. My brain hadn't updated yet :)

PR #410, opened 2025-02-25
remove `eval` calls introduced by `depd` wrapped functions

### LINKED ISSUE #291: Investigate e2e failures

dario-piotrowicz:
Some dependencies of Next.js (`raw-body` and `send`) use `depd` to deprecate some of their functions,
`depd` uses `eval` to generate a deprecated version of such functions, this causes `eval` warnings in
the terminal even if these functions are never called, the changes here by patching the depd `wrapfunction`
function so that it still retains the same type of behavior but without using `eval`

fixes #291

### DIFF
--- a/packages/cloudflare/src/cli/build/patches/plugins/patch-depd-deprecations.ts
+++ b/packages/cloudflare/src/cli/build/patches/plugins/patch-depd-deprecations.ts
@@ -35,7 +35,7 @@ rule:
 fix:
   function wrapfunction($FN, $MESSAGE) {
     if(typeof $FN !== 'function') throw new Error("argument fn must be a function");
-    return (...args) => {
+    return function deprecated_$FN(...args) {
       console.warn($MESSAGE);
       return $FN(...args);
     }
--- a/packages/cloudflare/src/cli/build/patches/plugins/patch-depd-deprecations.ts
+++ b/packages/cloudflare/src/cli/build/patches/plugins/patch-depd-deprecations.ts
@@ -1,0 +1,43 @@
+import { patchCode } from "../ast/util.js";
+import type { ContentUpdater } from "./content-updater.js";
+
+/**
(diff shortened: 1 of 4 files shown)


packages/cloudflare/src/cli/build/patches/plugins/patch-depd-deprecations.ts
```diff
@@ -0,0 +1,43 @@
...
+        has:
+          kind: identifier
+          pattern: eval
+fix:
+  function wrapfunction($FN, $MESSAGE) {
+    if(typeof $FN !== 'function') throw new Error("argument fn must be a function");
+    return (...args) => {
+      console.warn($MESSAGE);
+      return $FN(...args);
+    }
```
vicb, line 41: Maybe this to add a name + do not bind?
```suggestion
    return function deprecated_$FN(...args) {
      console.warn($MESSAGE);
      return $FN(...args);
    }
```

vicb (approved):
Thanks!

Issue #223, opened 2025-01-03
[BUG] [ERROR] Could not resolve "uuid"


gethassaan:
### Describe the bug

.open-next/.next/standalone/node_modules/@smithy/middleware-retry/dist-cjs/index.js:51:26: 51 │ var import_uuid = require("uuid");
  
The module "./dist/commonjs-browser/index.js" was not found on the file system:
 
.open-next/.next/standalone/node_modules/@smithy/middleware-retry/node_modules/uuid/package.json:34:19: 34 │         "require": "./dist/commonjs-browser/index.js"
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  
You can mark the path "uuid" as external to exclude it from the bundle, which will remove this error. You can also surround this "require" call with a try/catch block to handle this failure at run-time instead of bundle-time.

### Steps to reproduce

1. clone git@github.com:gethassaan/opennextjs-cloudflare-uuid-issue.git
2. add cloudflare credentials in `.env.local`
3. npm run preview

### Expected behavior

Preview / deployment should not fail with the `uuid` error and build should run smoothly.

### @opennextjs/cloudflare version

0.3.3

### Wrangler version

3.99.0

### next info output

```shell
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.2.0: root:xnu-11215.61.5~2/RELEASE_ARM64_T8103
  Available memory (MB): 16384
  Available CPU cores: 8
Binaries:
  Node: 22.9.0
  npm: 10.9.0
  Yarn: 1.22.22
  pnpm: N/A
Relevant Packages:
  next: 15.1.3 // Latest available version is detected (15.1.3).
  eslint-config-next: 15.1.3
  react: 19.0.0
  react-dom: 19.0.0
  typescript: 5.7.2
Next.js Config:
  output: N/A
```


### Additional context

The build completes without any errors for me, but when I try to deploy or run the preview using `wrangler deploy` or `wrangler dev --port 8771` it thorws an error of `Could not resolve "uuid"`.

allentv:
Running into the same problem too. Hence the upvote

vicb:
including a repro as a github repo that we can clone and investigate would help speed up the investigation.
Thanks

gethassaan:
hey @vicb, I have updated the bug description and I have included the example in this repo:  `git@github.com:gethassaan/opennextjs-cloudflare-uuid-issue.git`

[...]

Issue #219, opened 2025-01-01
[BUG][@opentelemetry] Could not resolve "../platform" 


NayamAmarshe:
### Describe the bug

I'm using firebase + Next.js 14 API Route Handlers. I've added `force-dynamic` to all the required routes and pages but when I do `npm run preview:worker` it throws the following error:

```

> dev:worker
> wrangler dev --port 8771


 ⛅️ wrangler 3.99.0
-------------------

Using vars defined in .dev.vars
Your worker has access to the following bindings:
- Vars:
  - NEXTJS_ENV: "(hidden)"
  - NODE_ENV: "(hidden)"
[wrangler:inf] Ready on http://localhost:8771
▲ [WARNING] Suspicious assignment to defined constant "process.env.NODE_ENV" [assign-to-define]

    .open-next/server-functions/default/handler.mjs:32217:4:
      32217 │     process.env.NODE_ENV = "production";
            ╵     ~~~~~~~~~~~~~~~~~~~~

  The expression "process.env.NODE_ENV" has been configured to be replaced with a constant using the "define" feature. If this expression is supposed to be a compile-time constant, then it doesn't make sense to assign to it here. Or if this expression is supposed to change at run-time, this "define" substitution should be removed.



✘ [ERROR] Build failed with 1 error:

  ✘ [ERROR] Could not resolve "../platform"
  
  
  .open-next/server-functions/default/node_modules/@opentelemetry/api/build/src/internal/global-utils.js:19:27:
        19 │ const platform_1 = require("../platform");
           ╵                            ~~~~~~~~~~~~~
  
  


▲ [WARNING] Suspicious assignment to defined constant "process.env.NODE_ENV" [assign-to-define]

    .open-next/server-functions/default/handler.mjs:32217:4:
      32217 │     process.env.NODE_ENV = "production";
            ╵     ~~~~~~~~~~~~~~~~~~~~

  The expression "process.env.NODE_ENV" has been configured to be replaced with a constant using the "define" feature. If this expression is supposed to be a compile-time constant, then it doesn't make sense to assign to it here. Or if this expression is supposed to change at run-time, this "define" substitution should be removed.


🪵  Logs were written to "/Users/macbook-air/Library/Preferences/.wrangler/logs/wrangler-2025-01-01_00-21-52_196.log"
```

### Steps to reproduce

1. Next.js 14 project with API route handlers and a dynamic page.
2. Use firebase db/auth on both static and dynamic routes.
[...]

PR #412, opened 2025-02-25
add `defineCloudflareConfig` utility

### LINKED ISSUE #114: feat: Easier way to create an OpenNext config for cloudflare
See @conico974 [comment](https://github.com/opennextjs/opennextjs-cloudflare/pull/113#discussion_r1823517211):

> It's probably a good idea to create some kind of function that could create the config object automatically and restrict what options could be set.
I started implemented something like that here https://github.com/opennextjs/opennextjs-aws/blob/main/packages/open-next/src/helpers/withCloudflare.ts
This one won't work but something like that really specific to cloudflare

dario-piotrowicz:
this change adds a new `defineCloudflareConfig` utility that developers can use in their `open-next.config.ts` file to easily generate a configuration compatible with the adapter

___

Docs PR: https://github.com/opennextjs/docs/pull/84

___

resolves #114

### DIFF
--- a/examples/bugs/gh-219/open-next.config.ts
+++ b/examples/bugs/gh-219/open-next.config.ts
@@ -1,25 +1,3 @@
-import type { OpenNextConfig } from "@opennextjs/aws/types/open-next.js";
+import { defineConfig } from "@opennextjs/cloudflare/config";
 
-const config: OpenNextConfig = {
-  default: {
-    override: {
-      wrapper: "cloudflare-node",
-      converter: "edge",
-      // Unused implementation
-      incrementalCache: "dummy",
-      tagCache: "dummy",
-      queue: "dummy",
-    },
-  },
-
-  middleware: {
-    external: true,
-    override: {
-      wrapper: "cloudflare-edge",
-      converter: "edge",
-      proxyExternalRequest: "fetch",
-    },
-  },
-};
-
-export default config;
+export default defineConfig();
--- a/examples/bugs/gh-219/open-next.config.ts
+++ b/examples/bugs/gh-219/open-next.config.ts
@@ -1,3 +1,3 @@
-import { defineConfig } from "@opennextjs/cloudflare/config";
+import { defineCloudflareConfig } from "@opennextjs/cloudflare/config";
 
-export default defineConfig();
+export default defineCloudflareConfig();
--- a/examples/bugs/gh-219/open-next.config.ts
+++ b/examples/bugs/gh-219/open-next.config.ts
@@ -1,3 +1,3 @@
-import { defineCloudflareConfig } from "@opennextjs/cloudflare/config";
+import { defineCloudflareConfig } from "@opennextjs/cloudflare";
 
 export default defineCloudflareConfig();
--- a/examples/bugs/gh-223/open-next.config.ts
+++ b/examples/bugs/gh-223/open-next.config.ts
@@ -1,3 +1,3 @@
-import { defineConfig } from "@opennextjs/cloudflare/config";
+import { defineCloudflareConfig } from "@opennextjs/cloudflare/config";
 
-export default defineConfig();
+export default defineCloudflareConfig();
--- a/examples/bugs/gh-223/open-next.config.ts
+++ b/examples/bugs/gh-223/open-next.config.ts
@@ -1,24 +1,3 @@
-import type { OpenNextConfig } from "@opennextjs/aws/types/open-next";
+import { defineConfig } from "@opennextjs/cloudflare/config";
 
-const config: OpenNextConfig = {
-  default: {
-    override: {
-      wrapper: "cloudflare-node",
-      converter: "edge",
-      incrementalCache: "dummy",
-      tagCache: "dummy",
-      queue: "dummy",
-    },
-  },
-
-  middleware: {
-    external: true,
-    override: {
-      wrapper: "cloudflare-edge",
-      converter: "edge",
-      proxyExternalRequest: "fetch",
-    },
-  },
-};
-
-export default config;
+export default defineConfig();
--- a/examples/bugs/gh-223/open-next.config.ts
+++ b/examples/bugs/gh-223/open-next.config.ts
@@ -1,3 +1,3 @@
-import { defineCloudflareConfig } from "@opennextjs/cloudflare/config";
+import { defineCloudflareConfig } from "@opennextjs/cloudflare";
 
 export default defineCloudflareConfig();
--- a/examples/e2e/app-pages-router/open-next.config.ts
+++ b/examples/e2e/app-pages-router/open-next.config.ts
@@ -1,6 +1,6 @@
+import { defineCloudflareConfig } from "@opennextjs/cloudflare";
 import kvCache from "@opennextjs/cloudflare/kv-cache";
 import memoryQueue from "@opennextjs/cloudflare/memory-queue";
-import { defineCloudflareConfig } from "@opennextjs/cloudflare";
 
 export default defineCloudflareConfig({
   incrementalCache: kvCache,
--- a/examples/e2e/app-pages-router/open-next.config.ts
+++ b/examples/e2e/app-pages-router/open-next.config.ts
@@ -1,6 +1,6 @@
 import kvCache from "@opennextjs/cloudflare/kv-cache";
 import memoryQueue from "@opennextjs/cloudflare/memory-queue";
-import { defineCloudflareConfig } from "@opennextjs/cloudflare/config";
+import { defineCloudflareConfig } from "@opennextjs/cloudflare";
 
 export default defineCloudflareConfig({
   incrementalCache: kvCache,
--- a/examples/e2e/app-pages-router/open-next.config.ts
+++ b/examples/e2e/app-pages-router/open-next.config.ts
@@ -1,8 +1,8 @@
 import kvCache from "@opennextjs/cloudflare/kv-cache";
 import memoryQueue from "@opennextjs/cloudflare/memory-queue";
-import { defineConfig } from "@opennextjs/cloudflare/config";
+import { defineCloudflareConfig } from "@opennextjs/cloudflare/config";
 
-export default defineConfig({
+export default defineCloudflareConfig({
   incrementalCache: kvCache,
   queue: memoryQueue,
 });
--- a/examples/e2e/app-pages-router/open-next.config.ts
+++ b/examples/e2e/app-pages-router/open-next.config.ts
@@ -1,27 +1,8 @@
-import type { OpenNextConfig } from "@opennextjs/aws/types/open-next.js";
 import kvCache from "@opennextjs/cloudflare/kv-cache";
 import memoryQueue from "@opennextjs/cloudflare/memory-queue";
+import { defineConfig } from "@opennextjs/cloudflare/config";
 
-const config: OpenNextConfig = {
-  default: {
-    override: {
-      wrapper: "cloudflare-node",
-      converter: "edge",
-      incrementalCache: () => kvCache,
-      queue: () => memoryQueue,
-      // Unused implementation
-      tagCache: "dummy",
-    },
-  },
-
-  middleware: {
-    external: true,
-    override: {
-      wrapper: "cloudflare-edge",
-      converter: "edge",
-      proxyExternalRequest: "fetch",
-    },
-  },
-};
-
-export default config;
+export default defineConfig({
+  incrementalCache: kvCache,
+  queue: memoryQueue,
+});
--- a/examples/e2e/app-pages-router/open-next.config.ts
+++ b/examples/e2e/app-pages-router/open-next.config.ts
@@ -1,8 +1,8 @@
 import { defineCloudflareConfig } from "@opennextjs/cloudflare";
-import kvCache from "@opennextjs/cloudflare/kv-cache";
+import kvIncrementalCache from "@opennextjs/cloudflare/kv-cache";
 import memoryQueue from "@opennextjs/cloudflare/memory-queue";
 
 export default defineCloudflareConfig({
-  incrementalCache: kvCache,
+  incrementalCache: kvIncrementalCache,
   queue: memoryQueue,
 });
--- a/packages/cloudflare/package.json
+++ b/packages/cloudflare/package.json
@@ -24,6 +24,11 @@
       "types": "./dist/api/index.d.ts",
       "default": "./dist/api/index.js"
     },
+    "./config": {
+      "import": "./dist/api/config.js",
+      "types": "./dist/api/config.d.ts",
+      "default": "./dist/api/config.js"
+    },
     "./*": {
       "import": "./dist/api/*.js",
       "types": "./dist/api/*.d.ts",
--- a/packages/cloudflare/package.json
+++ b/packages/cloudflare/package.json
@@ -24,11 +24,6 @@
       "types": "./dist/api/index.d.ts",
       "default": "./dist/api/index.js"
     },
-    "./config": {
-      "import": "./dist/api/config.js",
-      "types": "./dist/api/config.d.ts",
-      "default": "./dist/api/config.js"
-    },
     "./*": {
       "import": "./dist/api/*.js",
       "types": "./dist/api/*.d.ts",
--- a/packages/cloudflare/src/api/config.ts
+++ b/packages/cloudflare/src/api/config.ts
@@ -16,7 +16,7 @@ export type CloudflareConfigOptions = {
    *     incrementalCache: cache,
    *   });
    */
-  incrementalCache?: IncrementalCache | (() => IncrementalCache | Promise<IncrementalCache>);
+  incrementalCache?: "dummy" | IncrementalCache | (() => IncrementalCache | Promise<IncrementalCache>);
 
   /**
    * The tag cache implementation to use, for more details see the [Caching documentation](https://opennext.js.org/cloudflare/caching))
@@ -32,7 +32,7 @@ export type CloudflareConfigOptions = {
    *     tagCache: cache,
    *   });
    */
-  tagCache?: TagCache | (() => TagCache | Promise<TagCache>);
+  tagCache?: "dummy" | TagCache | (() => TagCache | Promise<TagCache>);
 
   /**
    * The revalidation queue implementation to use, for more details see the [Caching documentation](https://opennext.js.org/cloudflare/caching))
@@ -48,7 +48,7 @@ export type CloudflareConfigOptions = {
    *     queue: memoryQueue,
    *   });
    */
-  queue?: Queue | (() => Queue | Promise<Queue>);
+  queue?: "dummy" | "direct" | Queue | (() => Queue | Promise<Queue>);
 };
 
 /**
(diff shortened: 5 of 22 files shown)


dario-piotrowicz:
@conico974 what I'm doing here is very close to [`withCloudflare`](https://github.com/opennextjs/opennextjs-aws/blob/674dce6937266ae7f838d6343616cbb254a174af/packages/open-next/src/helpers/withCloudflare.ts) but this `defineConfig` is more restricted/focused on just our adapter, while as far as I can tell `withCloudflare` is much more flexible.

Do you think that this API makes sense? or do you think that we should try to align things more?

packages/cloudflare/src/api/config.ts
```diff
@@ -0,0 +1,72 @@
+import { OpenNextConfig } from "@opennextjs/aws/types/open-next";
+import type { IncrementalCache, Queue, TagCache } from "@opennextjs/aws/types/overrides";
+
+export type CloudflareConfigOptions = {
+  /**
+   * The incremental cache implementation to use
+   *
```
conico974, line 7: Add the expected one here for better intellisense (i.e. `import cache from "@opennextjs/cloudflare/kv-cache";`)

packages/cloudflare/src/api/config.ts
```diff
@@ -0,0 +1,72 @@
...
+   * see: https://opennext.js.org/aws/config/overrides/incremental_cache
+   */
+  incrementalCache?: IncrementalCache | (() => IncrementalCache | Promise<IncrementalCache>);
+  /**
+   * The tag cache implementation to use
+   *
```
conico974, line 13: Same as for incrementalCache

packages/cloudflare/src/api/config.ts
```diff
@@ -0,0 +1,72 @@
...
+   * see: https://opennext.js.org/aws/config/overrides/tag_cache
+   */
+  tagCache?: TagCache | (() => TagCache | Promise<TagCache>);
+
+  /**
+   * The revalidation queue implementation to use
+   *
```
conico974, line 20: Same as for incrementalCache

conico974 (approved):
LGTM, just few nit

>Do you think that this API makes sense? or do you think that we should try to align things more?

Yeah i think it makes perfect sense, this is enough. If people needs more complex options they can just provide their own `open-next.config.ts`

packages/cloudflare/src/api/config.ts
```diff
@@ -0,0 +1,72 @@
...
+      external: true,
+      override: {
+        wrapper: "cloudflare-edge",
+        converter: "edge",
+        proxyExternalRequest: "fetch",
+      },
+    },
+
+    dangerous: {
+      enableCacheInterception: false,
```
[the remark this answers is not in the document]
dario-piotrowicz: ok I'll remove it, I just copied it from what we already had 🙂👍

packages/cloudflare/package.json
```diff
@@ -24,6 +24,11 @@
       "types": "./dist/api/index.d.ts",
       "default": "./dist/api/index.js"
     },
+    "./config": {
```
vicb, line 27: is this needed given l32?
should this be re-exported from `api/index.ts` (by name)?

packages/cloudflare/src/api/config.ts
```diff
@@ -0,0 +1,68 @@
...
+
+function resolveOverride<T extends IncrementalCache | TagCache | Queue>(
+  value: T | (() => T | Promise<T>) | undefined
+): (() => Promise<T>) | "dummy" {
+  if (!value) {
+    return "dummy";
+  }
+
+  if (typeof value === "function") {
+    return async () => value();
```
vicb, line 92:
```suggestion
    return async () => await value();
```

vicb (approved):

examples/bugs/gh-219/open-next.config.ts
```diff
@@ -1,25 +1,3 @@
...
-    override: {
-      wrapper: "cloudflare-edge",
-      converter: "edge",
-      proxyExternalRequest: "fetch",
-    },
-  },
-};
-
-export default config;
+export default defineCloudflareConfig();
```
vicb, line 3: incrementalCache: "dummy" ?

examples/bugs/gh-223/open-next.config.ts
```diff
@@ -1,24 +1,3 @@
...
-    override: {
-      wrapper: "cloudflare-edge",
-      converter: "edge",
-      proxyExternalRequest: "fetch",
-    },
-  },
-};
-
-export default config;
+export default defineCloudflareConfig();
```
vicb, line 3: incrementalCache: "dummy" ?

packages/cloudflare/src/api/config.ts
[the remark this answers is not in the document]
dario-piotrowicz: removed 🙂

packages/cloudflare/package.json
vicb: what about also re-exporting cache stuff from the index? 
/cc @james-elicx

packages/cloudflare/src/api/config.ts
dario-piotrowicz: I've added some details

Screenshot 2025-02-26 at 10 36 30

wdyt? 🙂

packages/cloudflare/src/api/config.ts
dario-piotrowicz: Screenshot 2025-02-26 at 10 37 11

does this look good? 🙂

packages/cloudflare/src/api/config.ts
dario-piotrowicz: Screenshot 2025-02-26 at 10 37 24

does this look good? 🙂

examples/bugs/gh-219/open-next.config.ts
dario-piotrowicz: sorry I don't understand the question

dario-piotrowicz: (`"dummy"` is not a valid value for this API)

packages/cloudflare/package.json
dario-piotrowicz: > is this needed given l32?

no, my bad, thanks 👍

packages/cloudflare/src/api/config.ts
conico974: Yeah looks good

packages/cloudflare/package.json
dario-piotrowicz: > should this be re-exported from api/index.ts (by name)?

done

packages/cloudflare/src/api/config.ts
conico974: Looks good as well

packages/cloudflare/src/api/config.ts
conico974: Not sure if we should add a warning here or maybe update the docs to add a warning that this may not be fully suitable for production

packages/cloudflare/package.json
james-elicx: > what about also re-exporting cache stuff from the index?

My general preference is to have explicit imports for optional features in a public API, instead of lumping it in a barrel file. In this case, defineCloudflareConfig should always be used, but the cache adapters are all entirely optional + independent, and may/may not be used depending on use case.

I don't particularly mind either way, but that's my 2c.

packages/cloudflare/src/api/config.ts
dario-piotrowicz: @vicb @james-elicx any thoughts on the warning?

james-elicx: Maybe we could add "See the docs for more information" and a link to the docs here? Agree with some sort of warning, but I feel like it would be better to live in the docs and then defer people to the docs to find out more. What do you think?

packages/cloudflare/package.json
vicb: ATM we are in this weird state where `index.ts` re-export some of the public API but not all - it can be confusing.

Any drawback re-exporting from the barrel file? If the files can be tree shaken, I would prefer that.

packages/cloudflare/src/api/config.ts
dario-piotrowicz: mh... I'm already adding a link to the aws docs and adding two links seems too much noisy for me 😕

So I think that your suggestion would boil down to:
 - creating a cloudflare doc page (or pages) for these overrides (I think that if we do it we need to do this for all overrides not just the queue one)
 - having those cloudflare docs point to the aws ones for more in depth-details
 - in these overrides point to this new doc page/pages
 
 I don't mind the idea, as it might be valuable to add those docs in any case right?
 that said I think that a warning in the jsdoc is definitely going to be more visible in my opinion
 
Not to unnecessarily block (/scope creep) this PR I would just add a jsdoc comment here right now and create an issue to update the open-next cloudflare docs + these comments and take it from there

examples/bugs/gh-219/open-next.config.ts
vicb: nevermind

packages/cloudflare/package.json
vicb: d1TagCache, kvIncrementalCache, memoryQueue ?

dario-piotrowicz: done

packages/cloudflare/src/api/config.ts
james-elicx: The link would be to the caching page in the Cloudflare docs - they're all in there.

james-elicx: (not a blocker from me)

dario-piotrowicz: I see, ok I missed that one, ok I can point to that one thanks 🙂

packages/cloudflare/src/api/index.ts
```diff
@@ -1 +1,5 @@
 export * from "./cloudflare-context.js";
+export * from "./config.js";
```
vicb, line 2: Can we have an explicit list of exported APIs instead of `*`

packages/cloudflare/templates/defaults/open-next.config.ts
```diff
@@ -1,26 +1,7 @@
 // default open-next.config.ts file created by @opennextjs/cloudflare
+import { defineCloudflareConfig } from "@opennextjs/cloudflare/dist/api/config";
```
vicb, line 2: import from `@opennextjs/cloudflare` ?

packages/cloudflare/package.json
dario-piotrowicz: oof this is annoying.... exporting them from the index file doesn't work well because this makes `next build` also import them when it `require`s the Next.js config file (since we import `setupOpenNextCloudflareForDev` there) and the problem is that Next's module resolution doesn't resolve all the necessary files 😕 

Screenshot 2025-02-26 at 14 50 04

So this is actually no working, potentially changes upstream in aws would be needed to enable this 😕

packages/cloudflare/templates/defaults/open-next.config.ts
dario-piotrowicz: oops, good spot 👍

packages/cloudflare/package.json
dario-piotrowicz: @vicb I'm going to revert the change and keep the utils importable from the other files unless you have some suggestion?

vicb (approved):
Thanks!

packages/cloudflare/package.json
vicb: I guess revert the change for now and figure out later...
Looks like there is a mix of import with and w/o extensions

packages/cloudflare/src/api/config.ts
dario-piotrowicz: Updated @james-elicx , @conico974 please give the comments + the updated docs page a quick last look 🙏

james-elicx (approved):

packages/cloudflare/src/api/config.ts
```diff
@@ -0,0 +1,98 @@
...
+
+function resolveOverride<T extends IncrementalCache | TagCache | Queue>(
+  value: T | (() => T | Promise<T>) | undefined
+): (() => Promise<T>) | "dummy" {
+  if (!value) {
+    return "dummy";
+  }
+
+  if (typeof value === "function") {
+    return async () => await value();
```
james-elicx, line 94: not the end of the world, but might be good to wrap the value in Promise.resolve in case it's not async

vicb: Does it change anything? 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#description

james-elicx: not particularly

james-elicx: it ends up becoming more of a point of whether you prefer ensuring what you're awaiting is a promise. in other code bases i work in we have an eslint rule for it.

i imagine we could remove the async from this altogether and it would function the same

packages/cloudflare/src/api/config.ts
```diff
@@ -3,7 +3,9 @@ import type { IncrementalCache, Queue, TagCache } from "@opennextjs/aws/types/ov
 
 export type CloudflareConfigOptions = {
   /**
-   * The incremental cache implementation to use (for more details see the [Incremental Cache documentation](https://opennext.js.org/aws/config/overrides/incremental_cache))
+   *
+   *
```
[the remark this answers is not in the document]
dario-piotrowicz: oops... thanks 🙂👍

packages/cloudflare/src/api/index.ts
dario-piotrowicz: done

packages/cloudflare/src/api/config.ts
james-elicx: In the app-router config, we're currently using tagCache and queue without an async function...

https://github.com/opennextjs/opennextjs-cloudflare/blob/fdc58c1ed1d283cbb57c283f6b9902209b0e9b46/examples/e2e/app-router/open-next.config.ts#L11-L13

conico974: Yeah the typing is not correct here, we just await things in aws. It should have been `Promise<T> | T` https://github.com/opennextjs/opennextjs-aws/blob/e48951fcb5882b536e67581feac489cd1689ffe9/packages/open-next/src/core/createMainHandler.ts#L30-L36

FYI, when you'll start using multiple worker, it would be interesting to start using `import("./kvCache")` instead of top level import to lazy load these overrides. (The `open-next.config.ts` is always one of the first thing that is loaded and it is blocking execution)

dario-piotrowicz: https://github.com/opennextjs/opennextjs-cloudflare/pull/412/commits/04cf758edcb7cea7dd43038aace24d837a365995 better? 🙂

dario-piotrowicz: Nope that doesn't work 🤔 

It does need to be a function that returns promise, see the [aws `incrementalCache` type](https://github.com/opennextjs/opennextjs-aws/blob/e48951fcb5882b536e67581feac489cd1689ffe9/packages/open-next/src/types/open-next.ts#L195-L201)

james-elicx: I think the types should probably be updated. Don't consider this a blocker though.

vicb: https://github.com/opennextjs/opennextjs-aws/pull/755

packages/cloudflare/src/api/config.ts
```diff
@@ -0,0 +1,96 @@
...
+   * @example
+   *   import { defineCloudflareConfig } from "@opennextjs/cloudflare";
+   *   import cache from "@opennextjs/cloudflare/kv-cache";
+   *
+   *   export default defineCloudflareConfig({
+   *     incrementalCache: cache,
+   *   });
+   */
+  incrementalCache?: IncrementalCache | (() => IncrementalCache | Promise<IncrementalCache>);
```
vicb, line 19:
```suggestion
  incrementalCache?: "dummy" | IncrementalCache | (() => IncrementalCache | Promise<IncrementalCache>);
```

packages/cloudflare/src/api/config.ts
```diff
@@ -0,0 +1,96 @@
...
+   *
+   * @example
+   *   import { defineCloudflareConfig } from "@opennextjs/cloudflare";
+   *   import cache from "@opennextjs/cloudflare/d1-tag-cache";
+   *
+   *   export default defineCloudflareConfig({
+   *     tagCache: cache,
+   *   });
+   */
+  tagCache?: TagCache | (() => TagCache | Promise<TagCache>);
```
vicb, line 35:
```suggestion
  tagCache?: "dummy" | TagCache | (() => TagCache | Promise<TagCache>);
```

packages/cloudflare/src/api/config.ts
```diff
@@ -0,0 +1,96 @@
...
+   *
+   * @example
+   *   import { defineCloudflareConfig } from "@opennextjs/cloudflare";
+   *   import memoryQueue from "@opennextjs/cloudflare/memory-queue";
+   *
+   *   export default defineCloudflareConfig({
+   *     queue: memoryQueue,
+   *   });
+   */
+  queue?: Queue | (() => Queue | Promise<Queue>);
```
vicb, line 51:
```suggestion
  queue?: "dummy" | "direct" | Queue | (() => Queue | Promise<Queue>);
```

packages/cloudflare/src/api/config.ts
```diff
@@ -0,0 +1,96 @@
...
+): (() => T | Promise<T>) | "dummy" {
+  if (!value) {
+    return "dummy";
+  }
+
+  if (typeof value === "function") {
+    return () => value();
+  }
+
+  return async () => value;
```
vicb, line 95:
```suggestion
  return typeof value === "function" ? value : () => value;
```

examples/e2e/app-pages-router/open-next.config.ts
```diff
@@ -1,27 +1,8 @@
-import type { OpenNextConfig } from "@opennextjs/aws/types/open-next.js";
+import { defineCloudflareConfig } from "@opennextjs/cloudflare";
 import kvCache from "@opennextjs/cloudflare/kv-cache";
```
vicb, line 2: nit: across the files this is imported as `cache`, `kvCache`, or `incrementalCache`. Maybe you can make the name consistent in this PR?

packages/cloudflare/src/api/d1-tag-cache.ts
```diff
@@ -1,6 +1,6 @@
 import { debug, error } from "@opennextjs/aws/adapters/logger.js";
 import type { OpenNextConfig } from "@opennextjs/aws/types/open-next.js";
-import type { TagCache } from "@opennextjs/aws/types/overrides.js";
+import type { OriginalTagCache } from "@opennextjs/aws/types/overrides.js";
```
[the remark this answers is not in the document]
vicb: Oh right!
Thanks for clarifying.

examples/e2e/app-pages-router/open-next.config.ts
dario-piotrowicz: done

vicb (approved):

PR #417, opened 2025-02-26
define __filename

vicb:
jonkwheeler reported on Discord that there is an error with `__filename` not being defined, this should help


james-elicx (approved):

vicb:
Thanks James!

PR #421, opened 2025-02-26
chore: disable typescript + eslint during next.js builds

james-elicx:
The typescript and eslint checks slow down our e2e builds while providing zero value. I would like to propose disabling them completely. The e2es we copied from aws already do this.

It's not a significant gain, but it cuts off about 15-20 seconds from time spent building apps.

### DIFF
--- a/examples/bugs/gh-119/next.config.ts
+++ b/examples/bugs/gh-119/next.config.ts
@@ -1,7 +1,8 @@
 import type { NextConfig } from "next";
 
 const nextConfig: NextConfig = {
-  /* config options here */
+  typescript: { ignoreBuildErrors: true },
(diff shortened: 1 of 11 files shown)


vicb:
I agree it's worth running those each time.

Maybe  running them "sometimes" can have some value (locally vs CI, first day of the week, randomly).

WDYT

james-elicx:
> I agree it's worth running those each time.
> 
> Maybe running them "sometimes" can have some value (locally vs CI, first day of the week, randomly).
> 
> WDYT

I think it would make more sense to run them as part of the lint/ts:check/format workflow if we want to keep them instead of relying on the app build

vicb (approved):

PR #436, opened 2025-03-03
feat: auto-populating d1 cache data

james-elicx:
- Adds a CLI flag to enable auto-population of cache data for either local or remote.
- Adds a CLI flag to only populate cache (i.e. skip build completely).
- Populates D1 tag cache data when the adapter is used.

Docs PR and KV change to use auto-population will follow if we agree on this being the way forward

---
This is **part 1 of 3 in a stack** made with GitButler:
- <kbd>&nbsp;3&nbsp;</kbd> #479 
- <kbd>&nbsp;2&nbsp;</kbd> #443 
- <kbd>&nbsp;1&nbsp;</kbd> #436 👈

### DIFF
--- a/packages/cloudflare/src/cli/args.ts
+++ b/packages/cloudflare/src/cli/args.ts
@@ -10,7 +10,7 @@ export function getArgs(): {
   skipWranglerConfigCheck: boolean;
   outputDir?: string;
   minify: boolean;
-  populateCache?: { mode: CacheBindingMode; onlyPopulate: boolean };
+  populateCache?: { mode: CacheBindingMode; onlyPopulateWithoutBuilding: boolean };
 } {
   const { skipBuild, skipWranglerConfigCheck, output, noMinify, populateCache, onlyPopulateCache } =
     parseArgs({
@@ -63,7 +63,9 @@ export function getArgs(): {
       skipWranglerConfigCheck ||
       ["1", "true", "yes"].includes(String(process.env.SKIP_WRANGLER_CONFIG_CHECK)),
     minify: !noMinify,
-    populateCache: populateCache ? { mode: populateCache, onlyPopulate: !!onlyPopulateCache } : undefined,
+    populateCache: populateCache
+      ? { mode: populateCache, onlyPopulateWithoutBuilding: !!onlyPopulateCache }
+      : undefined,
   };
 }
 
--- a/packages/cloudflare/src/cli/args.ts
+++ b/packages/cloudflare/src/cli/args.ts
@@ -2,14 +2,15 @@ import { mkdirSync, type Stats, statSync } from "node:fs";
 import { resolve } from "node:path";
 import { parseArgs } from "node:util";
 
-import { CacheBindingMode } from "./build/utils/index.js";
+import type { CacheBindingMode } from "./build/utils/index.js";
+import { isCacheBindingMode } from "./build/utils/index.js";
 
 export function getArgs(): {
   skipNextBuild: boolean;
   skipWranglerConfigCheck: boolean;
   outputDir?: string;
   minify: boolean;
-  populateCache?: { mode: "local" | "remote"; onlyPopulate: boolean };
+  populateCache?: { mode: CacheBindingMode; onlyPopulate: boolean };
 } {
   const { skipBuild, skipWranglerConfigCheck, output, noMinify, populateCache, onlyPopulateCache } =
     parseArgs({
@@ -50,7 +51,7 @@ export function getArgs(): {
 
   if (
     (populateCache !== undefined || onlyPopulateCache) &&
-    (!populateCache?.length || !["local", "remote"].includes(populateCache))
+    (!populateCache?.length || !isCacheBindingMode(populateCache))
   ) {
     throw new Error(`Error: missing mode for populate cache flag, expected 'local' | 'remote'`);
   }
@@ -62,9 +63,7 @@ export function getArgs(): {
       skipWranglerConfigCheck ||
       ["1", "true", "yes"].includes(String(process.env.SKIP_WRANGLER_CONFIG_CHECK)),
     minify: !noMinify,
-    populateCache: populateCache
-      ? { mode: populateCache as CacheBindingMode, onlyPopulate: !!onlyPopulateCache }
-      : undefined,
+    populateCache: populateCache ? { mode: populateCache, onlyPopulate: !!onlyPopulateCache } : undefined,
   };
 }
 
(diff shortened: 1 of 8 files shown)


packages/cloudflare/src/cli/args.ts
```diff
@@ -2,48 +2,69 @@ import { mkdirSync, type Stats, statSync } from "node:fs";
...
 
   if (outputDir) {
     assertDirArg(outputDir, "output", true);
   }
 
+  if (
+    (populateCache !== undefined || onlyPopulateCache) &&
+    (!populateCache?.length || !["local", "remote"].includes(populateCache))
+  ) {
+    throw new Error(`Error: missing mode for populate cache flag, expected 'local' | 'remote'`);
```
conico974, line 55: Should this default to `local` ?

packages/cloudflare/src/cli/build/utils/populate-cache.ts
```diff
@@ -0,0 +1,61 @@
...
+    opts.packager,
+    ["exec", "wrangler", ...args, mode === "remote" && "--remote"].filter((v): v is string => !!v),
+    {
+      shell: true,
+      stdio: ["ignore", "ignore", "inherit"],
+    }
+  );
+
+  if (result.status !== 0) {
+    logger.error("Failed to populate cache");
```
conico974, line 37: Could this fail because the DB does not exist yet ? If that's the case, it should have a specific message for that case to avoid confusion

packages/cloudflare/src/cli/build/utils/populate-cache.ts
```diff
@@ -0,0 +1,61 @@
...
+    logger.info("Successfully populated cache");
+  }
+}
+
+export async function populateCache(opts: BuildOptions, config: OpenNextConfig, mode: CacheBindingMode) {
+  const { tagCache } = config.default.override ?? {};
+
+  if (tagCache) {
+    const name = await resolveCacheName(tagCache);
+    switch (name) {
```
conico974, line 48: `default` should probably log a message saying that it will do nothing

packages/cloudflare/src/cli/build/utils/populate-cache.ts
```diff
@@ -0,0 +1,61 @@
...
+  LazyLoadedOverride,
+  OpenNextConfig,
+} from "@opennextjs/aws/types/open-next.js";
+import type { IncrementalCache, TagCache } from "@opennextjs/aws/types/overrides.js";
+
+export type CacheBindingMode = "local" | "remote";
+
+async function resolveCacheName(
+  value:
+    | IncludedIncrementalCache
```
conico974, line 18: Incremental cache is not used here

packages/cloudflare/src/cli/build/utils/populate-cache.ts
```diff
@@ -0,0 +1,61 @@
...
+  if (tagCache) {
+    const name = await resolveCacheName(tagCache);
+    switch (name) {
+      case "d1-tag-cache": {
+        logger.info("\nPopulating D1 tag cache...");
+
+        runWrangler(opts, mode, [
+          "d1 execute",
+          "NEXT_CACHE_D1",
+          `--file ${JSON.stringify(path.join(opts.outputDir, "cloudflare/cache-assets-manifest.sql"))}`,
```
conico974, line 55: There is some case where this file would not be there ( `--onlyPopulate` but build wasn't run or if `disableIncrementalCache` or `disableTagCache` is set in `open-next.config.ts`)
Adding a check for the file before running the command is probably better

packages/cloudflare/src/cli/build/utils/populate-cache.ts
james-elicx: Any errors from wrangler will already be logged to the console due to the `inherit` for the `stderr` part of `stdio`, so it feels like we'd just be trying to tell the user the same thing as any error that is already shown to them if we did that

packages/cloudflare/src/cli/build/utils/populate-cache.ts
james-elicx: I've put a check for the output directory and the disable flags. Let me know if you think it would be better to check the specific file instead of just the output directory.

packages/cloudflare/src/cli/args.ts
james-elicx: It's a good question. I thought it would be better to have it be an explicit opt-in at first. I'm happy to change it if you want though, or maybe this is probably something that would be worth discussing as a team about opt-in vs opt-out. My only concern about this would be whether people would assume that their remote cache is being populated automatically without them having to explicitly state remote.

conico974: It should probably be discussed with the team to see what they think.
My reasoning was that `wrangler execute` by default runs locally and do the same for this.

I'm wondering if a more long term solution would be to have 2 commands here `preview` and `deploy` that would run the corresponding wrangler commands but with the cache populated at the right place. This also opens the possibility for more custom deployment in the future (i.e. function splitting or true external middleware).

packages/cloudflare/src/cli/build/utils/populate-cache.ts
```diff
@@ -0,0 +1,74 @@
...
+  if (!existsSync(opts.outputDir)) {
+    logger.error("Unable to populate cache: Open Next build not found");
+    process.exit(1);
+  }
+
+  if (!config.dangerous?.disableIncrementalCache && incrementalCache) {
+    logger.info("Incremental cache does not need populating");
+  }
+
+  if (!config.dangerous?.disableTagCache && tagCache) {
```
conico974, line 57:
```suggestion
  if (!config.dangerous?.disableTagCache && !config.dangerous?.disableIncrementalCache && tagCache) {
```
If `disableIncrementalCache` is set the file would not be generated either

packages/cloudflare/src/cli/args.ts
james-elicx: That's a nice idea

packages/cloudflare/src/cli/build/build.ts
```diff
@@ -62,6 +63,11 @@ export async function build(projectOpts: ProjectOptions): Promise<void> {
   logger.info(`@opennextjs/cloudflare version: ${cloudflare}`);
   logger.info(`@opennextjs/aws version: ${aws}`);
 
+  if (projectOpts.populateCache?.onlyPopulate) {
```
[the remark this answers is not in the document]
james-elicx: Yeah Nico had a similar comment. I think it makes sense, but would probably better as a general `deploy` command rather than a cache-specific one, but I feel like this is something that would be good to discuss as Vic may have some plans already around what commands we will expose.

vicb (approved):

conico974 (approved):
LGTM

Issue #373, opened 2025-02-12
[BUG] Dynamic Pages and PPR broken on Next.js Canary Above 15.1.1-canary.6


wesbos:
### Describe the bug

Been using Next.js Canary so I can get PPR support, but any page or component that uses dynamic components will cause the site to crash.

The error is the usual `   [cause]: Error: Invariant: renderHTML should not be called in minimal mode        at Object.renderHTMLImpl` but I did see some errors around pipe and stream in my cloudflare dashboard.

I've narrowed it down to working last in `next@15.1.1-canary.6`, so something in 15.1.1-canary.7 broke. I confirmed with the latest `15.2.0-canary.55` as well. 

Normally I wouldn't complain about canary support, but PPR is only available in canary at the moment, and the opennextjs docs say PPR is supported.

This seems to be an issue bigger that PPR, and any dynamic content is causing it to throw an error - so likely shoudl be fixed for when Next.js 15.2.0 is released

### Steps to reproduce

1. Go to https://github.com/wesbos/opennextjs-error-replication
2. pnpm install
3. pnpm run preview

### Expected behavior

it works

### @opennextjs/cloudflare version

0.5.0

### Wrangler version

3.108.1

### next info output

```shell
x
```

vicb:
Thanks for your report @wesbos - we'll look at that shortly

> Normally I wouldn't complain about canary support, but PPR is only available in canary at the moment, and the opennextjs docs say PPR is supported.

No problem, your feedback has been super valuable, keep it coming.

Issue #333, opened 2025-02-04
[BUG] Error: Invariant: renderHTML should not be called in minimal mode


verdverm:
### Describe the bug

I'm trying to make a post to the following NextJS API route and get this error back after the wrangler / opennext build. Not sure what's causing it

```
import { NextRequest, NextResponse } from 'next/server';
import { isValidHandle } from '@atproto/syntax'

import { client as atprotoOauthClient } from '@/lib/atproto/oauth/client'

type Session = { did: string }

export async function POST(req: NextRequest) {
  const data = await req.json()

  // Validate
  const handle = data.handle
  if (typeof handle !== 'string' || !isValidHandle(handle)) {
    return NextResponse.json({error: "invalid handle"})
  }

  // Initiate the OAuth flow
  try {
    const url = await atprotoOauthClient.authorize(handle, {
      scope: 'atproto transition:generic',
    })
    return NextResponse.json({ redirect: url.toString() })
  } catch (err) {
    return NextResponse.json({error: "oauth authorize failed", details: err, message: err.message})
  }
}
```

### Steps to reproduce

Not sure yet, can get CF trace details

I can see about producing a minimal reproducer in the next couple of days

### Expected behavior

I can post to oauth endpoint and not get errors

### @opennextjs/cloudflare version

0.5.3

### Wrangler version

3.109.2

### next info output

```shell
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.3.0: Thu Jan  2 20:23:36 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T8112
  Available memory (MB): 24576
  Available CPU cores: 8
Binaries:
  Node: 21.7.1
  npm: 10.5.0
  Yarn: 1.22.21
  pnpm: 8.14.0
Relevant Packages:
  next: 15.1.7 // Latest available version is detected (15.1.7).
  eslint-config-next: 15.1.0
  react: 19.0.0
  react-dom: 19.0.0
  typescript: 5.7.3
Next.js Config:
  output: standalone
```

### Additional context

running site is at https://verdverm.verdverm.workers.dev

verdverm:
hmm, have a feeling it might be `dotenv`, I think I saw an error at some earlier point in previous versions of the code about `fs.readFile` not being implemented yet...

either way the error message coming out doesn't help me to RCA this

verdverm:
[...]

PR #441, opened 2025-03-06
Fix for next 15.2 and renderHTML should not be called in minimal mode

### LINKED ISSUE #432: [BUG] next 15.2.0 breaks non static routes
### Describe the bug

I've upgraded my app to next 15.2.0 and now i'm getting a 500 on some routes.

From a quick check it seems to only happen on the routes that don't use `generateStaticParams`

```
✘ [ERROR] ⨯ Error: Invariant: renderHTML should not be called in minimal mode

      at Object.renderHTMLImpl
  (/path/to/proj/.wrangler/tmp/dev-Cd8zgp/worker.js)
      at <unknown>
  (/path/to/proj/.wrangler/tmp/dev-Cd8zgp/worker.js)
      at Object.trace
  (/path/to/proj/.wrangler/tmp/dev-Cd8zgp/worker.js)
      at Object.renderHTML
  (/path/to/proj/.wrangler/tmp/dev-Cd8zgp/worker.js)
      at doRender
  (/path/to/proj/.wrangler/tmp/dev-Cd8zgp/worker.js)
      at responseGenerator
  (/path/to/proj/.wrangler/tmp/dev-Cd8zgp/worker.js)
      at Object.get
  (/path/to/proj/.wrangler/tmp/dev-Cd8zgp/worker.js)
      at Object.renderToResponseWithComponentsImpl
```

Versions of opennext/next/vercel:

```
"next": "^15.2.0",
"@opennextjs/aws": "^3.5.0",
"@opennextjs/cloudflare": "^0.5.7",
"vercel": "^41.3.0",
```

### Steps to reproduce

1. create a route without `generateStaticParams`
2. run `opennextjs-cloudflare && wrangler dev`
3. open in browser and see error 500

### Expected behavior

1. no 500 should be thrown for non static routes

0.5.7

### Wrangler version

3.111.0

### next info output

```shell
Operating System:
  Platform: linux
  Arch: x64
  Version: #1 SMP PREEMPT_DYNAMIC Thu, 27 Feb 2025 18:09:44 +0000
  Available memory (MB): 31302
  Available CPU cores: 16
Binaries:
  Node: 20.18.2
  npm: 10.8.2
  Yarn: 1.22.19
  pnpm: 10.1.0
Relevant Packages:
  next: 15.2.0 // Latest available version is detected (15.2.0).
  eslint-config-next: 15.2.0
  react: 19.0.0
  react-dom: 19.0.0
  typescript: 5.8.2
Next.js Config:
  output: N/A
```

### Additional context

_No response_

conico974:
Should fix #432 #373 and #333

Next 15.2 introduced a change that create an `AbortController` https://github.com/vercel/next.js/pull/73975/files as a const outside of the request context.
This `AbortController` cause an issue while requiring app router pages that makes all these require return `{}` instead of the actual module in workerd. (see the comment for more detail)

It also replace the define of `process.env.NEXT_MINIMAL` inside of esbuild by a custom plugin that keep the `renderHTML` that can be used as a fallback by Next in case of error.

It also bump next version to 15.2.0 for the e2e examples

It's a draft for now, if workerd is not expected to throw for this case we could remove almost everything in this PR and just bump wrangler once fixed and merged

### DIFF
--- a/packages/cloudflare/src/cli/build/bundle-server.ts
+++ b/packages/cloudflare/src/cli/build/bundle-server.ts
@@ -17,6 +17,7 @@ import { patchFetchCacheSetMissingWaitUntil } from "./patches/plugins/fetch-cach
 import { inlineFindDir } from "./patches/plugins/find-dir.js";
 import { patchInstrumentation } from "./patches/plugins/instrumentation.js";
 import { inlineLoadManifest } from "./patches/plugins/load-manifest.js";
+import { patchNextMinimal } from "./patches/plugins/next-minimal.js";
 import { handleOptionalDependencies } from "./patches/plugins/optional-deps.js";
 import { patchDepdDeprecations } from "./patches/plugins/patch-depd-deprecations.js";
 import { fixRequire } from "./patches/plugins/require.js";
@@ -99,6 +100,7 @@ export async function bundleServer(buildOpts: BuildOptions): Promise<void> {
       inlineLoadManifest(updater, buildOpts),
       inlineBuildId(updater),
       patchDepdDeprecations(updater),
+      patchNextMinimal(updater),
       // Apply updater updaters, must be the last plugin
       updater.plugin,
     ],
@@ -136,7 +138,8 @@ export async function bundleServer(buildOpts: BuildOptions): Promise<void> {
       // We make sure that environment variables that Next.js expects are properly defined
       "process.env.NEXT_RUNTIME": '"nodejs"',
       "process.env.NODE_ENV": '"production"',
-      "process.env.NEXT_MINIMAL": "true",
+      "process.env.TURBOPACK": "false",
+      // Ideally here we should define the `process.env.__NEXT_EXPERIMENTAL_REACT`, but this would require that function splitting is enabled for the case where both versions are needed (i.e. page and app router routes)
     },
     platform: "node",
     banner: {
--- a/packages/cloudflare/src/cli/build/bundle-server.ts
+++ b/packages/cloudflare/src/cli/build/bundle-server.ts
@@ -138,8 +138,11 @@ export async function bundleServer(buildOpts: BuildOptions): Promise<void> {
       // We make sure that environment variables that Next.js expects are properly defined
       "process.env.NEXT_RUNTIME": '"nodejs"',
       "process.env.NODE_ENV": '"production"',
+      // The 2 following defines are used to reduce the bundle size by removing unnecessary code
+      // Next uses different precompiled renderers (i.e. `app-page.runtime.prod.js`) based on if you use `TURBOPACK` or some experimental React features
+      // Turbopack is not supported for build at the moment, so we disable it
       "process.env.TURBOPACK": "false",
-      // This define should be safe to use for Next 14.2+
+      // This define should be safe to use for Next 14.2+, earlier versions (13.5 and less) will cause trouble
       "process.env.__NEXT_EXPERIMENTAL_REACT": `${needsExperimentalReact(nextConfig)}`,
     },
     platform: "node",
--- a/packages/cloudflare/src/cli/build/bundle-server.ts
+++ b/packages/cloudflare/src/cli/build/bundle-server.ts
@@ -23,7 +23,7 @@ import { patchDepdDeprecations } from "./patches/plugins/patch-depd-deprecations
 import { fixRequire } from "./patches/plugins/require.js";
 import { shimRequireHook } from "./patches/plugins/require-hook.js";
 import { setWranglerExternal } from "./patches/plugins/wrangler-external.js";
-import { normalizePath, patchCodeWithValidations } from "./utils/index.js";
+import { needsExperimentalReact, normalizePath, patchCodeWithValidations } from "./utils/index.js";
 
 /** The dist directory of the Cloudflare adapter package */
 const packageDistDir = path.join(path.dirname(fileURLToPath(import.meta.url)), "../..");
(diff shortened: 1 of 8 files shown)


packages/cloudflare/src/cli/build/patches/plugins/next-minimal.ts
```diff
@@ -0,0 +1,92 @@
+import { patchCode } from "../ast/util.js";
+import { ContentUpdater } from "./content-updater.js";
+
+// We try to be as specific as possible to avoid patching the wrong thing here
+// It seems that there is a bug in the worker runtime. When the AbortController is created outside of the request context it throws an error (not sure if it's expected or not) except in this case.
```
vicb, line 5: Workers can not do IO outside of the the request - there should be a clear error message in this case (and it should be easy enough to try outside of the context of Next).

Note that we also plan to not run any Next code oustide of a request context which should also help here.

packages/cloudflare/src/cli/build/patches/plugins/next-minimal.spec.ts
```diff
@@ -0,0 +1,76 @@
+import { describe, expect, test } from "vitest";
+
+import { patchCode } from "../ast/util";
+import { abortControllerRule } from "./next-minimal";
+
+const appPageRuntimeProdJs = `let p = new AbortController;
```
vicb, line 6: Do we need to have test for both plain an minified code? (i.e. does this code change whether debug is enabled?)

pnpm-workspace.yaml
```diff
@@ -42,7 +42,7 @@ catalogs:
     "@types/react-dom": 19.0.0
     "@types/react": 19.0.0
     autoprefixer: 10.4.15
-    next: 15.1.0
+    next: 15.2.0
```
vicb, line 45: question: what ex is using that?

packages/cloudflare/src/cli/build/bundle-server.ts
```diff
@@ -136,7 +138,8 @@ export async function bundleServer(buildOpts: BuildOptions): Promise<void> {
       // We make sure that environment variables that Next.js expects are properly defined
       "process.env.NEXT_RUNTIME": '"nodejs"',
       "process.env.NODE_ENV": '"production"',
-      "process.env.NEXT_MINIMAL": "true",
```
vicb, line 139: Does that affect the bundle size (`pnpm wrangler dev --dry-run`)?

vicb (commented):
Thanks @conico974 for this fix.

I think the workerd error is expected - but we should fix this by not running next code at top level. It will also help with other issues (access `process.env` and `getCloudflareContext()` at top level). I hope to get to this next week. But it should not be a reason not to merge this PR. 

I have added a few questions/comments after a quick review.

packages/cloudflare/src/cli/build/patches/plugins/next-minimal.ts
conico974: Yeah, but in this case this is not actually doing any IO. Just creating an `AbortController` outside of the request context will trigger the error as described here https://github.com/cloudflare/workerd/issues/3657

It's the import of Next itself in this case that is causing trouble. I haven't found why it fails silently in this case

packages/cloudflare/src/cli/build/patches/plugins/next-minimal.spec.ts
conico974: No it is only a problem in the already minified `app-page.runtime.prod.js` or `app-page-experimental.runtime.prod.js`

pnpm-workspace.yaml
conico974: Only the e2e from aws if i'm not mistaken

packages/cloudflare/src/cli/build/bundle-server.ts
conico974: I was wrong actually, we need to define `__NEXT_EXPERIMENTAL_REACT`, otherwise it will bundle 2 versions of react and 2 versions of `app-page.runtime.prod.js` and `app-route.runtime.prod.js`. I've also checked and it should not cause any issue defining that for Next 14.2+.

On the app router e2e examples without the define, it's 800kb more unzipped, with it it's 1kb more (compared to defining `NEXT_MINIMAL`)

packages/cloudflare/src/cli/build/patches/plugins/next-minimal.ts
vicb: Thanks a lot for creating the issue!

FYI The service worker syntax is deprecated

```ts
addEventListener("fetch", event => {
  event.respondWith(new Response("Hello World"));
});
```

See https://developers.cloudflare.com/workers/reference/migrate-to-module-workers

pnpm-workspace.yaml
vicb: I missed that it was on the e2e catalog, looks fine!

packages/cloudflare/src/cli/build/bundle-server.ts
```diff
@@ -136,7 +138,9 @@ export async function bundleServer(buildOpts: BuildOptions): Promise<void> {
...
+      "process.env.TURBOPACK": "false",
```
vicb, line 141: can you add comments explaining the effect of `TURBOPACK` and `_NEXT_EXPERIMENTAL_REACT`?
Thanks!

vicb (approved):
Awesome PR, it will be well appreciated by the community 🎉

I have added a few minor comments

packages/cloudflare/src/cli/build/patches/plugins/next-minimal.ts
```diff
@@ -0,0 +1,92 @@
...
+        kind: function_declaration
+        stopBy: end
+        has:
+          kind: statement_block
+          has:
+            kind: try_statement
+            has:
+              kind: catch_clause
+              has:
+                kind: statement_block
```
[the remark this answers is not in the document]
conico974: Probably not, but as i said in the comment i just wanted to be extra careful on this one, this is a precompiled file and we want to target a `var a = new AbortController();`. This way it follows exactly the ast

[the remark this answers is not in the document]
vicb: Keeping a function declaration with the constraint on the return statement should be enough?
But ok to keep all if you think it's worth.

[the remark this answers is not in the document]
conico974: Better be safe than sorry :wink:

PR #443, opened 2025-03-09
feat: r2 adapter for the incremental cache

james-elicx:
This PR is based off of #436 to enable auto-populating of the cache entries.

**Changes**
- Adds a new R2 adapter for the incremental cache.
- Caches entries using the Cache API, and updates the Cache API in the background when an R2 fetch completes.
- Adds support for populating the R2 bucket during build.
- Enables the R2 adapter in the app-router e2es.

**Discussion points**
- We might need to duplicate the app-router tests for each adapter we add so that we can run them for multiple adapters. Or change how the e2es are run to support being run multiple times with different adapters.

---
This is **part 2 of 3 in a stack** made with GitButler:
- <kbd>&nbsp;3&nbsp;</kbd> #479 
- <kbd>&nbsp;2&nbsp;</kbd> #443 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #436


packages/cloudflare/src/api/r2-incremental-cache.ts
```diff
@@ -0,0 +1,171 @@
...
+ * The directory that the cache entries are stored in can be confused with the `NEXT_CACHE_R2_DIRECTORY`
+ * environment variable, and defaults to `incremental-cache`.
+ *
+ * The cache uses an instance of the Cache API (`incremental-cache`) to store a local version of the
+ * R2 cache entry to enable fast retrieval, with the cache being updated from R2 in the background.
+ */
+class R2IncrementalCache implements IncrementalCache {
+  readonly name = "r2-incremental-cache";
+
+  protected localCache: Cache | undefined;
```
conico974, line 27: We cannot use the `Cache API` this way without at least a **big big** warning.

The biggest issue is that the `lastModified` value will always be kind of broken, leading to some issue with ISR revalidation itself, or On-Demand revalidation. If an outdated entry is returned by the regional cache (but has already been revalidated elsewhere), Next will mark it as STALE and we will trigger a new revalidation even though it's already revalidated.

With `revalidateTag`, it's even worse because the old `lastModified` will make every entry a MISS, trigerring a render as if it were SSR.

And of course it will provide stale data at least one more time.

I think that the Cache API should be a different function altogether with the tradeoff being clear from the beginning and with several options as well to fully benefit from it. Something that could be called like that : 
```ts
  incrementalCache: withRegionalCache(r2IncrementalCache,{mode: 'enterprise-tag' | `short-lived` | 'long-lived' ,...someOptions}),
  tagCache: d1TagCache,
  queue: memoryQueue,
});
```
For the options there could be multiple one (or they could be separate implementation altogether):
- `enterprise-tag`, to use cloudflare cache tags which is an enterprise feature
- `short-lived`, to always trust if for a certain amount of time. This would act as a buffer for a short period of time and cloud make things cheaper
- `long-lived` to make it work like here
The last 2 options would require some change to the queue as well (maybe a `withRegionalCacheQueue`) as well to avoid revalidating multiple time

james-elicx: An explicit opt-in function that can wrap an incremental cache sounds like a reasonable alternative, will do that instead.

> If an outdated entry is returned by the regional cache (but has already been revalidated elsewhere), Next will mark it as STALE and we will trigger a new revalidation even though it's already revalidated.

This problem isn't unique to this implementation - that would happen with KV as well at the moment if the request happens in the other region before it's finished propagating. I know it's slightly different in that this intentionally returns from the cache and updates in the background in other regions, but nevertheless it's still a problem that exists, just not as severe.

conico974: That's true, big difference being the timeframe here.
With KV it can only happens during propagation time, with this it could happen hours or even days after.
But with a proper distributed queue, we can easily fix both this one and the KV issue as well. This only applies to ISR though, not to On-Demand revalidation where the issue i pointed out will persist

packages/cloudflare/src/api/r2-incremental-cache.ts
```diff
@@ -0,0 +1,85 @@
...
+    if (!r2) throw new IgnorableError("No R2 bucket");
+
+    debug(`Set ${key}`);
+
+    try {
+      const entry: IncrementalCacheEntry<IsFetch> = {
+        value,
+        // Note: `Date.now()` returns the time of the last IO rather than the actual time.
+        //       See https://developers.cloudflare.com/workers/reference/security-model/
+        lastModified: Date.now(),
```
conico974, line 56: Don't need that for R2 actually, we should use `uploaded` from the R2 object directly.
It would require some change in get

packages/cloudflare/src/api/regional-cache.ts
```diff
@@ -0,0 +1,133 @@
...
+      const localCacheKey = this.getCacheKey(key, isFetch);
+
+      // Check for a cached entry as this will be faster than the store response.
+      const cache = await this.getCacheInstance();
+      const cachedResponse = await cache.match(localCacheKey);
+      if (cachedResponse) {
+        debug("Get - cached response");
+
+        // Update the local cache after the R2 fetch has completed.
+        getCloudflareContext().ctx.waitUntil(
```
conico974, line 42: We likely don't want to do that in `short-lived` mode. The whole point of it in my mind is to avoid hitting the underlying cache as often.

packages/cloudflare/src/cli/build/utils/populate-cache.ts
```diff
@@ -51,7 +75,31 @@ export async function populateCache(opts: BuildOptions, config: OpenNextConfig,
...
+            "r2 object put",
+            JSON.stringify(fullDestPath),
+            `--file ${JSON.stringify(fsPath)}`,
+          ]);
+        });
+        logger.info(`Successfully populated cache with ${assets.length} assets`);
+        break;
+      }
+      default:
+        logger.info("INcremental cache does not need populating");
```
conico974, line 101:
```suggestion
        logger.info("Incremental cache does not need populating");
```

packages/cloudflare/src/api/regional-cache.ts
```diff
@@ -0,0 +1,133 @@
...
+    }
+  }
+
+  async set<IsFetch extends boolean = false>(
+    key: string,
+    value: CacheValue<IsFetch>,
+    isFetch?: IsFetch
+  ): Promise<void> {
+    try {
+      await Promise.all([
```
conico974, line 75: Shouldn't be `Promise.all` here, the underlying store is the source of truth.
If it fails we should not write to the regional cache

packages/cloudflare/src/api/regional-cache.ts
```diff
@@ -0,0 +1,133 @@
...
+      ]);
+    } catch (e) {
+      error(`Failed to get from regional cache`, e);
+    }
+  }
+
+  async delete(key: string): Promise<void> {
+    try {
+      const cache = await this.getCacheInstance();
+      await Promise.all([this.store.delete(key), cache.delete(this.getCacheKey(key))]);
```
conico974, line 92: Same as above

packages/cloudflare/src/api/regional-cache.ts
james-elicx: After thinking about it a bit more, I agree, but I think it would be better to put this behaviour behind an option instead of doing it based on the mode

packages/cloudflare/src/api/r2-incremental-cache.ts
james-elicx: I believe I should have captured everything in the recent changes. I haven't done the queue part as it sounds like you're looking at that as part of the DO stuff?

packages/cloudflare/src/api/regional-cache.ts
```diff
@@ -0,0 +1,161 @@
...
+      const cache = await this.getCacheInstance();
+      const localCacheKey = this.getCacheKey(key, isFetch);
+
+      // Check for a cached entry as this will be faster than the store response.
+      const cachedResponse = await cache.match(localCacheKey);
+      if (cachedResponse) {
+        debug("Get - cached response");
+
+        // Re-fetch from the store and update the regional cache in the background
+        if (this.opts.shouldLazilyUpdateOnCacheHit) {
```
conico974, line 51: I think by default it should be `true` for the `long-lived` and `false` for `short-lived`

conico974 (commented):
Just one little comment, i guess the only thing remaining is what to do for #436

james-elicx:
> Just one little comment, i guess the only thing remaining is what to do for #436

Yeah I think the main outstanding questions are;
- Default to local population vs explicitly opt in for population.
- Consider having multiple commands for the adapter instead where the population is based on the command.

I was hoping to be able to talk about those this week, but didn't realise that the meeting time was changing 😓. Hopefully Victor might have a chance to take a look at the PRs at some point, otherwise we can discuss next week I guess, but I was hoping to have wrapped up the population one by now.

examples/e2e/app-router/open-next.config.ts
```diff
@@ -1,10 +1,15 @@
...
 import d1TagCache from "@opennextjs/cloudflare/d1-tag-cache";
-import kvIncrementalCache from "@opennextjs/cloudflare/kv-cache";
+// import kvIncrementalCache from "@opennextjs/cloudflare/kv-cache";
+import r2IncrementalCache from "@opennextjs/cloudflare/r2-incremental-cache";
 import memoryQueue from "@opennextjs/cloudflare/memory-queue";
+import { withRegionalCache } from "@opennextjs/cloudflare/regional-cache";
 
 export default defineCloudflareConfig({
-  incrementalCache: kvIncrementalCache,
+  incrementalCache: withRegionalCache(r2IncrementalCache, {
```
vicb, line 9: Could you rather add an [other app](https://github.com/opennextjs/opennextjs-cloudflare/blob/e0ec01d50d3ae9f15294735f8fd28d84d29140ca/examples/common/apps.ts#L15)

packages/cloudflare/src/api/regional-cache.ts
```diff
@@ -0,0 +1,167 @@
...
+    );
+  }
+
+  protected async putToCache(key: Request, entry: IncrementalCacheEntry<boolean>): Promise<void> {
+    const cache = await this.getCacheInstance();
+
+    const age =
+      this.opts.mode === "short-lived"
+        ? ONE_MINUTE_IN_SECONDS
+        : entry.value.revalidate || ONE_YEAR_IN_SECONDS;
```
conico974, line 136: If i'm not mistaken `entry.value.revalidate` is only there for fetch cache entry.
We'd have to make a few changes in aws for this to work ok with ISR/SSG cache data

packages/cloudflare/src/api/regional-cache.ts
```diff
@@ -0,0 +1,167 @@
...
+
+const ONE_YEAR_IN_SECONDS = 31536000;
+const ONE_MINUTE_IN_SECONDS = 60;
+
+type Options = {
+  /**
+   * The mode to use for the regional cache.
+   *
+   * - `short-lived`: Re-use a cache entry for up to a minute after it has been retrieved.
+   * - `long-lived`: Re-use a cache entry until it is revalidated.
```
conico974, line 15: This is not correct, that's only the case for ISR and for the fetch cache (see my comment on line 136).
For now it should only be used for SSG and/or the fetch cache (with SSR)
We can just add a comment for now, we'll see how we could fix this later

examples/e2e/app-router/open-next.config.ts
conico974: My take on this : https://github.com/opennextjs/opennextjs-cloudflare/pull/470#discussion_r2003837485

The question then becomes, do we want R2 or KV to be recommended? KV because of its eventual consistency could cause some trouble for people (especially if they don't read the docs)

packages/cloudflare/src/api/regional-cache.ts
james-elicx: just looking at the types and looks like you're correct. i'll change this to fall back to 30 minutes for the moment until we change in aws

packages/cloudflare/src/api/r2-incremental-cache.ts
```diff
@@ -0,0 +1,80 @@
...
+import type { CacheValue, IncrementalCache, WithLastModified } from "@opennextjs/aws/types/overrides.js";
+import { IgnorableError } from "@opennextjs/aws/utils/error.js";
+
+import { getCloudflareContext } from "./cloudflare-context.js";
+
+/**
+ * An instance of the Incremental Cache that uses an R2 bucket (`NEXT_CACHE_R2_BUCKET`) as it's
+ * underlying data store.
+ *
+ * The directory that the cache entries are stored in can be confused with the `NEXT_CACHE_R2_PREFIX`
```
vicb, line 11:
```suggestion
 * The directory that the cache entries are stored in can be configured with the `NEXT_CACHE_R2_PREFIX`
```

packages/cloudflare/src/api/r2-incremental-cache.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    try {
+      await r2.delete(this.getR2Key(key));
+    } catch (e) {
+      error("Failed to delete from cache", e);
+    }
+  }
+
+  protected getR2Key(key: string, isFetch?: boolean): string {
+    const directory = getCloudflareContext().env.NEXT_CACHE_R2_PREFIX ?? "incremental-cache";
```
vicb, line 74: nit: could we init `this.directory` in the ctor?

packages/cloudflare/src/cli/build/utils/populate-cache.ts
```diff
@@ -37,11 +48,24 @@ function runWrangler(opts: BuildOptions, mode: CacheBindingMode, args: string[])
   if (result.status !== 0) {
     logger.error("Failed to populate cache");
     process.exit(1);
-  } else {
-    logger.info("Successfully populated cache");
   }
 }
 
+function getCacheAssetPaths(opts: BuildOptions) {
+  return globSync(path.join(opts.outputDir, "cache/**/*"), { withFileTypes: true })
```
vicb, line 55: use `windowsPathsNoEscape` (see other invocations in the repo)

vicb (approved):
LGTM with a few nits

packages/cloudflare/src/api/regional-cache.ts
```diff
@@ -0,0 +1,167 @@
...
+
+const ONE_MINUTE_IN_SECONDS = 60;
+const THIRTY_MINUTES_IN_SECONDS = ONE_MINUTE_IN_SECONDS * 30;
+
+type Options = {
+  /**
+   * The mode to use for the regional cache.
+   *
+   * - `short-lived`: Re-use a cache entry for up to a minute after it has been retrieved.
+   * - `long-lived`: Re-use a fetch cache entry until it is revalidated, or an ISR/SSG entry for up to 30 minutes.
```
conico974, line 15: We should add that it won't work as expected for On-Demand revalidation (it will only invalidate regionally)

conico974 (approved):
LGTM, just a small comment

packages/cloudflare/src/api/r2-incremental-cache.ts
james-elicx: it breaks the build as the constructor is called when compiling the opennext config
image

PR #445, opened 2025-03-10
fix: deployed worker unable to invoke itself in memory queue

james-elicx:
Service bindings are required for fetching the same worker in production.

A bit of a shame that our local e2es were unable to catch this as wrangler doesn't fail when doing fetches in local mode.

Docs PR - https://github.com/opennextjs/docs/pull/89

### DIFF
--- a/.changeset/spicy-seas-appear.md
+++ b/.changeset/spicy-seas-appear.md
@@ -3,3 +3,18 @@
 ---
 
 fix: deployed worker unable to invoke itself in memory queue
+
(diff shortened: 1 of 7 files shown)


.changeset/spicy-seas-appear.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+fix: deployed worker unable to invoke itself in memory queue
```
vicb, line 5: can you expand and add details on what users need to do (i.e. add `NEXT_CACHE_REVALIDATION_WORKER`).
Thanks!

conico974:
That's really weird, in the issue they say that it works fine with `direct` in production, which should do the exact same thing

vicb (approved):
Thanks for the fix @james-elicx, LGTM.

Maybe add a little more details to the changeset for the changelog

PR #452, opened 2025-03-12
fix: patch the webpack runtime when there is a single chunk

vicb:
---
This is **part 2 of 2 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> #452 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #451

### DIFF
--- a/packages/cloudflare/src/cli/build/patches/ast/webpack-runtime.ts
+++ b/packages/cloudflare/src/cli/build/patches/ast/webpack-runtime.ts
@@ -7,6 +7,7 @@
  *
  *  This patch unrolls the dynamic require for all the existing chunks:
  *
+ *  For multiple chunks:
  *     switch (chunkId) {
  *       case ID1: installChunk(require("./chunks/ID1"); break;
  *       case ID2: installChunk(require("./chunks/ID2"); break;
@@ -14,16 +15,20 @@
  *       case SELF_ID: installedChunks[chunkId] = 1; break;
  *       default: throw new Error(`Unknown chunk ${chunkId}`);
  *     }
+ *
+ * For a single chunk:
+ *     require("./chunks/CHUNK_ID.js");
  */
 
-import { readdirSync, readFileSync, writeFileSync } from "node:fs";
+import { existsSync, readdirSync, readFileSync, writeFileSync } from "node:fs";
 import { join } from "node:path";
 
 import { type BuildOptions, getPackagePath } from "@opennextjs/aws/build/helper.js";
 
 import { patchCode } from "./util.js";
 
-export function buildInlineChunksRule(chunks: number[]) {
+// Inline the code when there are multiple chunks
+export function buildMultipleChunksRule(chunks: number[]) {
   return `
 rule:
   pattern: ($CHUNK_ID, $_PROMISES) => { $$$ }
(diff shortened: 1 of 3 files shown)


packages/cloudflare/src/cli/build/patches/ast/webpack-runtime.ts
```diff
@@ -57,14 +82,26 @@ export async function patchWebpackRuntime(buildOpts: BuildOptions) {
...
+  if (existsSync(runtimeFile)) {
+    let code = readFileSync(runtimeFile, "utf-8");
+    code = patchCode(code, buildMultipleChunksRule(chunks));
+    code = patchCode(code, singleChunkRule);
+    writeFileSync(runtimeFile, code);
+  }
+
+  const apiRuntimeFile = join(dotNextServerDir, "webpack-api-runtime.js");
+  if (existsSync(apiRuntimeFile)) {
+    let code = readFileSync(runtimeFile, "utf-8");
```
james-elicx, line 102: is runtimeFile instead of apiRuntimeFile here intentional?

vicb: nope, good catch!

packages/cloudflare/src/cli/build/patches/ast/webpack-runtime.ts
```diff
@@ -44,8 +49,30 @@ ${chunks.map((chunk) => `         case ${chunk}: $INSTALL(require("./chunks/${ch
...
+  inside: {pattern: $_.$_.require = $$$_, stopBy: end}
+  all:
+    - has: {pattern: $INSTALL(require("./chunks/" + $$$)), stopBy: end}
+    - has: {pattern: $SELF_ID == $CHUNK_ID, stopBy: end}
+    - has: {pattern: "$INSTALLED_CHUNK[$CHUNK_ID] = 1", stopBy: end}
+fix: |
+  ($CHUNK_ID, _) => {
+    if (!$INSTALLED_CHUNK[$CHUNK_ID]) {
+      try {
+        $INSTALL(require("./chunks/$SELF_ID.js"));
```
conico974, line 67: Shouldn't this be something like that
```ts
if($CHUNK_ID === $SELF_ID) {
$INSTALL(require("./chunks/$SELF_ID.js"));
} else !$INSTALLED_CHUNK[$CHUNK_ID]
```
It probably does not matter that much but worth mentioning

packages/cloudflare/src/cli/build/patches/ast/webpack-runtime.ts
```diff
@@ -57,14 +84,28 @@ export async function patchWebpackRuntime(buildOpts: BuildOptions) {
...
+ * Inline chunks when the file exists.
+ *
+ * @param filename Path to the webpack runtime.
+ * @param chunks List of chunks in the chunks folder.
+ */
+function patchFile(filename: string, chunks: number[]) {
+  if (existsSync(filename)) {
+    let code = readFileSync(filename, "utf-8");
+    code = patchCode(code, buildMultipleChunksRule(chunks));
+    code = patchCode(code, singleChunkRule);
```
conico974, line 108: Is this one necessary for `webpack-runtime.js` as well or only for api ?

conico974 (approved):
LGTM, just 2 little questions

packages/cloudflare/src/cli/build/patches/ast/webpack-runtime.ts
vicb: I have only been able to reproduce it with the API but not sure if it can occur for `webpack-runtime.js` -> better safe than sorry ;)

vicb:
Thanks for the reviews

Issue #442, opened 2025-03-08
[BUG] Error: Dynamic require of "/.next/server/pages/api/hello.js" is not supported


jeff-polls:
### Describe the bug

500 error when pinging API routes in cloudflare preview mode, works when running `next dev`

I'm using the latest starter template with no changes:
```yarn create cloudflare my-next-app --framework=next --experimental```

I'm on a mac running Ventura, tried on node 20 and 22. 

### Steps to reproduce

```yarn create cloudflare my-next-app --framework=next --experimental```

Then run `yarn preview`

I can see the static page in the browser (although images and icons are broken), but when I try to ping the /api/hello endpoint I get a 500.

### Expected behavior

The API routes in the example project should work

### @opennextjs/cloudflare version

0.5.8

### Wrangler version

3.114.0

### next info output

```shell
✘ [ERROR] ⨯ Error: Dynamic require of "/.next/server/pages/api/hello.js" is not supported
```

vicb:
Could you please add a repro (link to a GH repo) and add the missing infos from the bug template.

jeff-polls:
@vicb thanks for the response. 

What repo? Did you see my note that the issue is happening using the start command from the docs without modifying any of the code?

yarn create cloudflare my-next-app --framework=next --experimental

Did you try running that command and then running yarn preview and trying to ping the API route?

vicb:
We will not spend time trying to reproduce if we don't know what platform, version yarn, ... you use.

So please add a link to the failing app and provide the info (`next info`) requested.

Thanks

jeff-polls:
hey @vicb, I didn't realize that I accidentally put my "Additional context" in the `next info` section and skipped over the `next info` section, my mistake.

It looks like the root cause of the issue is that your `yarn create cloudflare` command installs the wrong version of NextJS. When I ran the `next info` command it recommended upgrading and that resolved the issue. So I would recommend that you update the `yarn create cloudflare` command to avoid others running into this issue.

Again, my mistake for not seeing that in the template or I would have caught this earlier, but still think the issue should be avoided altogether and `yarn create cloudflare` should install the right version.

By the way, you keep asking for my to link the failing app, so I want to make sure you understand that the issue is with the default app that is created by `yarn create cloudflare`, with no changes. Does that make sense? For my own understanding, even though the issue is with the default app that the command creates, is there some reason you still needed me to push a repo and link it? Doesn't matter now, but just curious.

[...]

PR #453, opened 2025-03-12
fix: Pages router API routes with Next 14

vicb:
fixes https://github.com/opennextjs/opennextjs-cloudflare/issues/442

While the patch is the same, the path to the patched file is different in Next14 (`next/dist/server/future/helpers/module-loader/node-module-loader.js`). Update the patch to cover both Next14 (add test) and Next 15 (already tested)

### DIFF
--- a/packages/cloudflare/src/cli/build/patches/plugins/dynamic-requires.ts
+++ b/packages/cloudflare/src/cli/build/patches/plugins/dynamic-requires.ts
@@ -47,10 +47,9 @@ export function inlineDynamicRequires(updater: ContentUpdater, buildOpts: BuildO
   updater.updateContent(
     "inline-node-module-loader",
     {
-      filter: getCrossPlatformPathRegex(
-        String.raw`/next/dist/server/lib/module-loader/node-module-loader\.js$`,
-        { escape: false }
-      ),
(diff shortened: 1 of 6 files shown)


james-elicx (approved):

packages/cloudflare/src/cli/build/patches/plugins/dynamic-requires.ts
```diff
@@ -47,10 +47,9 @@ export function inlineDynamicRequires(updater: ContentUpdater, buildOpts: BuildO
   updater.updateContent(
     "inline-node-module-loader",
     {
-      filter: getCrossPlatformPathRegex(
-        String.raw`/next/dist/server/lib/module-loader/node-module-loader\.js$`,
-        { escape: false }
-      ),
+      filter: getCrossPlatformPathRegex(String.raw`/module-loader/node-module-loader\.js$`, {
```
conico974, line 50: That's exactly the type of use case for which i wanted the `VersionedField` in the Generic code patching PR to be on every field

vicb:
Thanks James (I had to rebase because it has a dep on my previous PR)

conico974 (approved):
LGTM, you did not add a changeset though

PR #455, opened 2025-03-12
sync with aws@3.5.2

vicb:


packages/cloudflare/package.json
```diff
@@ -73,7 +73,7 @@
   "dependencies": {
     "@ast-grep/napi": "^0.36.1",
     "@dotenvx/dotenvx": "catalog:",
-    "@opennextjs/aws": "https://pkg.pr.new/@opennextjs/aws@756",
+    "@opennextjs/aws": "https://pkg.pr.new/@opennextjs/aws@7e23eee",
```
vicb, line 76: Note: using the SHA instead of 3.5.2 to minimize diff with the next update.

conico974 (approved):
LGTM.
I was wondering why it failed, with the route it works better :smile:

vicb:
> LGTM. I was wondering why it failed, with the route it works better 😄

Yep, I was afraid it would be a long and painful debugging session but I was pleasantly surprised for once 🎉

Thanks for the review

PR #464, opened 2025-03-14
D1 next mode tag cache

conico974:


examples/e2e/app-router/open-next.config.ts
```diff
@@ -1,10 +1,11 @@
 import { defineCloudflareConfig } from "@opennextjs/cloudflare";
 import d1TagCache from "@opennextjs/cloudflare/d1-tag-cache";
+import d1NextModeTagCache from "@opennextjs/cloudflare/d1-next-tag-cache";
 import kvIncrementalCache from "@opennextjs/cloudflare/kv-cache";
 import memoryQueue from "@opennextjs/cloudflare/memory-queue";
 
 export default defineCloudflareConfig({
   incrementalCache: kvIncrementalCache,
-  tagCache: d1TagCache,
+  tagCache: d1NextModeTagCache,
```
vicb, line 9: Is this change expected (as you're also adding `d1-tag-next`)?

packages/cloudflare/src/api/d1-next-tag-cache.ts
```diff
@@ -0,0 +1,81 @@
+import { debug, error } from "@opennextjs/aws/adapters/logger.js";
+import type { OpenNextConfig } from "@opennextjs/aws/types/open-next.js";
+import type { NextModeTagCache } from "@opennextjs/aws/types/overrides.js";
+import { RecoverableError } from "@opennextjs/aws/utils/error.js";
+
+import { getCloudflareContext } from "./cloudflare-context.js";
+
+export class D1NextModeTagCache implements NextModeTagCache {
+  mode = "nextMode" as const;
+  name = "d1-next-mode-tag-cache";
```
vicb, line 10:
```suggestion
  readonly mode = "nextMode" as const;
  readonly name = "d1-next-mode-tag-cache";
```

packages/cloudflare/src/api/d1-next-tag-cache.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  async hasBeenRevalidated(tags: string[], lastModified?: number): Promise<boolean> {
+    const { isDisabled, db, tables } = this.getConfig();
+    if (isDisabled) return false;
+    try {
+      const result = await db
+        .prepare(
+          `SELECT COUNT(*) as cnt FROM ${JSON.stringify(tables.revalidations)} WHERE tag IN (${tags.map(() => "?").join(", ")}) AND revalidatedAt > ?`
```
vicb, line 18: add `LIMIT 1` to speed up the query?

vicb: Should there be an index on `revalidatedAt`?

packages/cloudflare/src/api/d1-next-tag-cache.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    if (!db || isDisabled) {
+      return { isDisabled: true as const };
+    }
+
+    return {
+      isDisabled: false as const,
+      db,
+      tables: {
+        revalidations: cfEnv.NEXT_CACHE_D1_REVALIDATIONS_TABLE ?? "revalidations",
```
vicb, line 63: extract `reavalidations` to a shared constant ?

vicb (approved):
LGTM, a few nits

examples/e2e/app-router/open-next.config.ts
conico974: Oops, i forgot to remove it.

packages/cloudflare/src/api/d1-next-tag-cache.ts
conico974: Not sure if we really need an index here for `revalidatedAt`, the tag is already the primary key.
I don't think the performance gain justify the cost from the size of the index. 
What do you think ?

vicb: Yep, we can check later if this proves to be a bottleneck

PR #469, opened 2025-03-17
feat: support passing options to init open-next cloudflare dev function

Juuldamen:
To allow same behaviour as `setupDevPlatform()` from next-on-pages. In this case I assume the removal of passing options was not intended. If not, are there any alternatives to the below?

```ts
import { initOpenNextCloudflareForDev } from "@opennextjs/cloudflare";

initOpenNextCloudflareForDev({
  persist: {
    path: "../../.wrangler/state/v3/custom-dir",
  },
});
```

When testing with ` pnpm --filter playground15 preview` the `.wrangler` directory is correctly created in the root of the project. But a `.wrangler` directory also appears within the playground15 directory. So I'm creating this PR so I can use a preview deployment to be able to test in my own app with D1. Or are there better ways to test in this repo itself?

image

### DIFF
--- a/.changeset/shaggy-bobcats-battle.md
+++ b/.changeset/shaggy-bobcats-battle.md
@@ -14,4 +14,4 @@ initOpenNextCloudflareForDev({
 });
 ```
 
-You can find the type with the available options [here](https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/api/integrations/platform/index.ts#L32) in the Wrangler source code. Please note that the `environment` field is not customizable as it's used internally by OpenNext.js. If you have a use case where overriding `environment` is necessary, please let us know by opening a [feature request](https://github.com/cloudflare/workers-sdk/issues).
+You can find the configuration type with it's available options [here](https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/api/integrations/platform/index.ts#L32) in the Wrangler source code.
--- a/.changeset/shaggy-bobcats-battle.md
+++ b/.changeset/shaggy-bobcats-battle.md
@@ -1,0 +1,15 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+Adds support for passing options to initOpenNextCloudflareForDev().
+
+For example:
+
+```ts
+initOpenNextCloudflareForDev({
+  persist: {
+    path: "../../.wrangler/state/v3/custom-dir",
+  },
+});
+```
--- a/.changeset/shaggy-bobcats-battle.md
+++ b/.changeset/shaggy-bobcats-battle.md
@@ -2,9 +2,9 @@
 "@opennextjs/cloudflare": patch
 ---
 
-Adds support for passing options to initOpenNextCloudflareForDev().
+Adds support for passing options to `initOpenNextCloudflareForDev()`. This allows you to configure how your Cloudflare bindings will behave during [local development](https://opennext.js.org/cloudflare/get-started#11-develop-locally).
 
-For example:
+For example, the below configuration will persist the local state of bindings to a custom directory. Which can be useful if you want to share the state between different apps that reuse the same bindings in a monorepo.
 
 ```ts
 initOpenNextCloudflareForDev({
@@ -13,3 +13,5 @@ initOpenNextCloudflareForDev({
   },
 });
 ```
+
+You can find the type with the available options [here](https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/api/integrations/platform/index.ts#L32) in the Wrangler source code. Please note that the `environment` field is not customizable as it's used internally by OpenNext.js. If you have a use case where overriding `environment` is necessary, please let us know by opening a [feature request](https://github.com/cloudflare/workers-sdk/issues).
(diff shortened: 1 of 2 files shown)


packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -250,12 +253,13 @@ async function monkeyPatchVmModuleEdgeContext(cloudflareContext: CloudflareConte
...
   CfProperties extends Record<string, unknown> = IncomingRequestCfProperties,
   Context = ExecutionContext,
->(): Promise<CloudflareContext<CfProperties, Context>> {
+>(options?: Omit<GetPlatformProxyOptions, "environment">): Promise<CloudflareContext<CfProperties, Context>> {
   // Note: we never want wrangler to be bundled in the Next.js app, that's why the import below looks like it does
   const { getPlatformProxy } = await import(/* webpackIgnore: true */ `${"__wrangler".replaceAll("_", "")}`);
   const { env, cf, ctx } = await getPlatformProxy({
     // This allows the selection of a wrangler environment while running in next dev mode
     environment: process.env.NEXT_DEV_WRANGLER_ENV,
+    ...options,
```
vicb, line 262: Change the order (to protect against javascript or cast)
```suggestion
    ...options,
    environment: process.env.NEXT_DEV_WRANGLER_ENV,   
```

vicb:
@dario-piotrowicz or @james-elicx do you have history/insights from next on pages?

james-elicx:
>  do you have history/insights from next on pages?

Nothing too much, it was just that some people wanted the ability to further configure what wrangler was going to do, e.g. where data was persisted like in the example above.

packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -166,12 +168,13 @@ async function getCloudflareContextAsync<
  * with the open-next Cloudflare adapter
  *
  * Note: this function should only be called inside the Next.js config file, and although async it doesn't need to be `await`ed
+ * @param options options how the function should operate and if/where to persist the platform data
  */
-export async function initOpenNextCloudflareForDev() {
+export async function initOpenNextCloudflareForDev(options?: Omit<GetPlatformProxyOptions, "environment">) {
```
dario-piotrowicz, line 173: Why omitting `environment`?

packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -166,12 +168,13 @@ async function getCloudflareContextAsync<
  * with the open-next Cloudflare adapter
  *
  * Note: this function should only be called inside the Next.js config file, and although async it doesn't need to be `await`ed
+ * @param options options how the function should operate and if/where to persist the platform data
```
dario-piotrowicz, line 171:
```suggestion
 *
 * @param options options on how the function should operate and if/where to persist the platform data
```

.changeset/shaggy-bobcats-battle.md
```diff
@@ -0,0 +1,15 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+Adds support for passing options to initOpenNextCloudflareForDev().
```
dario-piotrowicz, line 5:
```suggestion
Add support for passing options to `initOpenNextCloudflareForDev()`
```

Below you could also briefly explain what this does and what it is for 🙂

packages/cloudflare/src/api/cloudflare-context.ts
Juuldamen: Currently in `getCloudflareContextFromWrangler` the environment config is not taken into account because that function already handles it. So I wanted to reflect that in the type so people will not get confused here. If you think the environment should also be configurable I will change it.

```ts
  // getCloudflareContextFromWrangler()
  const { env, cf, ctx } = await getPlatformProxy({
    // This allows the selection of a wrangler environment while running in next dev mode
    ...options,
    environment: process.env.NEXT_DEV_WRANGLER_ENV,
  });
```

packages/cloudflare/src/api/cloudflare-context.ts
Juuldamen: Changed

packages/cloudflare/src/api/cloudflare-context.ts
dario-piotrowicz: Ah I see 🤔 

I think it might still be good to allow the user to provide an environment for the platform proxy though 🤔 

Could we remove this limitation? and instead have some code such as:
```ts
const environment = options.environment ?? process.env.NEXT_DEV_WRANGLER_ENV;

if (options.environment && process.env.NEXT_DEV_WRANGLER_ENV) {
    // console.warn saying that NEXT_DEV_WRANGLER_ENV is set but it's being overridden by the environment option
}

// getCloudflareContextFromWrangler()
const { env, cf, ctx } = await getPlatformProxy({
  ...options,
  environment,
 });
```
?

In this way `NEXT_DEV_WRANGLER` still works but developers do have the opportunity to use the `environment` option for `initOpenNextCloudflareForDev`

what do you think? 🙂

dario-piotrowicz (commented):
Thanks for the ping @Juuldamen 🙏 , sorry I didn't realize this PR was waiting for my review

I'll keep an eye on if from now on 🙂 

(I just left one last small comment)

.changeset/shaggy-bobcats-battle.md
dario-piotrowicz: Ah ok thanks, this looks good 🙂

packages/cloudflare/src/api/cloudflare-context.ts
Juuldamen: I agree. I've pushed the change

PR #460, opened 2025-03-13
Durable Queue Storage

conico974:
It is a follow up of #459
It adds storage to the Durable object allowing it to better handle failure
It also allows it to handle eventually consistent incremental cache implementation by checking for every route if the route has been revalidated (Note that this does not cover On-Demand revalidation )

---
This is **part 2 of 2 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> #460 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #459

PR #459, opened 2025-03-13
Durable Queue

conico974:
This PR implements a Durable Object for the Queue allowing proper deduplication of the ISR request.
I'll add handling of eventually consistent Incremental Cache (such as KV or using `withRegionalCache` from James PR) in another PR

The PR is split into 2 parts, this one and the one in #460 that add storage to it

---
This is **part 1 of 2 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> #460 
- <kbd>&nbsp;1&nbsp;</kbd> #459 👈

PR #470, opened 2025-03-17
Sharded tag cache

conico974:
This PR is another implementation of the `nextMode` for the Tag Cache.
It uses Durable Object with SQLite storage to store the Tag revalidations. It allows to apply sharding and use multiple SQLite DB at the same time for better scalability.
It also includes an optional cache layer to further improve performance and reduce the load on the durable objects.

It depends on both #459 and #460 (For some reason GitButler decided to give up on this, i had to do it manually)


examples/e2e/app-router/open-next.config.ts
```diff
@@ -2,9 +2,10 @@ import { defineCloudflareConfig } from "@opennextjs/cloudflare";
 import d1TagCache from "@opennextjs/cloudflare/d1-tag-cache";
 import kvIncrementalCache from "@opennextjs/cloudflare/kv-cache";
 import doQueue from "@opennextjs/cloudflare/durable-queue";
+import shardedTagCache from "@opennextjs/cloudflare/do-sharded-tag-cache";
 
 export default defineCloudflareConfig({
   incrementalCache: kvIncrementalCache,
-  tagCache: d1TagCache,
+  tagCache: shardedTagCache({ numberOfShards: 12, regionalCache: true }),
```
vicb, line 9: Do we want to modify this or create a new app instead?

packages/cloudflare/src/api/do-sharded-tag-cache.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+    });
+  });
+
+  describe("writeTags", () => {
+    beforeEach(() => {
+      globalThis.openNextConfig = {
+        dangerous: { disableTagCache: false },
+      };
+    });
+    it("should return if the cache is disabled", async () => {
```
vicb, line 134: "return" -> "return early"/"be a no op"/'not update"

packages/cloudflare/src/api/do-sharded-tag-cache.ts
```diff
@@ -0,0 +1,193 @@
...
+import type { NextModeTagCache } from "@opennextjs/aws/types/overrides.js";
+import { IgnorableError } from "@opennextjs/aws/utils/error.js";
+
+import { getCloudflareContext } from "./cloudflare-context";
+
+interface ShardedD1TagCacheOptions {
+  /**
+   * The number of shards that will be used.
+   * 1 shards means 1 durable object instance.
+   * Be aware that the more shards you have, the more requests you will make to the Durable Object
```
vicb, line 13: Can we reword to say that every request is send to all the shards or that the number of requests is linear with the number of shards?

packages/cloudflare/src/api/do-sharded-tag-cache.ts
```diff
@@ -0,0 +1,193 @@
...
+   * Because of the way tags are implemented in Next.js, some shards will have more requests than others. For these cases, it is recommended to enable the regional cache.
+   * @default false
+   */
+  regionalCache?: boolean;
+  /**
+   * The TTL for the regional cache in seconds
+   * Increasing this value will reduce the number of requests to the Durable Object, but it could make `revalidateTags`/`revalidatePath` call being longer to take effect
+   * @default 5
+   */
+  regionalCacheTtl?: number;
```
vicb, line 28: `regionalCacheTtlSec`?

packages/cloudflare/src/api/do-sharded-tag-cache.ts
```diff
@@ -0,0 +1,193 @@
...
+}
+class ShardedD1TagCache implements NextModeTagCache {
+  mode = "nextMode" as const;
+  public readonly name = "sharded-d1-tag-cache";
```
vicb, line 32:
```suggestion
  readonly mode = "nextMode" as const;
  readonly name = "sharded-d1-tag-cache";
```

examples/e2e/app-router/open-next.config.ts
conico974: I forgot to write something about this here.
My idea here is that what we consider as the recommended option (or the most "production ready"), we test it against the whole app-router (or the one we want/need) test suite. For the other one we just test them on smaller apps that focus only on the part we need to test.

packages/cloudflare/src/api/do-sharded-tag-cache.ts
conico974: Rephrased it to `The number of requests made to Durable Objects will scale linearly with the number of shards. For example, a request involving 5 tags may access between 1 and 5 shards, with the upper limit being the lesser of the number of tags or the number of shards`

vicb (approved):

PR #480, opened 2025-03-20
Fix memory queue with deduplication and add basic test app

conico974:
Memory queue was using `MessageGroupId` instead of `MessageDeduplicationId` for deduplicating
Use signal for timeout in fetch
Change logger to use the correct one (we should probably change that in aws , one is for build time and one is for runtime)
Add a log in case it fails without crashing


examples/overrides/memory-queue/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "react": "^19.0.0",
+    "react-dom": "^19.0.0",
+    "next": "catalog:e2e"
+  },
+  "devDependencies": {
+    "@opennextjs/cloudflare": "workspace:*",
+    "@playwright/test": "catalog:",
+    "@types/node": "catalog:",
+    "@types/react": "^19",
+    "@types/react-dom": "^19",
```
vicb, line 24:
```suggestion
    "@types/react": "catalog:e2e",
    "@types/react-dom": "catalog:e2e",
```

vicb (approved):
LGTM
- few nits
- could you please add a changeset

PR #481, opened 2025-03-20
fix res.revalidate

### LINKED ISSUE #463: [BUG] Fix `res.revalidate`
In pages router, `res.revalidate()` doesn't work. It throws an exception.
It probably needs to be patched to use the `NEXT_CACHE_REVALIDATION_WORKER` binding instead of calling itself

conico974:
Fixes #463


packages/cloudflare/src/cli/build/patches/plugins/res-revalidate.ts
```diff
@@ -0,0 +1,78 @@
+import { patchCode } from "@opennextjs/aws/build/patch/astCodePatcher.js";
```
vicb, line 1: Can you add a file level comment describing what the patch does (i.e. use `NEXT_CACHE_REVALIDATION_WORKER` to fetch).

vicb (approved):
Thanks.

LGTM.  

Could you please add a file level comment for the patch?

vicb:
Thanks for the update!

PR #482, opened 2025-03-20
Sync with latest `@opennextjs/aws`

vicb:
It looks like the new test ("Incremental Static Regeneration with data cache") is failing and this should be fixed before merging but there is already some changes to review...

@conico974 any idea what could go wrong?


conico974:
My guess is that the cache is not properly initialized, i'll need to check that tomorrow

conico974:
@vicb I think i've figured out what's going on here. The issue happen for `unstable_cache`, and i think it comes down to how the key are computed. It tries to fetch a different key than the one generated at build time.

Probably all coming down from this https://github.com/vercel/next.js/blob/9f68435d39665388995c984887562b6974c2ff98/packages/next/src/server/web/spec-extension/unstable-cache.ts#L88-L90 
`cb.toString()` is likely different from the one generated at build time.

For fixing the e2e we could just make it revalidates at least once before actually running what we want from the test

vicb:
Thanks for the investigation Nico!

vicb:
@conico974 would you have time to push a fix to this branch?

conico974:
@vicb Not right away, but i can do it later today

vicb:
> @vicb Not right away, but i can do it later today

No tush, Monday morning would be great, enjoy your weekend!

vicb:
Thanks Nico!
Rebased + use aws@3.5.3

conico974 (approved):
LGTM Thanks

vicb:
Thanks for the review Nico!

PR #484, opened 2025-03-21
Feat Sharding Replication

conico974:
Implements sharding replication for the Durable Object tag cache.
Instead of being sharded only based on the tag, each shard will now also be replicated
This allow to scale way more (virtually infinitely)
A single tag could be read from any of the N random replicas associated with the tag shard.
On write we still need to write to every single replicas associated with this tag shard

### DIFF
--- a/packages/cloudflare/src/api/cloudflare-context.ts
+++ b/packages/cloudflare/src/api/cloudflare-context.ts
@@ -24,7 +24,7 @@ declare global {
     // Durables object namespace to use for the sharded tag cache
     NEXT_CACHE_D1_SHARDED?: DurableObjectNamespace<DOShardedTagCache>;
     // Dead letter queue for the D1 sharded tag cache
-    NEXT_CACHE_D1_SHARDED_DLQ?: Queue
+    NEXT_CACHE_D1_SHARDED_DLQ?: Queue;
 
     // Asset binding
     ASSETS?: Fetcher;
--- a/packages/cloudflare/src/api/cloudflare-context.ts
+++ b/packages/cloudflare/src/api/cloudflare-context.ts
@@ -23,6 +23,8 @@ declare global {
     NEXT_CACHE_REVALIDATION_DURABLE_OBJECT?: DurableObjectNamespace<DurableObjectQueueHandler>;
     // Durables object namespace to use for the sharded tag cache
     NEXT_CACHE_D1_SHARDED?: DurableObjectNamespace<DOShardedTagCache>;
+    // Dead letter queue for the D1 sharded tag cache
+    NEXT_CACHE_D1_SHARDED_DLQ?: Queue
 
     // Asset binding
     ASSETS?: Fetcher;
--- a/packages/cloudflare/src/api/cloudflare-context.ts
+++ b/packages/cloudflare/src/api/cloudflare-context.ts
@@ -22,9 +22,11 @@ declare global {
     // Durable Object namespace to use for the durable object queue handler
     NEXT_CACHE_REVALIDATION_DURABLE_OBJECT?: DurableObjectNamespace<DurableObjectQueueHandler>;
     // Durables object namespace to use for the sharded tag cache
-    NEXT_CACHE_D1_SHARDED?: DurableObjectNamespace<DOShardedTagCache>;
-    // Dead letter queue for the D1 sharded tag cache
-    NEXT_CACHE_D1_SHARDED_DLQ?: Queue;
+    NEXT_CACHE_DO_SHARDED?: DurableObjectNamespace<DOShardedTagCache>;
+    // Queue of failed tag write
+    // It could be used for monitoring or to reprocess failed writes
+    // Entirely optional
+    NEXT_CACHE_DO_SHARDED_DLQ?: Queue;
 
     // Asset binding
     ASSETS?: Fetcher;
--- a/packages/cloudflare/src/api/do-sharded-tag-cache.ts
+++ b/packages/cloudflare/src/api/do-sharded-tag-cache.ts
@@ -7,11 +7,14 @@ import { IgnorableError } from "@opennextjs/aws/utils/error.js";
 import { getCloudflareContext } from "./cloudflare-context";
 
 const SOFT_TAG_PREFIX = "_N_T_/";
+export const DEFAULT_MAX_SOFT_SHARDS = 4;
+export const DEFAULT_MAX_HARD_SHARDS = 2;
 
 interface ShardedD1TagCacheOptions {
   /**
    * The number of shards that will be used.
    * 1 shards means 1 durable object instance.
+   * Soft (internal next tags used for `revalidatePath`) and hard tags (the one you define in your app) will be split in different shards.
    * The number of requests made to Durable Objects will scale linearly with the number of shards.
    * For example, a request involving 5 tags may access between 1 and 5 shards, with the upper limit being the lesser of the number of tags or the number of shards
    * @default 4
@@ -29,13 +32,38 @@ interface ShardedD1TagCacheOptions {
    * @default 5
    */
   regionalCacheTtlSec?: number;
+
+  /**
+   * Whether to enable double sharding
+   * Double sharding will split each shards into N shards to spread the load even more
+   * For example with N being 2, tag `tag1` could be send to shard `shard-hard-1-1` and `shard-hard-1-2`
+   * The first part will always stay the same for a given tag, the second part will be randomly generated between 1 and N
+   * On read you only need to read from one of the shards, but on write you need to write to all shards
+   * @default false
+   */
+  enableDoubleSharding?: boolean;
+
+  /**
+   * The number of shards that will be used for double sharding
+   * Soft shards are more often accessed than hard shards, so it is recommended to have more soft shards than hard shards
+   * @default { softShards: 4, hardShards: 2 }
+   */
+  doubleShardingOpts?: {
+    softShards: number;
+    hardShards: number;
+  };
 }
 class ShardedD1TagCache implements NextModeTagCache {
   readonly mode = "nextMode" as const;
   readonly name = "sharded-d1-tag-cache";
+  readonly maxSoftShards: number;
+  readonly maxHardShards: number;
   localCache?: Cache;
 
-  constructor(private opts: ShardedD1TagCacheOptions = { numberOfShards: 4 }) {}
+  constructor(private opts: ShardedD1TagCacheOptions = { numberOfShards: 4 }) {
+    this.maxSoftShards = opts.doubleShardingOpts?.softShards ?? DEFAULT_MAX_SOFT_SHARDS;
+    this.maxHardShards = opts.doubleShardingOpts?.hardShards ?? DEFAULT_MAX_HARD_SHARDS;
+  }
 
   private getDurableObjectStub(shardId: string) {
     const durableObject = getCloudflareContext().env.NEXT_CACHE_D1_SHARDED;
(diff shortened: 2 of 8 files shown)


packages/cloudflare/src/api/do-sharded-tag-cache.ts
```diff
@@ -27,13 +33,46 @@ interface ShardedD1TagCacheOptions {
...
+
+  /**
+   * The number of shards that will be used for double sharding
+   * Soft shards are more often accessed than hard shards, so it is recommended to have more soft shards than hard shards
+   * @default { softShards: 4, hardShards: 2 }
+   */
+  doubleShardingOpts?: {
+    softShards: number;
+    hardShards: number;
+  };
```
vicb, line 55: `numberOfSoftShards` to be consistent with l23

I wonder if "2 level sharding" would be clearer than "double sharding" across this PR?

packages/cloudflare/src/api/do-sharded-tag-cache.ts
```diff
@@ -43,17 +82,56 @@ class ShardedD1TagCache implements NextModeTagCache {
...
+  }
+
+  /**
+   * This function generates an array for the double sharding
+   * @param tags The tags to generate shards for
+   * @param shardType Whether to generate shards for soft or hard tags
+   * @param generateAllShards Whether to generate all shards or only one
+   * @returns An array of shardId and tag
+   */
+  private generateShardArray(tags: string[], shardType: "soft" | "hard", generateAllShards = false) {
```
vicb, line 96: nit: `{generateAllShards = false}` would be nicer than a bool (everywhere applicable)

packages/cloudflare/src/api/do-sharded-tag-cache.ts
```diff
@@ -130,18 +208,39 @@ class ShardedD1TagCache implements NextModeTagCache {
...
-        await stub.writeTags(shardedTags);
-        // Depending on the shards and the tags, deleting from the regional cache will not work for every tag
-        await this.deleteRegionalCache(shardId, shardedTags);
+        await this.performWriteTagsWithRetry(shardId, shardedTags, currentTime);
       })
     );
   }
 
+  async performWriteTagsWithRetry(shardId: string, tags: string[], lastModified: number, retryNumber = 0) {
+    if (retryNumber >= this.maxWriteRetries) {
```
vicb, line 222: nit: this could be moved l240 to be closer to the related logic (it would also avoid an extra-call but not the main reason)

packages/cloudflare/src/api/do-sharded-tag-cache.ts
conico974: Yeah i wasn't so sure about the name either.
I also had the idea of vertical sharding as opposed to the first level which would be horizontal. Not sure if it's really that much clearer
What do you think ?

vicb (approved):
Nice!
A few nits related to naming

packages/cloudflare/src/api/do-sharded-tag-cache.ts
vicb: @james-elicx what do you think as a native speaker?

conico974: For context, there is one additional level of sharding that we could add and that would be `regional sharding` (this could improve latencies a lot). Idea would be to use either `colo`,`country` or `continent` from [the request cf properties](https://developers.cloudflare.com/workers/runtime-apis/request/#incomingrequestcfproperties) to add to the durable object name

james-elicx: Shouldn't it be `maxSoftShards`?

james-elicx: Would it be fair to describe this as shard replication? From what you're saying, it sounds like a replication layer for each shard

conico974: Yeah that's probably better, so this one would be `shardReplicationOptions` and i would then add `enableRegionalReplication` as an option inside it in a future PR.
And instead of `numberOfSoftShards` we'd have `numberOfSoftReplicas`
@vicb What do you think about that ?

vicb: I really like shard + replica!

Could we also make the id something like: "tag:soft,shard:1,replica:3"?

conico974: Yeah, i'll make the change tomorrow

packages/cloudflare/src/api/do-sharded-tag-cache.ts
```diff
@@ -43,17 +82,64 @@ class ShardedD1TagCache implements NextModeTagCache {
...
+  }
+
+  /**
+   * This function generates an array for the double sharding
+   * @param tags The tags to generate shards for
+   * @param shardType Whether to generate shards for soft or hard tags
+   * @param generateAllShards Whether to generate all shards or only one
+   * @returns An array of shardId and tag
+   */
+  private generateShardArray({
```
vicb, line 96: IIUC it is not really a "Shard" but a DO ID?

generateDOIdArray?

packages/cloudflare/src/api/do-sharded-tag-cache.ts
```diff
@@ -43,17 +82,64 @@ class ShardedD1TagCache implements NextModeTagCache {
...
+
+    const hardTags = this.generateShardArray({ tags, shardType: "hard", generateAllShards });
     // For each tag, we generate a message group id
-    const messageGroupIds = tags.map((tag) => ({
-      shardId: generateShardId(tag, this.opts.numberOfShards, "shard"),
-      tag,
-    }));
+    const messageGroupIds = [...softTags, ...hardTags];
     // We group the tags by shard
     const shards = new Map<string, string[]>();
```
vicb, line 144:
```suggestion
    const tagsByDOId = new Map<string, string[]>();
```

packages/cloudflare/src/api/do-sharded-tag-cache.ts
```diff
@@ -151,8 +258,13 @@ class ShardedD1TagCache implements NextModeTagCache {
   }
 
   async getCacheKey(shardId: string, tags: string[]) {
+    // We need to remove the last random part from the shardId
+    const shardIdWithoutRandom = shardId.replace(/-\d+$/, "");
```
vicb, line 262: Should "shardId" (/DO id) be a class so that we can extend/unit test more easily ?

packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -23,6 +23,8 @@ declare global {
     NEXT_CACHE_REVALIDATION_DURABLE_OBJECT?: DurableObjectNamespace<DurableObjectQueueHandler>;
     // Durables object namespace to use for the sharded tag cache
     NEXT_CACHE_D1_SHARDED?: DurableObjectNamespace<DOShardedTagCache>;
+    // Dead letter queue for the D1 sharded tag cache
```
vicb, line 26: Is this really a DLQ or a queue of failed write instead? 
Maybe describe what's it used for and that it's optional?

packages/cloudflare/src/api/do-sharded-tag-cache.ts
conico974: Yeah good idea

packages/cloudflare/src/api/cloudflare-context.ts
conico974: Yeah it's more a queue of failed write.
It's mostly for monitoring, if you get a lot of error here, it probably means that you'll need to increase either the shard count or replicas number (Or something really wrong is going on)

packages/cloudflare/src/api/do-sharded-tag-cache.ts
conico974: @vicb I've created a simple class here for now (Not sure about the name though), let me know what you think ?

packages/cloudflare/src/api/do-sharded-tag-cache.ts
```diff
@@ -6,10 +6,16 @@ import { IgnorableError } from "@opennextjs/aws/utils/error.js";
...
-interface ShardedD1TagCacheOptions {
+const SOFT_TAG_PREFIX = "_N_T_/";
+export const DEFAULT_MAX_SOFT_SHARDS = 4;
+export const DEFAULT_MAX_HARD_SHARDS = 2;
+export const DEFAULT_MAX_WRITE_RETRIES = 3;
+
+interface ShardedDOTagCacheOptions {
   /**
    * The number of shards that will be used.
    * 1 shards means 1 durable object instance.
```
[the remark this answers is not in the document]
vicb: Would "size" / "baseSize" work?

packages/cloudflare/src/api/do-sharded-tag-cache.ts
```diff
@@ -27,46 +34,159 @@ interface ShardedD1TagCacheOptions {
    * @default 5
    */
   regionalCacheTtlSec?: number;
+
+  /**
+   * Whether to enable shard replication
+   * Shard replication will duplicate each shards into N replicas to spread the load even more
+   * For example with N being 2, tag `tag1` could be read from 2 different durable object instance
+   * On read you only need to read from one of the shards, but on write you need to write to all shards
```
vicb, line 42: """
All replicas of the a shard contain the same value - write are sent to all of the replicas.
This allows most frequent read operations to be sent to only one replica to spread the load.
"""

packages/cloudflare/src/api/do-sharded-tag-cache.ts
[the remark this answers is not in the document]
conico974: I guess, we should probably keep shard in it though. Something like `baseShardSize` would probably work

packages/cloudflare/src/api/do-sharded-tag-cache.ts
```diff
@@ -27,46 +34,159 @@ interface ShardedD1TagCacheOptions {
...
-class ShardedD1TagCache implements NextModeTagCache {
+
+interface TagCacheDOIdOptions {
+  baseShardId: string;
+  numberOfReplicas: number;
+  shardType: "soft" | "hard";
+  replicaId?: number;
+}
+export class TagCacheDOId {
+  baseShardId: string;
```
vicb, line 72: Should we call "base-soft/hard" the shard?

```suggestion
  shardId: string;
```

packages/cloudflare/src/api/do-sharded-tag-cache.ts
```diff
@@ -27,46 +34,159 @@ interface ShardedD1TagCacheOptions {
...
+    let doubleShardArray: Array<number | undefined> = [1];
+    const isSoft = shardType === "soft";
+    const numReplicas = isSoft ? this.numSoftReplicas : this.numHardReplicas;
+    if (this.opts.enableShardReplication) {
+      doubleShardArray = generateAllShards
+        ? Array.from({ length: numReplicas }, (_, i) => i + 1)
+        : [undefined];
+    }
+    return doubleShardArray
+      .map((replicaId) => {
```
vicb, line 136: I think `.map(...).flat()` is `.flatMap()`?

packages/cloudflare/src/api/do-sharded-tag-cache.ts
```diff
@@ -27,46 +34,159 @@ interface ShardedD1TagCacheOptions {
...
+      .flat();
+  }
+
   /**
    * Same tags are guaranteed to be in the same shard
    * @param tags
    * @returns A map of shardId to tags
    */
-  generateShards(tags: string[]) {
+  generateShards({ tags, generateAllShards = false }: { tags: string[]; generateAllShards?: boolean }) {
```
vicb, line 159: A few comments:

I don't think "Shard" is the best in the fn name, options, and usage. Maybe better names could be "groupTagsByDo", "generateAllDO", ...

The function currently returns a map but the key is not used. Should it return a map of DOId -> tags

packages/cloudflare/src/api/do-sharded-tag-cache.ts
```diff
@@ -27,46 +34,159 @@ interface ShardedD1TagCacheOptions {
...
+  private generateDOIdArray({
+    tags,
+    shardType,
+    generateAllShards = false,
+  }: {
+    tags: string[];
+    shardType: "soft" | "hard";
+    generateAllShards: boolean;
+  }) {
+    let doubleShardArray: Array<number | undefined> = [1];
```
vicb, line 127:
```suggestion
    let replicaIndexes: Array<number | undefined> = [1];
```

packages/cloudflare/src/api/do-sharded-tag-cache.ts
conico974: > I don't think "Shard" is the best in the fn name, options, and usage. Maybe better names could be "groupTagsByDo", "generateAllDO", ...

Yeah, probably a good idea, we should stay consistent here. I'll try to make a big clean tomorrow morning to remove the remaining ones

> The function currently returns a map but the key is not used. Should it return a map of DOId -> tags

We could do that, i'll keep the map internally just for convenience though, it's way easier to work with `map.has` with a string than with a class

packages/cloudflare/src/api/do-sharded-tag-cache.ts
```diff
@@ -130,50 +248,71 @@ class ShardedD1TagCache implements NextModeTagCache {
...
   }
 
-  async getFromRegionalCache(shardId: string, tags: string[]) {
+  async getFromRegionalCache(doId: TagCacheDOId, tags: string[]) {
     try {
       if (!this.opts.regionalCache) return;
       const cache = await this.getCacheInstance();
       if (!cache) return;
-      const key = await this.getCacheKey(shardId, tags);
+      const key = await this.getCacheKey(doId, tags);
```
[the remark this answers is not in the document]
vicb: Oops that should be ok with the usage of the regional cache, missed that

packages/cloudflare/src/api/do-sharded-tag-cache.ts
conico974: We could add a rule in eslint for that. Pretty sure there is one in biome in the aws repo

packages/cloudflare/src/api/do-sharded-tag-cache.ts
```diff
@@ -27,46 +34,162 @@ interface ShardedD1TagCacheOptions {
...
+   * On read you only need to read from one of the shards, but on write you need to write to all shards
+   * @default false
+   */
+  enableShardReplication?: boolean;
+
+  /**
+   * The number of replicas that will be used for shard replication
+   * Soft shard replicas are more often accessed than hard shard replicas, so it is recommended to have more soft replicas than hard replicas
+   * Soft replicas are for internal next tags used for `revalidatePath` (i.e. `_N_T_/layout`, `_N_T_/page1`), hard replicas are the tags defined in your app
+   * @default { numberOfSoftShards: 4, numberOfHardShards: 2 }
```
vicb, line 53: shards -> replicas

vicb (approved):
Cool, thanks a lot for the updates 🚀

PR #487, opened 2025-03-21
feat: support passing the wrangler environment when populating the cache

james-elicx:
- Extracts the env flag's value from the passthrough args.
- Gives the value to the populate cache command.


vicb (approved):
Thanks James 🚀

PR #495, opened 2025-03-24
refactor: code cleanup

vicb:
minor cleanup/simplification.
There is some duplication but I think the code is simpler? (Couldn't get a generic `resolve` fn to work)


conico974 (approved):
LGTM

PR #496, opened 2025-03-24
Feat cache for the queue

conico974:
Durable object queue should already be able to handle most situation, but there is a case where for a single route, while being stale, you reach the limits of what a single durable object is capable of handling.

For this kind of cases, adding cache in front will help (but in case of error between the worker and the durable object, revalidation may take more time to happen)

I would appreciate some feedback on this, it might not even be worth it


packages/cloudflare/src/api/overrides/queue/queue-cache.ts
```diff
@@ -0,0 +1,120 @@
...
+        return;
+      }
+      if (!this.waitForQueueAck) {
+        await this.putToCache(msg);
+      }
+
+      await this.originalQueue.send(msg);
+      if (this.waitForQueueAck) {
+        await this.putToCache(msg);
+      }
```
vicb, line 51: nit: I think the below version more readable, but up to you

```suggestion
      if (!this.waitForQueueAck) {
        await this.putToCache(msg);
        await this.originalQueue.send(msg);
      } else {
        await this.originalQueue.send(msg);
        await this.putToCache(msg);
      }
```

packages/cloudflare/src/api/overrides/queue/queue-cache.ts
```diff
@@ -0,0 +1,120 @@
...
+  private async getCache() {
+    if (!this.cache) {
+      this.cache = await caches.open("durable-queue");
+    }
+    return this.cache;
+  }
+
+  private getCacheUrlString(msg: QueueMessage) {
+    return `queue/${msg.MessageGroupId}/${msg.MessageDeduplicationId}`;
+  }
```
vicb, line 68: Could you please change the return type to String to be consistent with other overiddes (i.e. regional-cache)

vicb: I meant to comment on the next method

packages/cloudflare/src/api/overrides/queue/queue-cache.ts
```diff
@@ -0,0 +1,120 @@
...
+    this.localCache.set(this.getCacheUrlString(msg), Date.now());
+    const cacheKey = this.getCacheKey(msg);
+    const cache = await this.getCache();
+    await cache.put(
+      cacheKey,
+      new Response(null, {
+        status: 200,
+        headers: {
+          "Cache-Control": `max-age=${this.regionalCacheTtlSec}`,
+          "Cache-Tag": `_N_T_/${msg.MessageBody.url}`,
```
vicb, line 84: Could please you add a comment here to explain the tag value?

packages/cloudflare/src/api/overrides/queue/queue-cache.ts
```diff
@@ -0,0 +1,120 @@
...
+}
+
+const DEFAULT_QUEUE_CACHE_TTL_SEC = 5;
+
+class QueueCache implements Queue {
+  readonly name;
+  readonly regionalCacheTtlSec: number;
+  readonly waitForQueueAck: boolean;
+  cache: Cache | undefined;
+  localCache: Map<string, number> = new Map();
```
vicb, line 27: can you please add a comment that this is mapping key to insertAtSec

packages/cloudflare/src/api/overrides/queue/queue-cache.ts
```diff
@@ -0,0 +1,120 @@
...
+      return true;
+    }
+  }
+
+  /**
+   * Remove any value older than the TTL from the local cache
+   */
+  private clearLocalCache() {
+    const now = Date.now();
+    for (const [key, value] of this.localCache.entries()) {
```
vicb, line 112: nits: 
- rename value to insertAtSec.
- maybe pre-compute `insertAtSecMax` and change the if condiition to `(insertAtSec > insertAtSecMax)`

packages/cloudflare/src/api/overrides/queue/queue-cache.ts
```diff
@@ -0,0 +1,123 @@
...
+  /**
+   * Remove any value older than the TTL from the local cache
+   */
+  private clearLocalCache() {
+    const now = Date.now();
+    const insertAtSecMax = this.regionalCacheTtlSec * 1000;
+    for (const [key, insertAtSec] of this.localCache.entries()) {
+      if (now - insertAtSec > insertAtSecMax) {
+        this.localCache.delete(key);
+      }
```
vicb, line 118: I meant something like

```suggestion
    const insertAtSecMax = Date.now() - this.regionalCacheTtlSec * 1000;
    for (const [key, insertAtSec] of this.localCache.entries()) {
      if (insertAtSec < insertAtSecMax) {
        this.localCache.delete(key);
      }
```

conico974: Ho yeah right, sorry i forgot to turn on my brain :smile:

vicb (approved):
LGTM

Let's add this to the next minor

PR #497, opened 2025-03-25
refactor: switch to bundled middleware

vicb:
This PR:
- Switch to using a bundled middleware - we were previsouly generating an external middleware but wrangler was bundling it together with the server bundle. We will revisit once we can deploy the middleware to a different worker.
- The server handler is now imported right before being used so that the env vars are guaranteed to be populated and the ALS initializzed.

Depends on https://github.com/opennextjs/opennextjs-aws/pull/798

### DIFF
--- a/.changeset/smooth-cups-allow.md
+++ b/.changeset/smooth-cups-allow.md
@@ -2,4 +2,4 @@
 "@opennextjs/cloudflare": patch
 ---
 
-Switch to bundle middleware
+Switch to bundled middleware
--- a/.changeset/smooth-cups-allow.md
+++ b/.changeset/smooth-cups-allow.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+Switch to bundle middleware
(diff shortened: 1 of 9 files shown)


.changeset/smooth-cups-allow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+Switch to bundle middleware
```
conico974, line 5:
```suggestion
Switch to bundled middleware
```

packages/cloudflare/src/cli/build/open-next/createServerBundle.ts
```diff
@@ -234,7 +234,9 @@ async function generateBundle(
       overrides,
     }),
 
-    openNextExternalMiddlewarePlugin(path.join(options.openNextDistDir, "core/edgeFunctionHandler.js")),
+    ...(config.middleware?.external
```
conico974, line 237: Should we do this ? If people do that middleware will not work for them (because of the update of the template)

Unless i'm missing something, at least for now we should probably not allow that (or at least put a warning)

conico974 (approved):
LGTM, just a few minor nits

packages/cloudflare/src/cli/build/open-next/createServerBundle.ts
vicb: I added a comment - `ensureCloudflareConfig` should error if users are trying to enable the external middleware

PR #503, opened 2025-03-26
fix: bun failing to spawn wrangler

james-elicx:
Bun seems to be the only package manager that does not support `exec` in the same way. Instead, they use it for their built-in shell.


conico974 (approved):
LGTM

PR #507, opened 2025-03-27
fix for `initOpenNextCloudflareForDev`

### LINKED ISSUE #502: [BUG] Build Fails Due to Missing Durable Object Classes in Worker Configuration
### Describe the bug

When attempting to create an optimized production build, the process fails due to missing Durable Object classes in the worker configuration. The following warnings and errors are encountered:

```
workerd/server/server.c++:1951: warning: A DurableObjectNamespace in the config referenced the class "DurableObjectQueueHandler", but no such Durable Object class is exported from the worker. Please make sure the class name matches, it is exported, and the class extends 'DurableObject'. Attempts to call to this Durable Object class will fail at runtime, but historically this was not a startup-time error. Future versions of workerd may make this a startup-time error.

workerd/server/server.c++:1951: warning: A DurableObjectNamespace in the config referenced the class "DOShardedTagCache", but no such Durable Object class is exported from the worker. Please make sure the class name matches, it is exported, and the class extends 'DurableObject'. Attempts to call to this Durable Object class will fail at runtime, but historically this was not a startup-time error. Future versions of workerd may make this a startup-time error.

service core:user:__WRANGLER_EXTERNAL_DURABLE_OBJECTS_WORKER: Worker "core:user:__WRANGLER_EXTERNAL_DURABLE_OBJECTS_WORKER"'s binding "DurableObjectQueueHandler" refers to a service "core:user:worker", but no such service is defined.

unhandledRejection [MiniflareCoreError [ERR_RUNTIME_FAILURE]: The Workers runtime failed to start. There is likely additional logging output above.] {
  code: 'ERR_RUNTIME_FAILURE',
  cause: undefined
}

Error: Command failed: npm run build
```

### Steps to reproduce

wrangler.json

```
{
  "$schema": "node_modules/wrangler/config-schema.json",
  "main": ".open-next/worker.js",
  "name": "my app",
  "compatibility_date": "2024-12-30",
  "compatibility_flags": ["nodejs_compat"],
  "minify": true,
  "assets": {
    "binding": "ASSETS",
  },
  "placement": {
    "mode": "smart",
  },
  "observability": {
    "enabled": true,
  },
  "r2_buckets": [
    {
      "bucket_name": "dimlly-isr-tags",
      "binding": "dimlly_isr_tags",
    },
  ],

  "durable_objects": {
    "bindings": [
      {
        "name": "NEXT_CACHE_REVALIDATION_DURABLE_OBJECT",
        "class_name": "DurableObjectQueueHandler",
      },
      {
        "name": "NEXT_CACHE_DO_SHARDED",
        "class_name": "DOShardedTagCache",
      },
    ],
  },
  "migrations": [
    {
      "tag": "v1",
      "new_sqlite_classes": ["DurableObjectQueueHandler", "DOShardedTagCache"],
    },
  ],
}
```

open-next.config.ts

```
import { defineCloudflareConfig } from '@opennextjs/cloudflare';
import kvIncrementalCache from '@opennextjs/cloudflare/overrides/incremental-cache/kv-incremental-cache';
import memoryQueue from '@opennextjs/cloudflare/overrides/queue/do-queue';
import doShardedTagCache from '@opennextjs/cloudflare/overrides/tag-cache/do-sharded-tag-cache';

  tagCache: doShardedTagCache({ baseShardSize: 12, regionalCache: true }),
  queue: memoryQueue,
});
```

### Expected behavior

Fix 

0.6.0

### Wrangler version

4.5.0

### next info output

```shell
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.3.0;
Binaries:
  Node: 23.5.0
  npm: 10.9.2
  Yarn: N/A
  pnpm: 9.15.4
Relevant Packages:
  next: 15.2.4
  eslint-config-next: 15.2.4
  react: 19.0.0
  react-dom: 19.0.0
  typescript: 5.8.2
Next.js Config:
  output: N/A
```

### Additional context

_No response_

conico974:
A quick fix for `initOpenNextCloudflareForDev` while waiting for a better solution in wrangler itself.
Fix #502

### DIFF
--- a/packages/cloudflare/src/api/cloudflare-context.ts
+++ b/packages/cloudflare/src/api/cloudflare-context.ts
@@ -288,7 +288,9 @@ async function patchWranglerConfig() {
   const originalWranglerConfigPath = `${process.cwd()}/wrangler.jsonc`;
   const patchedWranglerConfigPath = `${process.cwd()}/.open-next/wrangler.jsonc`;
   try {
-    const fs = await import("node:fs/promises");
+    // We can't use `node:fs/promises` here in next 14 it seems
+    // eslint-disable-next-line unicorn/prefer-node-protocol
+    const fs = await import("fs/promises");
     //@ts-ignore
     existingWranglerConfig = JSON.parse(await fs.readFile(originalWranglerConfigPath, "utf-8"));
     if (!existingWranglerConfig) {
--- a/packages/cloudflare/src/api/cloudflare-context.ts
+++ b/packages/cloudflare/src/api/cloudflare-context.ts
@@ -286,7 +286,7 @@ async function patchWranglerConfig() {
   //First thing here, we try to load the `wrangler.jsonc` file
   let existingWranglerConfig: Record<string, unknown> | null = null;
   const originalWranglerConfigPath = `${process.cwd()}/wrangler.jsonc`;
-  const patchedWranglerConfigPath = `${process.cwd()}/.open-next/wrangler.jsonc`;
+  const patchedWranglerConfigPath = `${process.cwd()}/.open-next/dev-wrangler.jsonc`;
   try {
     // We can't use `node:fs/promises` here in next 14 it seems
     // eslint-disable-next-line unicorn/prefer-node-protocol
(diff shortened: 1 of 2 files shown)


packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -282,6 +282,44 @@ async function monkeyPatchVmModuleEdgeContext(cloudflareContext: CloudflareConte
   };
 }
 
+async function patchWranglerConfig() {
+  //First thing here, we try to load the `wrangler.jsonc` file
+  let existingWranglerConfig: Record<string, unknown> | null = null;
+  const originalWranglerConfigPath = `${process.cwd()}/wrangler.jsonc`;
```
vicb, line 288: But:
- this does not support `toml` and `json` extensions
- this does not support custom config paths

I think the best workaround for now would be to set `script_name` to the worker name. It is not ideal either as `wrangler` then shows a misleading warning :(

ref: https://github.com/cloudflare/workers-sdk/issues/8687

conico974: I know it was more just a quick patch to test a fix for some cases and to get feedback (that's why it's a draft)

The big issue with the `script_name` is that in preview if you use one of the DO, it fails.
I think the good quick fix is in the workers-sdk to have an option to ignore DO for `getPlatformProxy`, we don't need more than that, just that it doesn't crash at build time.

conico974:
I'm gonna close this then, hopefully we'll get a fix pretty soon for https://github.com/cloudflare/workers-sdk/issues/8687

Issue #508, opened 2025-03-27
[BUG] TypeError: handler3 is not a function


LilaRest:
### Describe the bug

After upgrading from [0.5.12](https://github.com/opennextjs/opennextjs-cloudflare/releases/tag/%40opennextjs%2Fcloudflare%400.5.12) to [0.6.2](https://github.com/opennextjs/opennextjs-cloudflare/releases/tag/%40opennextjs%2Fcloudflare%400.6.2), running the preview server with `opennextjs-cloudflare build && opennextjs-cloudflare preview` leads to the app building and starting correctly, but when visiting `http://localhost:8787` I'm facing this error `TypeError: handler3 is not a function`.

**EDIT 1**: I tried a deployment and the same happens here. So it's not limited to preview.

**EDIT 2**: Downgrading to 0.6.1 and 0.6.0 didn't solve the issue, so it has likely be introduced in 0.6.0

**EDIT 3**: Downgrading to 0.5.12 solved the issue, without having to downgrade Wrangler. So that confirms that the issue is very likely related to `opennextjs-cloudflare`.

![Image](https://github.com/user-attachments/assets/c48fb718-d1d4-4152-9681-8f1c5b87cd82)

### Steps to reproduce

1. Upgrade to [0.6.2](https://github.com/opennextjs/opennextjs-cloudflare/releases/tag/%40opennextjs%2Fcloudflare%400.6.2)
2. Run preview

### Expected behavior

The preview running without error :)

### @opennextjs/cloudflare version

0.6.2

### Wrangler version

4.5.0

### next info output

```shell
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.4.0: Wed Mar 12 21:20:55 PDT 2025; root:xnu-11417.101.13~4/RELEASE_ARM64_T6031
  Available memory (MB): 131072
  Available CPU cores: 16
Binaries:
  Node: 23.3.0
  npm: 10.9.0
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  next: 15.2.4 // Latest available version is detected (15.2.4).
  eslint-config-next: N/A
  react: 19.0.0
  react-dom: 19.0.0
  typescript: 5.8.2
Next.js Config:
  output: N/A
```

vicb:
Could you please try to create a minimal repro and attached it to this issue (link to a GH repository).

There is not much we can do without more information.

vicb:
[...]

PR #509, opened 2025-03-27
fix: nextjs handler not detected in worker

ItsWendell:
Fixes https://github.com/opennextjs/opennextjs-cloudflare/issues/508

Full disclosure (also see issue above):

I don't know why this issue occurs. I just found this fix by experimenting. In the original situation, the worker detects the handler as a 'Getter', but if you use the 'handler' getter the result is undefined, but only on 'cold' boots. In a warm worker the Getter is initialized and it just works. By changing how we export in `handler.mjs` to a named export, it no longer detects it as a Getter but as a async function and it now works properly on cold boots. If someone could explain why this happens, please do!

### DIFF
--- a/.changeset/little-seals-impress.md
+++ b/.changeset/little-seals-impress.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+fix: nextjs handler not detected in worker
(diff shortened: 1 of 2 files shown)


vicb:
For ref, once thing that I tried is

```ts
      // @ts-expect-error: resolved by wrangler build
      const server = await import("./server-functions/default/handler.mjs");

      while (!server.handler) {
        await new Promise((resolve) => setTimeout(resolve, 0));
      }
```

It also fixes the issue. Meaning that `await` does not seem to work as expected...

ItsWendell:
> ```ts
>       while (!server.handler) {
> ```

Interesting! Do you know after how many event loops it resolves? Also tests run in wrangler's engine right? Interesting that it didn't catch this. Let me know if you figure out anything more, interesting to see what the core of the wildcard export / import issue is.

PR #512, opened 2025-03-27
fix: yarn v4 not passing args to wrangler correctly

james-elicx:
Yarn classic and yarn modern appear to pass args around differently. Yarn modern uses the `packageManager` field in package.json, so this PR uses that to compare the semver to check if it's used or not.

### DIFF
--- a/packages/cloudflare/src/cli/utils/run-wrangler.ts
+++ b/packages/cloudflare/src/cli/utils/run-wrangler.ts
@@ -28,6 +28,7 @@ function isYarnModern(options: BuildOptions) {
 
   if (!packageJson.packageManager?.startsWith("yarn")) return false;
 
+  const [, version] = packageJson.packageManager.split("@");
   return version ? compareSemver(version, ">=", "4.0.0") : false;
 }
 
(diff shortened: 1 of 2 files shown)


packages/cloudflare/src/cli/utils/run-wrangler.ts
```diff
@@ -12,18 +15,37 @@ type WranglerOptions = {
...
+  const packageJson: { packageManager?: string } = JSON.parse(
+    readFileSync(path.join(options.monorepoRoot, "package.json"), "utf-8")
+  );
+
+  if (!packageJson.packageManager?.startsWith("yarn")) return false;
+
+  const [, version] = packageJson.packageManager.split("@");
+  if (!version) return false;
+
+  return compareSemver(version, ">=", "4.0.0");
```
vicb, line 34:
```suggestion
  return version ? compareSemver(version, ">=", "4.0.0") : false;
```

What if version is ^4, ~4, 4.x?

vicb: > What if version is ^4, ~4, 4.x?

Maybe it's always a fixed version there?

vicb (approved):

PR #515, opened 2025-03-28
perf: optimize SQL queries

vicb:
`SELECT COUNT(*)` scans all the rows which takes longer and cost more. `LIMIT 1` is not effective for those queries as they always return a single row.


conico974 (approved):
LGTM

vicb:
Thanks!

PR #519, opened 2025-04-01
drop the AbortController patch

vicb:


vicb:
> LGTM. We should probably keep a small comment somewhere though about this with the link to the workerd issue

I was wondering if we should keep the patch (but not use it) and maybe remove it later.

WDYT?

conico974:
Yeah we can keep it for now and remove it once the fix is made in workerd, we could just add a small comment to explain why it's not needed anymore

vicb:
I update the PR:
- restored the rule and added comments on the rule
- stop applying the patch for now

conico974 (approved):

vicb:
Thanks Nico!

PR #520, opened 2025-04-01
Define `process.version` and `process.versions.node`

### LINKED ISSUE #333: [BUG] Error: Invariant: renderHTML should not be called in minimal mode

vicb:
fixes #333 

Jose (and other packages) relies on `process.version` and `process.versions.node`.

Defining them here until https://github.com/unjs/unenv/pull/493 is merged

/cc @petebacondarwin @mhart

### DIFF
--- a/packages/cloudflare/src/cli/templates/worker.ts
+++ b/packages/cloudflare/src/cli/templates/worker.ts
@@ -1,4 +1,5 @@
 import { AsyncLocalStorage } from "node:async_hooks";
+import process from "node:process";
 
 import type { CloudflareContext } from "../../api";
 // @ts-expect-error: resolved by wrangler build
@@ -67,6 +68,13 @@ function populateProcessEnv(url: URL, env: CloudflareEnv) {
   if (processEnvPopulated) {
     return;
   }
+
+  // Some packages rely on `process.version` and `process.versions.node` (i.e. Jose@4)
+  // TODO: Remove when https://github.com/unjs/unenv/pull/493 is merged
+  Object.assign(process, { version: process.version || "v22.14.0" });
+  // @ts-expect-error Node type does not match workerd
+  Object.assign(process.versions, { node: "22.14.0", ...process.versions });
+
   processEnvPopulated = true;
 
   for (const [key, value] of Object.entries(env)) {
--- a/.changeset/cyan-ears-know.md
+++ b/.changeset/cyan-ears-know.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+Define `process.version` and `process.versions.node`


james-elicx (approved):

packages/cloudflare/src/cli/templates/worker.ts
```diff
@@ -67,6 +68,13 @@ function populateProcessEnv(url: URL, env: CloudflareEnv) {
   if (processEnvPopulated) {
     return;
   }
+
+  // Some packages rely on `process.version` and `process.versions.node` (i.e. Jose@4)
+  // TODO: Remove when https://github.com/unjs/unenv/pull/493 is merged
+  Object.assign(process, { version: process.version || "v22.14.0" });
```
conico974, line 74: I'm wondering if this should be customizable. For this specific use case it's fine, but maybe some other libs may need a different version. 
By default it is `22.14` but could be set to whatever the user may need

james-elicx: IMO the libraries depending on process.version is an issue itself that they should seek to address. I believe jose no longer depends on process.version in its newer versions, and rightly so considering there are many non-node runtimes that it would be impossible to give a realistic value for...

conico974: Totally agree, but since we can't control that, i think a customizable option is the best we can do

sommeeeer: Could probably have a temporary env variable to customize it just in case. Since this will be removed once that PR gets merged anyways.

vicb: > I'm wondering if this should be customizable. For this specific use case it's fine, but maybe some other libs may need a different version.

I'm on the fence here...
Being able to customize the version  might help but then what happens if there are conflicting libs?
If we use a fixed version, we will get bug reports if some packages need a different version and we could start logging that.

> Could probably have a temporary env variable to customize it just in case. Since this will be removed once that PR gets merged anyways.

Well, the unenv PR sets this to 22.14 as well.

conico974: > If we use a fixed version, we will get bug reports if some packages need a different version and we could start logging that.

To be fair, we will likely get bug report even if we have the option to customize (and we could add a comment alongside to post issues related to that). And at least there will be a fix for these cases not involving another lib

As for the conflicting libs, there is not much we can do in these cases anyway with this (having a fixed version won't help either)

vicb: We will probably have a fixed version in `unenv` (see the PR linked in the description).

So I'm +1 to merge as is and update if later if needed. Having knobs means documentation, maintenance, ... I would like to avoid this if possible.

conico974: Fine by me, let's merge as is

conico974 (approved):

packages/cloudflare/src/cli/templates/worker.ts
vicb: 🤞

vicb:
Thanks for the feedback!
Let's merge and see how it goes.

PR #522, opened 2025-04-01
Log exceptions in requirePage when OPEN_NEXT_DEBUG=1

vicb:


packages/cloudflare/src/cli/build/patches/plugins/dynamic-requires.ts
```diff
@@ -135,7 +135,16 @@ function requirePage($PAGE, $DIST_DIR, $IS_APP_PATH) {
     process.env.__NEXT_PRIVATE_RUNTIME_TYPE = $IS_APP_PATH ? 'app' : 'pages';
   try {
     ${getRequires("pagePath", jsFiles, serverDir)}
```
conico974, line 137: We probably want to do something similar for the node module loader here ? https://github.com/opennextjs/opennextjs-cloudflare/blob/566a2673e216e942f845dfae594c897420e87d87/packages/cloudflare/src/cli/build/patches/plugins/dynamic-requires.ts#L97

conico974 (approved):
LGTM

vicb:
Thanks Nico!

PR #526, opened 2025-04-02
Release 1.0.0-beta.0

vicb:


packages/cloudflare/src/api/overrides/incremental-cache/r2-incremental-cache.ts
```diff
@@ -12,7 +12,7 @@ import { getCloudflareContext } from "../../cloudflare-context.js";
  * environment variable, and defaults to `incremental-cache`.
  */
 class R2IncrementalCache implements IncrementalCache {
-  readonly name = "r2-incremental-cache";
+  readonly name = "cf-r2-incremental-cache";
```
james-elicx, line 15: can you update the populate cache file to use this new name as well please

conico974 (approved):
LGTM, just one small comment

packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
```diff
@@ -150,17 +158,19 @@ class RegionalCache implements IncrementalCache {
  *
  * The regional cache uses the Cache API.
  *
- * **WARNING:** If an entry is revalidated in one region, it will trigger an additional revalidation if
+ * **WARNING:**
+ * If an entry is revalidated in one region, it will trigger an additional revalidation if
```
conico974, line 162:
```suggestion
 * If an entry is revalidated on demand in one region (using either `revalidateTag`, `revalidatePath` or `res.revalidate` ), it will trigger an additional revalidation if
```

james-elicx (approved):

packages/cloudflare/src/api/overrides/tag-cache/do-sharded-tag-cache.spec.ts
```diff
@@ -1,10 +1,6 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
-import doShardedTagCache, {
-  DEFAULT_HARD_REPLICAS,
-  DEFAULT_SOFT_REPLICAS,
-  TagCacheDOId,
-} from "./do-sharded-tag-cache";
+import shardedDOTagCache, { TagCacheDOId } from "./do-sharded-tag-cache";
```
james-elicx, line 3: should we rename this to DOTagCacheID?

james-elicx (approved):

packages/cloudflare/src/api/overrides/tag-cache/do-sharded-tag-cache.spec.ts
vicb: What about even simpler `DOId` which I think is what this is? (and it's only exported for tests).

/cc @conico974

vicb:
I push one more commit for TagCacheDOId -> DOId

changeset does not work for pre-releases :( 
(It is supposed to but there is a long running issue on the repo and I'm hitting this)
So I guess I'll generate a changelog in this PR for 1.0.0-beta.0

packages/cloudflare/src/api/overrides/tag-cache/do-sharded-tag-cache.spec.ts
james-elicx: Ah wasn't aware it was only being exported for tests, in that case i guess it doesn't really matter as much

PR #536, opened 2025-04-07
Fix R2 bucket population

### LINKED ISSUE #535: [BUG] R2 cache population
## populate cache

The command `opennextjs-cloudflate populateCache` populates a bucket named `NEXT_INC_CACHE_R2_BUCKET`. 

`NEXT_INC_CACHE_R2_BUCKET` is actually the name of the binding rather than the bucket.

We probably should use `getPlatformProxy()` to retrieve the binding and call `.put()` on it. The command should take an env a directory to persist data.

## cache handler

`getR2Key` returns `${directory}/${process.env.NEXT_BUILD_ID ?? "no-build-id"}/${key}.${isFetch ? "fetch" : "cache"}`

The `key` is typically `/path/to/file.cache` so we will end up with `BUILD_ID//key` which will not match the uploaded file.

@james-elicx thoughts?

vicb:
Fixes #535

### DIFF
--- a/.changeset/popular-berries-deliver.md
+++ b/.changeset/popular-berries-deliver.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+Fix R2 bucket population
--- a/packages/cloudflare/package.json
+++ b/packages/cloudflare/package.json
@@ -55,7 +55,8 @@
     "@dotenvx/dotenvx": "catalog:",
     "@opennextjs/aws": "3.5.4",
     "enquirer": "^2.4.1",
-    "glob": "catalog:"
+    "glob": "catalog:",
+    "ts-tqdm": "^0.8.6"
   },
   "devDependencies": {
     "@cloudflare/workers-types": "catalog:",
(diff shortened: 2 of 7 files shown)


james-elicx (approved):
thanks!

james-elicx (approved):

vicb:
Added a commit to hide the (expected) warnings for local mode and add a progress bar

```text

# before

▲ [WARNING] By default, `wrangler r2` commands access a local simulator of your R2 bucket, the same as that used by `wrangler dev`. To access your remote R2 bucket, re-run the command with the --remote flag


▲ [WARNING] By default, `wrangler r2` commands access a local simulator of your R2 bucket, the same as that used by `wrangler dev`. To access your remote R2 bucket, re-run the command with the --remote flag


▲ [WARNING] By default, `wrangler r2` commands access a local simulator of your R2 bucket, the same as that used by `wrangler dev`. To access your remote R2 bucket, re-run the command with the --remote flag

# after

Populating R2 incremental cache...
100%|█████████████████████████████████████████████| 50/50 [00:33:<00:00:, 1.51it/s]
Successfully populated cache with 50 assets

```

PR #542, opened 2025-04-07
drop the vercel-commerce example

vicb:
- The app does not build because of it uses the edge runtime (this could be fix by changing the way font is retrieved)
- It had no e2e
- It requires setting SHOPIFY_REVALIDATION_SECRET and SHOPIFY_STOREFRONT_ACCESS_TOKEN so it errors out of the box


conico974 (approved):

vicb:
Thanks Nico!

Issue #538, opened 2025-04-07
Auto-population for KV


james-elicx:
We need to migrate the KV adapter to populate the KV namespace instead of relying on static assets as a fallback. This also unlocks being able to use TTL for KV properly.

PR #546, opened 2025-04-07
feat: populate kv incremental cache

james-elicx:
This PR does the following:
- Replaces KV cache with a simplified version that only deals with KV.
- Aligns implementation with the one in r2 inc cache, using debug/error utils from opennext.
- Populates KV binding with assets.

Should already be covered by existing E2Es.


packages/cloudflare/src/api/overrides/incremental-cache/kv-incremental-cache.ts
```diff
@@ -95,68 +42,47 @@ class KVIncrementalCache implements IncrementalCache {
...
           value,
           // Note: `Date.now()` returns the time of the last IO rather than the actual time.
           //       See https://developers.cloudflare.com/workers/reference/security-model/
           lastModified: Date.now(),
-        })
+        }),
+        {
+          // When available, we only cache for the max revalidate time
+          ...(value.revalidate && {
+            expirationTtl: value.revalidate,
```
conico974, line 61: You likely don't want to do that actually. 
The incremental cache should work in a `stale-while-revalidate` way. (And for  `fallback:false` page it would break things)

james-elicx: i'll remove ttl for now - we can figure out how that should work separately to these changes

packages/cloudflare/src/api/overrides/incremental-cache/kv-incremental-cache.ts
```diff
@@ -23,69 +21,30 @@ class KVIncrementalCache implements IncrementalCache {
...
-    const assets = cfEnv.ASSETS;
-
-    if (!(kv || assets)) {
-      throw new IgnorableError(`No KVNamespace nor Fetcher`);
-    }
+    const kv = getCloudflareContext().env.NEXT_INC_CACHE_KV;
+    if (!kv) throw new IgnorableError("No KV Namespace");
 
-    this.debug(`Get ${key}`);
+    debug(`Get ${key}`);
```
vicb, line 27: I think this should be the other way around i.e. align R2 on what KV was before this PR (see https://github.com/opennextjs/opennextjs-cloudflare/issues/538)

packages/cloudflare/src/api/overrides/incremental-cache/kv-incremental-cache.ts
```diff
@@ -23,69 +21,30 @@ class KVIncrementalCache implements IncrementalCache {
...
     isFetch?: IsFetch
   ): Promise<WithLastModified<CacheValue<IsFetch>> | null> {
-    const cfEnv = getCloudflareContext().env;
-    const kv = cfEnv.NEXT_INC_CACHE_KV;
-    const assets = cfEnv.ASSETS;
-
-    if (!(kv || assets)) {
-      throw new IgnorableError(`No KVNamespace nor Fetcher`);
-    }
+    const kv = getCloudflareContext().env.NEXT_INC_CACHE_KV;
```
vicb, line 24: We can do `getCloudflareContext().env.NEXT_INC_CACHE_KV` in the ctor (see #538)

packages/cloudflare/src/api/overrides/incremental-cache/kv-incremental-cache.ts
```diff
@@ -1,16 +1,14 @@
...
-export const STATUS_DELETED = 1;
+import { IncrementalCacheEntry } from "./internal.js";
 
 export const NAME = "cf-kv-incremental-cache";
 
 /**
- * Open Next cache based on cloudflare KV and Assets.
+ * Open Next cache based on Cloudflare KV.
  *
  * Note: The class is instantiated outside of the request context.
```
vicb, line 13: This is no more true

packages/cloudflare/src/api/overrides/incremental-cache/kv-incremental-cache.ts
james-elicx: Sorry i didn't know you edited the comment directly.

packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -100,6 +101,25 @@ export async function populateCache(
...
+        logger.info("\nPopulating KV incremental cache...");
+
+        const assets = getCacheAssetPaths(options);
+        for (const { fsPath, destPath } of tqdm(assets)) {
+          runWrangler(
+            options,
+            [
+              "kv key put",
+              JSON.stringify(destPath),
+              "--binding NEXT_INC_CACHE_KV",
```
vicb, line 114: - `NEXT_INC_CACHE_KV` should be shared between the handler and here (see #538)
- we probably want to check if `NEXT_INC_CACHE_KV` exists instead of erroring as many times as there are assets

packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -100,6 +101,25 @@ export async function populateCache(
         logger.info(`Successfully populated cache with ${assets.length} assets`);
         break;
       }
+      case KV_CACHE_NAME: {
+        logger.info("\nPopulating KV incremental cache...");
```
vicb, line 105: can we extract the logic to functions so that the switch does not become huge?

packages/cloudflare/src/api/overrides/incremental-cache/kv-incremental-cache.ts
james-elicx: We still cannot do this.

image

vicb: > Sorry i didn't know you edited the comment directly.

I should have added a comment instead, sorry fo that.

When/how is this error triggered?
This class should not be used while `next dev`ing

james-elicx: It comes from getCloudflareContext -> getCloudflareContextSync. It throws an error because there was no cloudflare context initialised on the global scope

.changeset/metal-lemons-sparkle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+feat: auto-populate kv incremental cache
```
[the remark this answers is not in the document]
vicb: Also update the title of the PR

packages/cloudflare/src/api/overrides/incremental-cache/kv-incremental-cache.ts
vicb: We need to dig into that, it should not happen as it should be initialized at top level in worker.ts.

Also we should drop `next dev` only code from the Cloudflare build.

packages/cloudflare/src/api/overrides/incremental-cache/kv-incremental-cache.ts
```diff
@@ -23,69 +23,30 @@ class KVIncrementalCache implements IncrementalCache {
...
-            entry.value.data?.headers?.expires
-          ) {
-            const expiresTime = new Date(entry.value.data.headers.expires).getTime();
-            if (!isNaN(expiresTime) && expiresTime <= Date.now()) {
-              this.debug(`found expired entry (expire time: ${entry.value.data.headers.expires})`);
-              return null;
-            }
-          }
-        }
+      if (!entry || "lastModified" in entry) {
```
vicb, line 37: I think the following might be more readbale:

```ts
if (!entry) {
  return null;
}

return {
  value: entry,
  // blabla
  lastModified: entry.lastModified ?? BUILD_TS
};        
```

vicb (approved):
Thanks for all the updates, looks great!

packages/cloudflare/src/api/overrides/incremental-cache/kv-incremental-cache.ts
james-elicx: you'd need a ternary for both value and lastModified due to typescript at which point it isn't really an improvement...

could split that if statement into two though if that helps?

vicb: > could split that if statement into two though if that helps?

I think it would help understanding the code. But up to you (fine with either)

james-elicx:
```suggestion
      if (!entry) return null;
      
       if ("lastModified" in entry) {
```

packages/cloudflare/src/api/overrides/incremental-cache/kv-incremental-cache.ts
```diff
@@ -23,69 +23,32 @@ class KVIncrementalCache implements IncrementalCache {
...
-            const expiresTime = new Date(entry.value.data.headers.expires).getTime();
-            if (!isNaN(expiresTime) && expiresTime <= Date.now()) {
-              this.debug(`found expired entry (expire time: ${entry.value.data.headers.expires})`);
-              return null;
-            }
-          }
-        }
+      if (!entry) return null;
+      
+       if ("lastModified" in entry) {
```
james-elicx, line 39:
```suggestion
      if ("lastModified" in entry) {
```

PR #551, opened 2025-04-08
Set "cache-tag" on the regional cache entries

SamyPesse:
This PR aims at improving the use of the regional cache by making sure that entries on it can be purged by cache-tag on Cloudflare.

In our application (https://github.com/GitbookIO/gitbook), we purge the cache both using `revalidateTag` but also using the Cloudflare API (https://developers.cloudflare.com/cache/how-to/purge-cache/purge-by-tags/).

I think this header has no negative impact on any other OpenNext users while unblocking us but also potentially unblocking someone who is using the long-lived cache and needs to quickly purge it (they could do it from the CF dashboard).

### DIFF
--- a/.changeset/happy-bananas-buy.md
+++ b/.changeset/happy-bananas-buy.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+Set `Cache-Tag` on the entries created by the regional cache, to be purged using the Cloudflare API.
--- a/.changeset/happy-bananas-buy.md
+++ b/.changeset/happy-bananas-buy.md
@@ -2,4 +2,4 @@
 "@opennextjs/cloudflare": patch
 ---
 
-Set `Cache-Tag` on the entries created by the regional cache, to be purged using the Cloudflare API.
+Set `Cache-Tag` on the entries created by the regional cache, to be purged using the Cloudflare API or dashboard.
--- a/packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
+++ b/packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
@@ -36,8 +36,8 @@ type Options = {
 };
 
 interface PutToCacheInput {
-  requestKey: Request;
-  entryKey: string;
+  key: string;
+  isFetch: boolean | undefined;
   entry: IncrementalCacheEntry<boolean>;
 }
 
@@ -80,7 +80,7 @@ class RegionalCache implements IncrementalCache {
               const { value, lastModified } = rawEntry ?? {};
 
               if (value && typeof lastModified === "number") {
-                await this.putToCache({ requestKey, entryKey: key, entry: { value, lastModified } });
+                await this.putToCache({ key, isFetch, entry: { value, lastModified } });
               }
             })
           );
@@ -94,9 +94,7 @@ class RegionalCache implements IncrementalCache {
       if (!value || typeof lastModified !== "number") return null;
 
       // Update the locale cache after retrieving from the store.
-      getCloudflareContext().ctx.waitUntil(
-        this.putToCache({ requestKey, entryKey: key, entry: { value, lastModified } })
-      );
+      getCloudflareContext().ctx.waitUntil(this.putToCache({ key, isFetch, entry: { value, lastModified } }));
 
       return { value, lastModified };
     } catch (e) {
@@ -114,8 +112,8 @@ class RegionalCache implements IncrementalCache {
       await this.store.set(key, value, isFetch);
 
       await this.putToCache({
-        requestKey: this.getCacheKey(key, isFetch),
-        entryKey: key,
+        key,
+        isFetch,
         entry: {
           value,
           // Note: `Date.now()` returns the time of the last IO rather than the actual time.
@@ -155,7 +153,8 @@ class RegionalCache implements IncrementalCache {
     );
   }
 
-  protected async putToCache({ requestKey, entryKey, entry }: PutToCacheInput): Promise<void> {
+  protected async putToCache({ key, isFetch, entry }: PutToCacheInput): Promise<void> {
+    const requestKey = this.getCacheKey(key, isFetch);
     const cache = await this.getCacheInstance();
 
     const age =
@@ -165,7 +164,7 @@ class RegionalCache implements IncrementalCache {
 
     // We default to the entry key if no tags are found.
     // so that we can also revalidate page router based entry this way.
-    const tags = getTagsFromCacheEntry(entry) ?? [entryKey];
+    const tags = getTagsFromCacheEntry(entry) ?? [key];
     await cache.put(
       requestKey,
       new Response(JSON.stringify(entry), {
@@ -212,7 +211,7 @@ export function withRegionalCache(cache: IncrementalCache, opts: Options) {
 /**
  * Extract the list of tags from a cache entry.
  */
-export function getTagsFromCacheEntry(entry: IncrementalCacheEntry<boolean>): string[] | undefined {
+function getTagsFromCacheEntry(entry: IncrementalCacheEntry<boolean>): string[] | undefined {
   if ("tags" in entry.value && entry.value.tags) {
     return entry.value.tags;
   }
--- a/packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
+++ b/packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
@@ -66,7 +66,7 @@ class RegionalCache implements IncrementalCache {
   ): Promise<WithLastModified<CacheValue<IsFetch>> | null> {
     try {
       const cache = await this.getCacheInstance();
-      const urlKey = this.getCacheKey(key, isFetch);
+      const urlKey = this.getCacheUrlKey(key, isFetch);
 
       // Check for a cached entry as this will be faster than the store response.
       const cachedResponse = await cache.match(urlKey);
@@ -131,7 +131,7 @@ class RegionalCache implements IncrementalCache {
       await this.store.delete(key);
 
       const cache = await this.getCacheInstance();
-      await cache.delete(this.getCacheKey(key));
+      await cache.delete(this.getCacheUrlKey(key));
     } catch (e) {
       error("Failed to delete from regional cache", e);
     }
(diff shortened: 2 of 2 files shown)


packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
```diff
@@ -175,3 +183,25 @@ class RegionalCache implements IncrementalCache {
...
+    entry.value.meta &&
+    "headers" in entry.value.meta &&
+    entry.value.meta.headers
+  ) {
+    const rawTags = entry.value.meta.headers["x-next-cache-tags"];
+    if (typeof rawTags === "string") {
+      return rawTags.split(",");
+    }
+  }
+  return [];
```
conico974, line 206: If we don't find tags here, we probably want to return the original key here so that we could also revalidate page router based entry this way.

packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
```diff
@@ -175,3 +183,25 @@ class RegionalCache implements IncrementalCache {
 export function withRegionalCache(cache: IncrementalCache, opts: Options) {
```
conico974, line 183: We should add an option here to change the ttl of the cache from the default 30 minutes if we're adding cache tag (default would stay at 30 mins)

conico974 (commented):
Ideally, we'd want the purge to be automatic.
This means we'd have another override for the tag cache that would work in the same way as this one (by wrapping the tag cache used) and that would call cloudflare api to purge cache entry automatically on `writeTags`.
We'd also want to provide this https://opennext.js.org/aws/config/overrides/automatic_cdn_invalidation so that it would work for `res.revalidate` from page router as well.

packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
SamyPesse: Agree, I'll add one, it's just a bit redundant with `mode`, it could have been nice to have instead:

```
maxAge: 'short-lived' | 'long-lived' | number;
```

For now, I'll add a `defaultLongLivedAge`

packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
SamyPesse: I've done it in https://github.com/opennextjs/opennextjs-cloudflare/pull/551/commits/cedca1c3dea5ae90a66f8f964188c3fdd8e3f346

Let me know if understood correctly as I'm not sure how the page router works exactly.

SamyPesse:
> Ideally, we'd want the purge to be automatic.
This means we'd have another override for the tag cache that would work in the same way as this one (by wrapping the tag cache used) and that would call cloudflare api to purge cache entry automatically on writeTags.
We'd also want to provide this https://opennext.js.org/aws/config/overrides/automatic_cdn_invalidation so that it would work for res.revalidate from page router as well.

It makes sense, I guess it can be done in another PR?

@conico974 I've made the other suggested changes, let me know if they match what you had in mind.

packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
```diff
@@ -18,6 +18,14 @@ type Options = {
    */
   mode: "short-lived" | "long-lived";
 
+  /**
+   * The default age to use for long-lived cache entries.
+   * When no revalidate is provided, the default age will be used.
+   *
+   * @default `THIRTY_MINUTES_IN_SECONDS`
+   */
+  defaultLongLivedAge?: number;
```
conico974, line 27: We use ttl in other places, for consistency we should use something similar here, something like `defaultLongLivedTtlSec`

packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
```diff
@@ -135,18 +143,32 @@ class RegionalCache implements IncrementalCache {
     );
   }
 
-  protected async putToCache(key: Request, entry: IncrementalCacheEntry<boolean>): Promise<void> {
+  protected async putToCache(
+    key: Request,
```
conico974, line 147: For readability i think it would be better to have an object instead of 3 params to this function, something like 
```ts
interface PutToCacheOpts {
  requestKey: string,
  entryKey: string,
  entry: IncrementalCacheEntry<boolean>
}
```

conico974 (commented):
I just made 2 small comments but other than that, it looks good
> It makes sense, I guess it can be done in another PR?

Yeah we can do it in another PR later.

packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
conico974: That's basically what i had in mind yeah

packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
SamyPesse: Agree, I changed it

conico974 (approved):
LGTM Thanks

.changeset/happy-bananas-buy.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+Set `Cache-Tag` on the entries created by the regional cache, to be purged using the Cloudflare API.
```
vicb, line 5:
```suggestion
Set `Cache-Tag` on the entries created by the regional cache, to be purged using the Cloudflare API or dashboard.
```

packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
```diff
@@ -18,6 +18,14 @@ type Options = {
    */
   mode: "short-lived" | "long-lived";
 
+  /**
+   * The default age to use for long-lived cache entries.
+   * When no revalidate is provided, the default age will be used.
+   *
+   * @default `THIRTY_MINUTES_IN_SECONDS`
+   */
+  defaultLongLivedTtlSec?: number;
```
vicb, line 27:
```suggestion
  longLivedTtlSec?: number;
```

packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
```diff
@@ -18,6 +18,14 @@ type Options = {
    */
   mode: "short-lived" | "long-lived";
 
+  /**
+   * The default age to use for long-lived cache entries.
```
vicb, line 22:
```suggestion
   * The TTL of long-lived cache entries.
```

packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
```diff
@@ -135,18 +155,28 @@ class RegionalCache implements IncrementalCache {
...
 
-  protected async putToCache(key: Request, entry: IncrementalCacheEntry<boolean>): Promise<void> {
+  protected async putToCache({ requestKey, entryKey, entry }: PutToCacheInput): Promise<void> {
     const cache = await this.getCacheInstance();
 
     const age =
       this.opts.mode === "short-lived"
         ? ONE_MINUTE_IN_SECONDS
-        : entry.value.revalidate || THIRTY_MINUTES_IN_SECONDS;
+        : entry.value.revalidate || this.opts.defaultLongLivedTtlSec || THIRTY_MINUTES_IN_SECONDS;
```
vicb, line 164: Shouldn't the TTL be longer when revalidate = false?

packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
```diff
@@ -27,6 +35,12 @@ type Options = {
   shouldLazilyUpdateOnCacheHit?: boolean;
 };
 
+interface PutToCacheInput {
+  requestKey: Request;
```
vicb, line 39: Maybe pass `entryKey` and `isFetch` instead of `requestKey`?

`requestKey` could be changed to a string, would be nice to apply `.replaceAll(/\/+/, "/")` as we do in other cache handler because keys start with '/'

packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
```diff
@@ -169,9 +199,33 @@ class RegionalCache implements IncrementalCache {
...
  * @default `false` for the `short-lived` mode, and `true` for the `long-lived` mode.
  */
 export function withRegionalCache(cache: IncrementalCache, opts: Options) {
   return new RegionalCache(cache, opts);
 }
+
+/**
+ * Extract the list of tags from a cache entry.
+ */
+export function getTagsFromCacheEntry(entry: IncrementalCacheEntry<boolean>): string[] | undefined {
```
vicb, line 215: Does this needs to be exported?

vicb:
Thanks @SamyPesse for the PR, could you please prepare a PR for the docs at https://github.com/opennextjs/docs

@SamyPesse or @conico974 could you please create a feature request for the follow up.

Thanks!

packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
conico974: No at the moment there is no revalidate value at all for ISR/SSG route, so as a default we fallback to a smaller value

packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
SamyPesse: Changed

packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
SamyPesse: No, I removed the `export`

SamyPesse:
Actually, for the documentation, it could be better to wait for https://github.com/opennextjs/opennextjs-cloudflare/issues/555 before writing it as by itself this PR doesn't make it a standalone solution for most people using Opennext.

vicb:
I added one last commit + rebase, let me know if this looks good

SamyPesse:
Perfect! thanks @vicb

PR #563, opened 2025-04-10
Extract the worker init code to a separate file

vicb:
I have been willing to do this for a long time...

The init code does not belong to the server bundle, it can be used for i.e. an external middleware.

Further benefits:
- We now have a way to inject values a build time via ESBuild defines, that will be useful for basePath support
- It will make it easier to add other entry points (scheduled, ...)

### DIFF
--- a/packages/cloudflare/src/cli/templates/init.ts
+++ b/packages/cloudflare/src/cli/templates/init.ts
@@ -94,8 +94,6 @@ function initRuntime() {
 
   Object.assign(globalThis, {
     Request: CustomRequest,
-    // This is only needed for an external middleware bundle
-    __dangerous_ON_edge_converter_returns_request: true,
     //@ts-expect-error Inline at build time by ESBuild
     __BUILD_TIMESTAMP_MS__: __BUILD_TIMESTAMP_MS__,
   });
--- a/packages/cloudflare/src/cli/templates/init.ts
+++ b/packages/cloudflare/src/cli/templates/init.ts
@@ -59,9 +59,6 @@ function initRuntime() {
   // @ts-expect-error Node type does not match workerd
   Object.assign(process.versions, { node: "22.14.0", ...process.versions });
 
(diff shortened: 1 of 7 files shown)


packages/cloudflare/src/cli/templates/init.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  const mode = env.NEXTJS_ENV ?? "production";
+  if (nextEnvVars[mode]) {
+    for (const key in nextEnvVars[mode]) {
+      process.env[key] ??= nextEnvVars[mode][key];
+    }
+  }
+
+  // Set the default Origin for the origin resolver.
+  process.env.OPEN_NEXT_ORIGIN = JSON.stringify({
```
conico974, line 124: Not needed anymore now that the middleware is bundled

conico974 (approved):
LGTM, just one small comment

packages/cloudflare/src/cli/templates/init.ts
vicb: Same goes for `globalThis.__dangerous_ON_edge_converter_returns_request = true;` but that init could also be used for the middleware bundle, I will add comments

packages/cloudflare/src/cli/templates/init.ts
```diff
@@ -0,0 +1,133 @@
...
+        });
+      }
+      super(input, init);
+    }
+  };
+
+  Object.assign(globalThis, {
+    Request: CustomRequest,
+    // This is only needed for an external middleware bundle
+    __dangerous_ON_edge_converter_returns_request: true,
```
conico974, line 98: Do we want to add that actually for an external middleware ? This means that the output of the middleware will be a Request.
It's fine when everything was in 1 worker, but if we have multiple worker it won't work as expected (or it will require a custom converter/wrapper to forward the request)

vicb: That's right, I'll remove this.
Thanks for catching that.

PR #581, opened 2025-04-14
fix(tagCache): gracefully handle empty tag list

### LINKED ISSUE #576: [BUG] Worker tries to revalidate empty tags resulting in: D1_ERROR: No SQL statements detected.
### Describe the bug

Trying to fetch an api route from a client component when using the recommended setup for caching results in a `D1_ERROR: No SQL statements detected.` error due to the D1 batch command receiving an invalid SQL command.

### Steps to reproduce

1. clone [this](https://github.com/ilovemesomeramen/opennextjs-cloudflare-empty-revalidation) repo
2. preview with `npm run preview`
3. visit http://localhost:8787
4. inspect worker logs

### Expected behavior

I assume there should not be any re-validation here in the first place, or at least empty tags should not be re-validated 

1.0.0-beta.2

### Wrangler version

4.10.0

### next info output

```shell
▲ [WARNING]                             You have defined bindings to the following internal Durable Objects:

                                - {"name":"NEXT_CACHE_DO_QUEUE","class_name":"DOQueueHandler"}
                                These will not work in local development, but they should work in production.
  
                                If you want to develop these locally, you can define your DO in a separate Worker, with a
  separate configuration file.
                                For detailed instructions, refer to the Durable Objects section here:
  https://developers.cloudflare.com/workers/wrangler/api#supported-bindings


workerd/server/server.c++:1942: warning: A DurableObjectNamespace in the config referenced the class "DOQueueHandler", but no such Durable Object class is exported from the worker. Please make sure the class name matches, it is exported, and the class extends 'DurableObject'. Attempts to call to this Durable Object class will fail at runtime, but historically this was not a startup-time error. Future versions of workerd may make this a startup-time error.
/bin/sh: line 1: yarn: command not found

Operating System:
  Platform: linux
  Arch: x64
  Version: #1 SMP PREEMPT_DYNAMIC Mon, 24 Mar 2025 06:33:36 +0000
  Available memory (MB): 63483
  Available CPU cores: 16
Binaries:
  Node: 23.9.0
  npm: 11.2.0
  Yarn: N/A
  pnpm: 10.7.0
Relevant Packages:
  next: 15.2.4 // There is a newer version (15.3.0) available, upgrade recommended! 
  eslint-config-next: N/A
  react: 19.1.0
  react-dom: 19.1.0
  typescript: 5.8.3
Next.js Config:
  output: N/A
```

### Additional context

this same error also occurs using a deployed version on cloudflare

vicb:
fixes #576

### DIFF
--- a/.changeset/cuddly-mails-happen.md
+++ b/.changeset/cuddly-mails-happen.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+fix(tagCache): gracefully handle empty tag list
--- a/packages/cloudflare/src/api/overrides/tag-cache/d1-next-tag-cache.ts
+++ b/packages/cloudflare/src/api/overrides/tag-cache/d1-next-tag-cache.ts
@@ -36,7 +36,8 @@ export class D1NextModeTagCache implements NextModeTagCache {
 
   async writeTags(tags: string[]): Promise<void> {
     const { isDisabled, db } = this.getConfig();
-    if (isDisabled) return Promise.resolve();
+    // TODO: Remove `tags.length === 0` when https://github.com/opennextjs/opennextjs-aws/pull/828 is used
+    if (isDisabled || tags.length === 0) return Promise.resolve();
     const result = await db.batch(
       tags.map((tag) =>
         db


conico974 (commented):
We should do it here aswell https://github.com/opennextjs/opennextjs-cloudflare/blob/68d6bb4736b75cf1850c6f21b8ef2aa2e059f51b/packages/cloudflare/src/api/overrides/tag-cache/do-sharded-tag-cache.ts#L256

vicb:
> We should do it here aswell
> 
> https://github.com/opennextjs/opennextjs-cloudflare/blob/68d6bb4736b75cf1850c6f21b8ef2aa2e059f51b/packages/cloudflare/src/api/overrides/tag-cache/do-sharded-tag-cache.ts#L256

Woudn't that code path work with an empty tag list?

conico974 (approved):
You're right, it should work just fine. Thanks

vicb:
> You're right, it should work just fine. Thanks

Note that I would have added it if the root cause was not fixed upstream.

Thanks for the review!

Issue #548, opened 2025-04-08
[BUG] postgres throws errors at runtime


ka-raa-ge:
### Describe the bug

Using the postgres-js driver to access a database works correctly on the first request, but fails with the following errors on every other subsequent request.
```
X [ERROR] A hanging Promise was canceled. This happens when the worker runtime is waiting for a Promise from JavaScript to resolve, but has detected that the Promise cannot possibly ever resolve because all code and events related to the Promise's I/O context have already finished.
[wrangler:inf] GET / 500 Internal Server Error (9ms)
X [ERROR] Uncaught (in response) Error: The script will never generate a response.
X [ERROR] Uncaught (async) Error: Promise will never complete.
```

### Steps to reproduce

1. Create an opennext project using `npm create cloudflare@latest -- my-next-app --framework=next --platform=workers`
2. Install postgres `npm install --save postgres`
3. Create a database script. Making sure to have the `DATABASE_URL` defined to a valid pg database: 
```
import postgres from "postgres";
const connectionString = process.env.DATABASE_URL;
export const client = postgres(connectionString!);
```
4. Replace the contents of `app/page.tsx` with the following. Change the db query to match a test table in your databse.
```
import { client } from "@/db/db";

export const dynamic = "force-dynamic";

export default async function Home() {
  const result = await client`SELECT * FROM users;`;
  return <h1>{JSON.stringify(result)}</h1>;
}
```
5. Run the worker `npm run preview`
6. Open the page in the browser, refresh the page to trigger the error.

### Expected behavior

The request should consistently resolve, fetching from the database and return the data.
An equivalent request on a bare-bones typsecript cloudflare worker project resolves the same request without issues.

### @opennextjs/cloudflare version

~1.0.0-beta.0 || ^1.0.0

### Wrangler version

^4.8.0

### next info output

```shell
Operating System:
  Platform: linux
  Arch: x64
  Version: #1 SMP Tue Nov 5 00:21:55 UTC 2024
  Available memory (MB): 31983
  Available CPU cores: 20
Binaries:
  Node: 22.14.0
  npm: 10.9.2
  Yarn: N/A
  pnpm: 10.8.0
Relevant Packages:
  next: 15.2.4 // Latest available version is detected (15.2.4).
  eslint-config-next: 15.2.4
  react: 19.1.0
  react-dom: 19.1.0
  typescript: 5.8.3
Next.js Config:
  output: N/A
```

vicb:
[...]

PR #593, opened 2025-04-18
Allow using external packages with a workerd build condition

vicb:
Should help with https://github.com/opennextjs/opennextjs-cloudflare/issues/548

This PR change `bundleServer()` to use the "workerd" build condition to allow including code specifically created for workerd.

When a package uses the "workerd" build conditions, we copy all the files from the source package - Next would not trace/copy files for the workerd runtime.

Note that the packages have to be declared in the `serverExternalPackages` of the Next config. Otherwise they are bundled during the Next build and not in the `nft.json` files as a result OpenNext can not detect them. This behavior is not ideal, happy to hear about any better idea!

TODO

- [x] add more comments
- [x] add some docs
- [x] use an `@opennext/aws` release in the package.json

/cc @ka-raa-ge @mhart

### DIFF
--- a/.changeset/large-adults-report.md
+++ b/.changeset/large-adults-report.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+Use the workerd build condition by default


packages/cloudflare/src/cli/build/utils/workerd.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+    expect(hasBuildCondition(exports, "workerd")).toBe(true);
+    expect(hasBuildCondition(exports, "default")).toBe(true);
+    expect(hasBuildCondition(exports, "module")).toBe(false);
+  });
+
+  test("only consider leaves", () => {
+    const exports = {
+      ".": "/path/to/index.js",
+      "./server": {
+        worked: {
```
conico974, line 41:
```suggestion
        workerd: {
```

packages/cloudflare/src/cli/build/utils/workerd.ts
```diff
@@ -0,0 +1,54 @@
...
+  for (const [src, dst] of nodePackages.entries()) {
+    try {
+      const { exports } = JSON.parse(await fs.readFile(path.join(src, "package.json"), "utf8"));
+      const match = src.match(isNodeModuleRegex);
+      if (
+        match?.groups?.pkg &&
+        externalPackages.includes(match.groups.pkg) &&
+        hasBuildCondition(exports, "workerd")
+      ) {
+        fs.cp(src, dst, { recursive: true, force: true });
```
conico974, line 48: We should log something here in debug in case something goes wrong. It may help debugging and identify package that cause trouble

packages/cloudflare/src/cli/build/utils/workerd.ts
```diff
@@ -0,0 +1,54 @@
...
+export async function copyWorkerdPackages(options: BuildOptions, nodePackages: Map<string, string>) {
+  const isNodeModuleRegex = getCrossPlatformPathRegex(`.*/node_modules/(?<pkg>.*)`, { escape: false });
+
+  // Copy full external packages when they use "workerd" build condition
+  const nextConfig = loadConfig(path.join(options.appBuildOutputPath, ".next"));
+  const externalPackages = nextConfig.serverExternalPackages ?? [];
+  for (const [src, dst] of nodePackages.entries()) {
+    try {
+      const { exports } = JSON.parse(await fs.readFile(path.join(src, "package.json"), "utf8"));
+      const match = src.match(isNodeModuleRegex);
```
conico974, line 42: We should warn the user if `!externalPackages.includes(match.groups.pkg) &&
        hasBuildCondition(exports, "workerd")`

vicb: I'm not sure what the best thing to do is here, do you have something in mind?

`hasBuildCondition(...)` check for traced packages (`nodePackages`).

When I test with `postgres`, it would not be traced unless added in `serverExternalPackages` - if not there it is bundled.

I think the warning would trigger for `react-dom` and packages in [server-external-packages.json](https://github.com/vercel/next.js/blob/1f93d4975029cf591355fec095d7d16eaa1075a7/packages/next/src/lib/server-external-packages.json)

conico974: Ho yeah right.

I think we should still add the warning here, at least for the packages you mentionned (we should probably filter some like `react-dom`)

One thing we could do is maintain a list of known deps that cause trouble like `postgres` and look if they're in the user package.json
That's not ideal, but at least it will help a bit.

Another option would be to look at the lockfile, but it will show a lot of false positive (especially for monorepo) and would be complex to maintain (especially for bun and the binary lockfile)

I can't think of anything else at the moment

conico974: FYI, i don't consider this blocking, we could figure this out later if you prefer

vicb:
I have updated the PR so that we can disabled using the workerd condition if it causes troubles.

Can I please get a review?

conico974 (approved):
LGTM
We should add something in the docs for this. And probably something related to that in the troubleshooting sections

vicb:
> LGTM We should add something in the docs for this. And probably something related to that in the troubleshooting sections

I am thinking about updating https://opennext.js.org/cloudflare/troubleshooting#my-app-fails-to-build-when-i-import-a-specific-npm-package for now.

We should list the packages we know using the workerd condition: `postgres` for now

Thanks for the review

PR #599, opened 2025-04-22
fix a few windows path issues

james-elicx:
**Changes:**

- Caches playwright browsers.
- Fixes several path normalisation issues.

### DIFF
--- a/.changeset/olive-lemons-dance.md
+++ b/.changeset/olive-lemons-dance.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+Fix multiple Windows path issues.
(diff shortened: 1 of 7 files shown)


vicb:
I have squashed the commits and rebased on main

vicb:
I was thinking about that earlier today, we have 3 choices:

- drop windows specific code
- merge the PR now knowing that we won't invest further in windows w/o WSL
- keep the PR opened if someone wants to take over

Let's discuss during the next weekly?

james-elicx:
I don't think it makes sense to drop Windows-specific code - if someone wants to come along and fix it in the future then there will still be a decent starting point instead of starting from scratch.

It's just that we won't invest any time going forwards in ensuring new (or old) code supports Windows.

vicb (approved):
Let's get this merged.

Could you please rebase on main when you get a chance?

vicb (approved):
Thanks James 🙏

PR #608, opened 2025-04-28
Feat tag-cache-filter

conico974:
Add a new `withFilter` "tag cache" that allows to filter the tag used.
It is mostly useful to reduce the load on the underlying tag cache.
It also introduce a `softTagFilter` function for those that uses `revalidateTag`, but not `revalidatePath`

### DIFF
--- a/.changeset/neat-hornets-call.md
+++ b/.changeset/neat-hornets-call.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+Add a new `withFilter` tag cache to allow to filter the tag cache used
--- a/packages/cloudflare/src/api/overrides/tag-cache/tag-cache-filter.ts
+++ b/packages/cloudflare/src/api/overrides/tag-cache/tag-cache-filter.ts
@@ -18,20 +18,19 @@ interface WithFilterOptions {
  * Creates a new tag cache that filters tags based on the provided filter function.
  * This is usefult to remove tags that are not used by the app, this could reduce the number of request to the underlying tag cache.
  */
-export function withFilter({
-  originalTagCache,
-  filterFn,
-}: WithFilterOptions): NextModeTagCache {
+export function withFilter({ originalTagCache, filterFn }: WithFilterOptions): NextModeTagCache {
   return {
     name: `filtered-${originalTagCache.name}`,
     mode: "nextMode",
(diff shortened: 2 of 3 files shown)


.changeset/neat-hornets-call.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+Add a new `withFilter` tag cache to allow to filter the tag cache used
```
vicb, line 5:
```suggestion
Add a new `withFilter` tag cache to allow to filter the tags used
```

packages/cloudflare/src/api/overrides/tag-cache/tag-cache-filter.ts
```diff
@@ -0,0 +1,58 @@
+import { NextModeTagCache } from "@opennextjs/aws/types/overrides";
+
+interface WithFilterOptions {
+  /**
+   * The original tag cache.
+   * Call to this will receive only the filtered tags.
+   */
+  originalTagCache: NextModeTagCache;
```
vicb, line 8: maybe?
```suggestion
  /**
   * The wrapped tag cache.
   * Call to this will receive only the filtered tags.
   */
  tagCache: NextModeTagCache;
```

packages/cloudflare/src/api/overrides/tag-cache/tag-cache-filter.ts
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * The function to filter tags.
+   * @param tag The tag to filter.
+   * @returns true if the tag should be forwarde, false otherwise.
```
vicb, line 12:
```suggestion
   * @returns true if the tag should be forwarded, false otherwise.
```

vicb (approved):
LGTM with a few minor comments

packages/cloudflare/src/api/overrides/tag-cache/tag-cache-filter.ts
```diff
@@ -0,0 +1,58 @@
...
+   */
+  filterFn: (tag: string) => boolean;
+}
+
+/**
+ * Creates a new tag cache that filters tags based on the provided filter function.
+ * This is usefult to remove tags that are not used by the app, this could reduce the number of request to the underlying tag cache.
```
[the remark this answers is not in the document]
sommeeeer: `useful`

PR #610, opened 2025-04-29
add the global_fetch_strictly_public flag to all configs

vicb:


conico974 (approved):
LGTM
We should also remove the service binding, probably in a different PR though

james-elicx (approved):

vicb:
> We should also remove the service binding, probably in a different PR though

Yes, we should make it optional for some time to avoid breaking backward compatibility and remove it later. Will be follow up PRs.

Thanks for the review

PR #613, opened 2025-04-29
Sync with aws for 3.6 and composable cache

conico974:
- [x] Change to use 3.6

I'll put the new experimental e2e app in another PR for readability

PR #614, opened 2025-04-29
Add the experimental e2e app

conico974:
The node middleware has been disabled for now, it breaks because of a dynamic require.
Not sure if it makes sense to fix it (at least not in this PR)
Require #613 to be merged first


vicb:
> Not sure if it makes sense to fix it (at least not in this PR)

+1 to postpone that. Could you please create a tracking issue though.

examples/e2e/experimental/e2e/nodeMiddleware.test.ts
```diff
@@ -0,0 +1,31 @@
+import { expect, test } from "@playwright/test";
+
+test.describe("Node Middleware", () => {
```
vicb, line 3: Please add a link to the tracking issue when created

examples/e2e/experimental/next.config.ts
```diff
@@ -0,0 +1,22 @@
...
+  eslint: {
+    ignoreDuringBuilds: true,
+  },
+  typescript: {
+    // Ignore type errors during build for now, we'll need to figure this out later
+    ignoreBuildErrors: true,
+  },
+  experimental: {
+    ppr: "incremental",
+    // Node middleware is not supported yet in cloudflare
```
vicb, line 16: Please add the link here also

examples/e2e/experimental/src/middleware.ts
```diff
@@ -0,0 +1,27 @@
+// import crypto from "node:crypto";
```
vicb, line 1: and here!

pnpm-workspace.yaml
```diff
@@ -40,11 +40,11 @@ catalog:
...
-    "@types/react-dom": 19.0.3
-    "@types/react": 19.0.8
+    "@types/react-dom": 19.0.0
+    "@types/react": 19.0.0
     autoprefixer: 10.4.15
     next: ~15.3.0
     postcss: 8.4.27
-    react-dom: ^19.0.0
-    react: ^19.0.0
+    react-dom: 19.0.0
```
vicb, line 48: Out of curiosity, why removing `^`?
(I think the versions comes from a Next app, including types).

vicb (approved):
Thanks!

pnpm-workspace.yaml
conico974: I wanted to pinpoint react to a specific version, I realized after that it was a typing issue but did not revert that change

vicb: Do we want to revert that?

conico974: Not a big fan of these in package.json personally, so unless we have a good reason to do it, i'd rather have it fixed

PR #616, opened 2025-04-29
feat: prefix for kv cache keys

### LINKED ISSUE #556: [Feature] Support directory for KV inc cache
R2 uses a [directory](https://github.com/opennextjs/opennextjs-cloudflare/blob/31b31dd2dd2baad784fe2d29f01c059fa641dae4/packages/cloudflare/src/api/overrides/incremental-cache/r2-incremental-cache.ts#L81)

We should add it to KV for consistency

See discussion on https://github.com/opennextjs/opennextjs-cloudflare/pull/553

james-elicx:
- Shares cache key function between inc caches.
- Adds prefix for KV cache keys.

fixes #556

### DIFF
--- a/packages/cloudflare/src/cli/commands/populate-cache.ts
+++ b/packages/cloudflare/src/cli/commands/populate-cache.ts
@@ -16,12 +16,11 @@ import { unstable_readConfig } from "wrangler";
 
 import {
   BINDING_NAME as KV_CACHE_BINDING_NAME,
-  computeCacheKey as computeKVCacheKey,
   NAME as KV_CACHE_NAME,
+  PREFIX_ENV_NAME as KV_CACHE_PREFIX_ENV_NAME,
 } from "../../api/overrides/incremental-cache/kv-incremental-cache.js";
 import {
   BINDING_NAME as R2_CACHE_BINDING_NAME,
-  computeCacheKey as computeR2CacheKey,
   NAME as R2_CACHE_NAME,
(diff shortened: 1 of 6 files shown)


packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -113,8 +113,8 @@ function populateR2IncrementalCache(
   const assets = getCacheAssets(options);
 
   for (const { fullPath, key, buildId, isFetch } of tqdm(assets)) {
-    const cacheKey = computeR2CacheKey(key, {
-      directory: process.env[R2_CACHE_PREFIX_ENV_NAME],
+    const cacheKey = computeCacheKey(key, {
+      prefix: process.env[R2_CACHE_PREFIX_ENV_NAME],
```
vicb, line 117: `process.env` will be the Node.js env here. So users will have to set the prefix var there in a different way from how they set the runtime value.

This sounds error prone (and I don't think this is something we document?)

Maybe we could retrieve the value via the platform proxy or via a parameter?

What do you think?

vicb (commented):
Thanks James!

I added one question about how the prefix is retrieved when populating the cache.

The PR otherwise LGTM

packages/cloudflare/src/cli/commands/populate-cache.ts
james-elicx: I think platform proxy sounds like a good idea. I'll put up a separate PR for those changes

vicb (approved):
Thanks!

PR #619, opened 2025-04-30
Fix node timers

### LINKED ISSUE #587: [BUG] Adding Sentry causes app to crash in preview mode
### Describe the bug

After integrating Sentry, we observed that wrangler fails to run the app in preview mode, resulting in a crash.

### Steps to reproduce

[Link](https://github.com/argo22com/opennextjs-sentry) to minimum reproduction.

1) npm run preview
2) When Wrangler shows Ready on http://localhost:8787, press b in the terminal to open the browser.
3) The error will appear in the browser and terminal.

### Expected behavior

The app should run without crashing.

@opennextjs/cloudflare@1.0.0-beta.2

### Wrangler version

4.10.0

### next info output

```shell
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.2.0: Fri Dec  6 19:01:59 PST 2024; root:xnu-11215.61.5~2/RELEASE_ARM64_T6000
  Available memory (MB): 16384
  Available CPU cores: 10
Binaries:
  Node: 23.6.0
  npm: 10.9.2
  Yarn: N/A
  pnpm: 7.14.2
Relevant Packages:
  next: 15.2.5 // There is a newer version (15.3.0) available, upgrade recommended!
  eslint-config-next: 15.2.5
  react: 19.1.0
  react-dom: 19.1.0
  typescript: 5.8.3
Next.js Config:
  output: N/A
```

### Additional context

I also tested with @opennextjs/cloudflare@0.6.6 — the result was the same.

For Sentry, I tried both the [manual setup from Sentry’s documentation](https://docs.sentry.io/platforms/javascript/guides/nextjs/manual-setup/) and the server-side configuration mentioned in your [OpenNext AWS docs](https://opennext.js.org/aws/common_issues#sentry-server-side-setup). In both cases, the app still crashes when using preview mode with Wrangler.

conico974:
It is expected that the global timers (i.e. `setInterval`, `setTimeout`) in Node returns a `Nodejs.Timeout` instead of a number. On cloudflare workers they return a number.

This can cause some unexpected crash like in #587, where sentry tries to call `unref()` on the created interval.
This PR fix #587

The only downsides to doing that is that in the middleware we'll have access to the Node version as well, but i don't think this is too big of an issue

### DIFF
--- a/.changeset/famous-weeks-deliver.md
+++ b/.changeset/famous-weeks-deliver.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+global `setTimeout`, `setInterval` and `setImmediate` now use the one from node
--- a/.changeset/famous-weeks-deliver.md
+++ b/.changeset/famous-weeks-deliver.md
@@ -2,4 +2,4 @@
 "@opennextjs/cloudflare": patch
 ---
 
-global `setTimeout`, `setInterval` and `setImmediate` now use the one from node
+global timer functions now use the one from node:timers
--- a/packages/cloudflare/src/cli/build/bundle-server.ts
+++ b/packages/cloudflare/src/cli/build/bundle-server.ts
@@ -146,7 +146,7 @@ export async function bundleServer(buildOpts: BuildOptions): Promise<void> {
     },
     banner: {
       // We need to import them here, assigning them to `globalThis` does not work because node:timers use `globalThis` and thus create an infinite loop
-      js: `import {setInterval, setTimeout, setImmediate} from "node:timers"`,
+      js: `import {setInterval, clearInterval, setTimeout, clearTimeout, setImmediate, clearImmediate} from "node:timers"`,
     },
     platform: "node",
   });
--- a/packages/cloudflare/src/cli/build/bundle-server.ts
+++ b/packages/cloudflare/src/cli/build/bundle-server.ts
@@ -146,6 +146,7 @@ export async function bundleServer(buildOpts: BuildOptions): Promise<void> {
     },
     banner: {
       // We need to import them here, assigning them to `globalThis` does not work because node:timers use `globalThis` and thus create an infinite loop
+      // See https://github.com/cloudflare/workerd/blob/d6a764c/src/node/internal/internal_timers.ts#L56-L70
       js: `import {setInterval, clearInterval, setTimeout, clearTimeout, setImmediate, clearImmediate} from "node:timers"`,
     },
     platform: "node",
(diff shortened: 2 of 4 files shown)


vicb:
Thanks for the fix @conico974 

FYI this was discussed internally and we decided that mixing APIs would not be a problem (documented [here](https://developers.cloudflare.com/workers/runtime-apis/nodejs/timers/)).

I'll raise the issue again internally. 

One possible fix would be to update unenv (with the caveat that external modules would not be polyfilled).

For now I'm +1 for this PR.

packages/cloudflare/src/cli/build/bundle-server.ts
```diff
@@ -144,6 +144,10 @@ export async function bundleServer(buildOpts: BuildOptions): Promise<void> {
       // This define should be safe to use for Next 14.2+, earlier versions (13.5 and less) will cause trouble
       "process.env.__NEXT_EXPERIMENTAL_REACT": `${needsExperimentalReact(nextConfig)}`,
     },
+    banner: {
```
vicb, line 147: We could move the code to `init.ts#initRuntime`

conico974: Already tried that as well. 
I think that the current implementation in `node:timers` internally use `globalThis.setInterval` which make it call itself

vicb: Oh good catch

https://github.com/cloudflare/workerd/blob/d6a764c78360ce7a7905504b5ac1b8ece13db343/src/node/internal/internal_timers.ts#L56-L70

packages/cloudflare/src/cli/build/bundle-server.ts
```diff
@@ -144,6 +144,10 @@ export async function bundleServer(buildOpts: BuildOptions): Promise<void> {
...
+      // We need to import them here, assigning them to `globalThis` does not work because node:timers use `globalThis` and thus create an infinite loop
```
vicb, line 148:
```suggestion
      // We need to import them here, assigning them to `globalThis` does not work because node:timers use `globalThis` and thus create an infinite loop
      // See https://github.com/cloudflare/workerd/blob/d6a764c/src/node/internal/internal_timers.ts#L56-L70
```

vicb (approved):
Thanks!

Added a link to workerd code

.changeset/famous-weeks-deliver.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+global `setTimeout`, `setInterval` and `setImmediate` now use the one from node
```
vicb, line 5: nit: becsause of clear...
```suggestion
global timer functions now use the one from node:timers
```

PR #632, opened 2025-05-05
add e2e for next/head and next/amp in pages router

vicb:
sync @opennextjs/aws#845

Note: the AMP test is disabled because it fails - the root cause seems to be `Error: [unenv] https.request
  is not implemented yet!`


conico974 (approved):
LGTM Thanks

PR #639, opened 2025-05-05
polyfill import.meta.url

conico974:
This PR polyfill `import.meta.url` for packages that may rely on it.
I'm not 100% sure that is what we'd want to end up doing long term.
I feel like this could backfire in the future, a better but more complicated fix would be to replace these values at build time.

### DIFF
--- a/packages/cloudflare/src/cli/templates/init.ts
+++ b/packages/cloudflare/src/cli/templates/init.ts
@@ -94,6 +94,11 @@ function initRuntime() {
     __BUILD_TIMESTAMP_MS__: __BUILD_TIMESTAMP_MS__,
     __NEXT_BASE_PATH__: __NEXT_BASE_PATH__,
   });
+
+  // Some packages rely on `import.meta.url` (i.e. payload) but it is undefined in workerd
+  // It cause a bunch of issue, and will make even import crash
+  // TODO: verify that it won't cause other issues later with other packages
+  import.meta.url = "file:///worker.js";
 }
 
 /**


packages/cloudflare/src/cli/templates/init.ts
```diff
@@ -94,6 +94,11 @@ function initRuntime() {
     __BUILD_TIMESTAMP_MS__: __BUILD_TIMESTAMP_MS__,
     __NEXT_BASE_PATH__: __NEXT_BASE_PATH__,
   });
+
+  // Some packages rely on `import.meta.url` (i.e. payload) but it is undefined in workerd
+  // It cause a bunch of issue, and will make even import crash
+  // TODO: verify that it won't cause other issues later with other packages
+  import.meta.url = "file:///worker.js";
```
vicb, line 101:
```suggestion
  // Some packages rely on `import.meta.url` but it is undefined in workerd
  // For example It cause a bunch of issues, and will make even import crash with payload
  import.meta.url ??= "file:///worker.js";
```

vicb (approved):
We do already have `globalThis.__dirname ??= "";` so I am +1 to add this.

Maybe move this addition at the same location earlier in the file?

Thanks!

packages/cloudflare/src/cli/templates/init.ts
```diff
@@ -61,6 +61,9 @@ function initRuntime() {
 
   globalThis.__dirname ??= "";
   globalThis.__filename ??= "";
+  // Some packages rely on `import.meta.url` but it is undefined in workerd
+  // For example it causes a bunch of issues, and will make even import crash with payload
+  import.meta.url ??= "file:///worker.js";
```
james-elicx, line 66:
```suggestion
  import.meta.url ??= "file://worker.js";
```

is this intentional?

conico974: Yeah that's the default file uri scheme

PR #641, opened 2025-05-06
Feat perf improvement

conico974:
- `enableCacheInterception` is now enabled by default, it loads ISR/SSG pages from cache without waiting for the js page bundle to load
- `routePreloadingBehavior` is now set to `withWaitUntil`, which means a single route js will be lazy loaded on cold start, but other routes will be preloaded using `waitUntil` for better performance

### DIFF
--- a/packages/cloudflare/src/api/config.ts
+++ b/packages/cloudflare/src/api/config.ts
@@ -32,7 +32,7 @@ export type CloudflareOverrides = {
   /**
    * Enable cache interception
    * Disable this if you want to use PPR
-   * @default true
+   * @default false
    */
   enableCacheInterception?: boolean;
 };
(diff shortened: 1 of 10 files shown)


packages/cloudflare/src/api/config.ts
```diff
@@ -28,6 +28,13 @@ export type CloudflareOverrides = {
    * Sets the revalidation queue implementation
    */
   queue?: "direct" | Override<Queue>;
+
+  /**
+   * Enable cache interception
+   * Disable this if you want to use PPR
+   * @default true
```
vicb, line 35: Knowing that it breaks PPR, I think should `@default false` for now
WDYT?

packages/cloudflare/src/api/config.ts
```diff
@@ -28,6 +28,13 @@ export type CloudflareOverrides = {
    * Sets the revalidation queue implementation
    */
   queue?: "direct" | Override<Queue>;
+
+  /**
+   * Enable cache interception
+   * Disable this if you want to use PPR
```
vicb, line 34:
```suggestion
   * Should be `false` when PPR is used
```

PR #644, opened 2025-05-06
fix cache population for R2

conico974:
`remote` used to be the default but should now be specified explicitly.


vicb (approved):
I think we can remove any ref to `excludeRemoteFlag` as it is used only in this case

PR #655, opened 2025-05-08
Feat Automatic cache purge

conico974:
This PR will allow to automatically purge Cache Api entry everywhere where we use them

### DIFF
--- a/packages/cloudflare/src/api/cloudflare-context.ts
+++ b/packages/cloudflare/src/api/cloudflare-context.ts
@@ -54,6 +54,10 @@ declare global {
     // Disable SQLite for the durable object queue handler
     // This can be safely used if you don't use an eventually consistent incremental cache (i.e. R2 without the regional cache for example)
     NEXT_CACHE_DO_QUEUE_DISABLE_SQLITE?: string;
+
+    // Below are the optional env variables for purging the cache
+    CACHE_ZONE_ID?: string;
+    CACHE_API_TOKEN?: string;
   }
 }
 
--- a/packages/cloudflare/src/api/overrides/internal.ts
+++ b/packages/cloudflare/src/api/overrides/internal.ts
@@ -31,11 +31,10 @@ export function computeCacheKey(key: string, options: KeyOptions) {
   return `${prefix}/${buildId}/${hash}.${cacheType}`.replace(/\/+/g, "/");
 }
 
-
 export async function purgeCacheByTags(tags: string[]) {
-  const {env} = getCloudflareContext()
+  const { env } = getCloudflareContext();
 
-  if(!env.CACHE_ZONE_ID && !env.CACHE_API_TOKEN) {
+  if (!env.CACHE_ZONE_ID && !env.CACHE_API_TOKEN) {
     // THIS IS A NO-OP
     debugCache("purgeCacheByTags", "No cache zone ID or API token provided. Skipping cache purge.");
     return;
@@ -44,30 +43,31 @@ export async function purgeCacheByTags(tags: string[]) {
   try {
     const response = await fetch(
       `https://api.cloudflare.com/client/v4/zones/${env.CACHE_ZONE_ID}/purge_cache`,
-    {
-      headers: {
-        "Authorization": `Bearer ${env.CACHE_API_TOKEN}`,
-        "Content-Type": "application/json",
-      },
-      method: "POST",
-      body: JSON.stringify({
-        tags,
-      }),
-    })
+      {
+        headers: {
+          Authorization: `Bearer ${env.CACHE_API_TOKEN}`,
+          "Content-Type": "application/json",
+        },
+        method: "POST",
+        body: JSON.stringify({
+          tags,
+        }),
+      }
+    );
     if (!response.ok) {
       const text = await response.text();
       throw new Error(`Failed to purge cache: ${response.status} ${text}`);
     }
-    const bodyResponse = await response.json() as {
+    const bodyResponse = (await response.json()) as {
       success: boolean;
       errors: Array<{ code: number; message: string }>;
       messages: Array<{ code: number; message: string }>;
-    }
+    };
     if (!bodyResponse.success) {
       throw new Error(`Failed to purge cache: ${JSON.stringify(bodyResponse.errors)}`);
     }
     debugCache("purgeCacheByTags", "Cache purged successfully for tags:", tags);
-  }catch (error) {
+  } catch (error) {
     console.error("Error purging cache by tags:", error);
   }
 }
--- a/packages/cloudflare/src/api/overrides/internal.ts
+++ b/packages/cloudflare/src/api/overrides/internal.ts
@@ -2,6 +2,8 @@ import { createHash } from "node:crypto";
 
 import type { CacheEntryType, CacheValue } from "@opennextjs/aws/types/overrides.js";
 
+import { getCloudflareContext } from "../cloudflare-context.js";
+
 export type IncrementalCacheEntry<CacheType extends CacheEntryType> = {
   value: CacheValue<CacheType>;
   lastModified: number;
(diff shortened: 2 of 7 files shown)


packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -54,6 +54,10 @@ declare global {
     // Disable SQLite for the durable object queue handler
     // This can be safely used if you don't use an eventually consistent incremental cache (i.e. R2 without the regional cache for example)
     NEXT_CACHE_DO_QUEUE_DISABLE_SQLITE?: string;
+
+    // Below are the optional env variables for purging the cache
+    CACHE_ZONE_ID?: string;
+    CACHE_API_TOKEN?: string;
```
vicb, line 60: Suggest: add documentation for each of the variable

packages/cloudflare/src/api/overrides/internal.ts
```diff
@@ -28,3 +30,44 @@ export function computeCacheKey(key: string, options: KeyOptions) {
...
+      }
+    );
+    if (!response.ok) {
+      const text = await response.text();
+      throw new Error(`Failed to purge cache: ${response.status} ${text}`);
+    }
+    const bodyResponse = (await response.json()) as {
+      success: boolean;
+      errors: Array<{ code: number; message: string }>;
+      messages: Array<{ code: number; message: string }>;
```
vicb, line 64: unused?

packages/cloudflare/src/api/cloudflare-context.ts
conico974: Yeah, i'm still working on it.
I'm testing a few things especially around Cache purge limits https://developers.cloudflare.com/cache/how-to/purge-cache/#hostname-tag-prefix-url-and-purge-everything-limits
We could add a DO to handle these cases, i think the easiest way would be to just add a buffer DO.
Given how this is used, i think you'd need a custom solution anyway if you exceeds the provided req/s regularly.

conico974 (commented):
I'm not entirely sure about naming here. If you have better ideas

packages/cloudflare/src/api/durable-objects/bucket-cache-purge.ts
```diff
@@ -0,0 +1,85 @@
+import { DurableObject } from "cloudflare:workers";
+
+import { internalPurgeCacheByTags } from "../overrides/internal";
+
+const DEFAULT_BUFFER_TIME = 5; // seconds
```
vicb, line 5:
```suggestion
const DEFAULT_BUFFER_TIME_SECONDS = 5;
```

packages/cloudflare/src/api/durable-objects/bucket-cache-purge.ts
```diff
@@ -0,0 +1,85 @@
...
+      this.ctx.storage.setAlarm(Date.now() + this.bufferTimeInSeconds * 1000);
+    }
+  }
+
+  override async alarm() {
+    let tags = this.ctx.storage.sql
+      .exec<{ tag: string }>(
+        `
+      SELECT * FROM cache_purge LIMIT 100
+    `
```
vicb, line 49: Does 100 comes from the free plan limit?
Maybe add a constant with a comment
```suggestion
        "SELECT * FROM cache_purge LIMIT 100"
```

packages/cloudflare/src/api/overrides/internal.ts
```diff
@@ -28,3 +30,61 @@ export function computeCacheKey(key: string, options: KeyOptions) {
...
+    debugCache("purgeCacheByTags", "No cache zone ID or API token provided. Skipping cache purge.");
+    return;
+  }
+
+  try {
+    const response = await fetch(
+      `https://api.cloudflare.com/client/v4/zones/${env.CACHE_PURGE_ZONE_ID}/purge_cache`,
+      {
+        headers: {
+          Authorization: `Bearer ${env.CACHE_PURGE_ZONE_ID}`,
```
vicb, line 66:
```suggestion
          Authorization: `Bearer ${env.CACHE_PURGE_API_TOKEN}`,
```

packages/cloudflare/src/api/durable-objects/bucket-cache-purge.ts
```diff
@@ -0,0 +1,86 @@
...
+      );
+      // Delete the tags from the sql table
+      this.ctx.storage.sql.exec(
+        `
+        DELETE FROM cache_purge
+        WHERE tag IN (${tags.map(() => "?").join(",")})
+      `,
+        tags.map((row) => row.tag)
+      );
+      if (tags.length < 100) {
```
vicb, line 71:
```suggestion
      if (tags.length < MAX_NUMBER_OF_TAGS_PER_PURGE) {
```

packages/cloudflare/src/api/durable-objects/bucket-cache-purge.ts
```diff
@@ -0,0 +1,86 @@
...
+        // If we have less than 100 tags, we can stop
```
vicb, line 72:
```suggestion
        // If we have less than MAX_NUMBER_OF_TAGS_PER_PURGE tags, we can stop
```

packages/cloudflare/src/api/durable-objects/bucket-cache-purge.ts
```diff
@@ -0,0 +1,86 @@
...
+        `
+      SELECT * FROM cache_purge LIMIT ${MAX_NUMBER_OF_TAGS_PER_PURGE}
+    `
+      )
+      .toArray();
+    if (tags.length === 0) {
+      // No tags to purge, we can stop
+      // It shouldn't happen, but just in case
+      return;
+    }
```
vicb, line 57: Maybe you can move this at the beginning of the loop and adjust the while condition to avoid duplication?

packages/cloudflare/src/api/durable-objects/bucket-cache-purge.ts
```diff
@@ -0,0 +1,86 @@
...
+    do {
+      await internalPurgeCacheByTags(
+        this.env,
+        tags.map((row) => row.tag)
+      );
+      // Delete the tags from the sql table
+      this.ctx.storage.sql.exec(
```
vicb, line 64: Do we want to delete only if `internalPurgeCacheByTags` was successful?

conico974: I guess we could do that, but it means we have to keep every single tag, it doesn't look like we can know if it only fails partially

conico974: Actually keeping them might cause an issue, we could end up in a loop trying to purge something that somehow can't get purged. And if it happens it could end up blocking a lot of other tag from being invalidated

vicb: I tested the cache purge API:

When there is an error, it could 200 with:

```json
{
  success: false,
  errors: [
    {
      code: 1100,
      message: 'Tag exceeds maximum length of 1024 characters: ....'
    },
    { code: 1016, message: 'One or more cache purge errors' }
  ],
  messages: [],
  result: null
}

{
  success: false,
  errors: [
    {
      code: 1315,
      message: 'Non-printable characters are forbidden in tags'
    },
    { code: 1016, message: 'One or more cache purge errors' }
  ],
  messages: [],
  result: null
}
```

In this case we should delete the tags.

It could also be rate limited (HTTP Status 429)

```
{
  success: false,
  errors: [
    {
      code: 1134,
      message: 'Unable to purge, rate limit reached. Please wait and consider throttling your request speed'
    }
  ],
  messages: [],
  result: null
}
```

For this particular case, we should bail out and not delete the tags

vicb (commented):
LGTM

Could you please add a changeset (minor)

vicb (approved):
Thanks!

PR #656, opened 2025-05-08
Revert to external middleware

conico974:
This is just a proposal for now, but this is the best way that i can think of to really reduce cold start for ISR/SSG (in combination with `enableCacheInterception`)

By using that we can go from 600-800ms of cpu time for ISR/SSG during cold start to around 30-40ms. Tested with the app-router e2e app with `enableCacheInterception` and cache purge as well.

This has the drawback of increasing the size of the app.

We should also think if we want to move back the main handler import at the top as well. This could reduce cpu time for SSR route themselves, but could cause issue with thing that are loaded right away like instrumentation (especially if there is I/O)

### DIFF
--- a/packages/cloudflare/src/cli/templates/init.ts
+++ b/packages/cloudflare/src/cli/templates/init.ts
@@ -96,6 +96,7 @@ function initRuntime() {
     Request: CustomRequest,
     __BUILD_TIMESTAMP_MS__: __BUILD_TIMESTAMP_MS__,
     __NEXT_BASE_PATH__: __NEXT_BASE_PATH__,
+    __dangerous_ON_edge_converter_returns_request: true,
   });
 }
 
--- a/packages/cloudflare/src/cli/templates/init.ts
+++ b/packages/cloudflare/src/cli/templates/init.ts
@@ -96,6 +96,9 @@ function initRuntime() {
     Request: CustomRequest,
     __BUILD_TIMESTAMP_MS__: __BUILD_TIMESTAMP_MS__,
     __NEXT_BASE_PATH__: __NEXT_BASE_PATH__,
+    // The external middleware will use the convertTo function of the `edge` converter
+    // by default it will try to fetch the request, but since we are running everything in the same worker
+    // we need to use the request as is.
     __dangerous_ON_edge_converter_returns_request: true,
   });
 }
--- a/packages/cloudflare/src/cli/templates/worker.ts
+++ b/packages/cloudflare/src/cli/templates/worker.ts
@@ -1,5 +1,7 @@
 //@ts-expect-error: Will be resolved by wrangler build
 import { runWithCloudflareRequestContext } from "./cloudflare/init.js";
+// @ts-expect-error: Will be resolved by wrangler build
+import { handler as middlewareHandler } from "./middleware/handler.mjs";
 
 //@ts-expect-error: Will be resolved by wrangler build
 export { DOQueueHandler } from "./.build/durable-objects/queue.js";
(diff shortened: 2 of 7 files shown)


packages/cloudflare/src/cli/templates/init.ts
```diff
@@ -96,6 +96,7 @@ function initRuntime() {
     Request: CustomRequest,
     __BUILD_TIMESTAMP_MS__: __BUILD_TIMESTAMP_MS__,
     __NEXT_BASE_PATH__: __NEXT_BASE_PATH__,
+    __dangerous_ON_edge_converter_returns_request: true,
```
vicb, line 99: Looking back at https://github.com/opennextjs/opennextjs-cloudflare/pull/563, should the value differ if the middleware is ext/int?

packages/cloudflare/src/cli/templates/worker.ts
```diff
@@ -32,10 +34,17 @@ export default {
...
 
+      // - `Request`s are handled by the Next server
+      const reqOrResp = await middlewareHandler(request, env, ctx);
+
+      if (reqOrResp instanceof Response) {
+        return reqOrResp;
+      }
+
       // @ts-expect-error: resolved by wrangler build
       const { handler } = await import("./server-functions/default/handler.mjs");
```
vicb, line 45: Thought: The code is always parsed wether or not we return on line 41. Meaning that the perf difference could be explained by the global init. To have the init done only once we could have a top level `serverHandler` and have

```ts
serverHandler ??= (await import("./server-functions/default/handler.mjs")).handler;
```

We probably want something similar for the `middlewareHandler` - we the current top level handler, we would have the same issues as with the server handler: `process.env` not being initialized at that point and not possible to do IO at top level.

packages/cloudflare/src/cli/templates/init.ts
conico974: It isn't used when the middleware is internal, when that's the case we don't use the converter `convertTo` but the `StreamCreator`.
Middleware (or any edge function) directly returns a `Response` that we can convert right away

packages/cloudflare/src/cli/templates/worker.ts
conico974: > We probably want something similar for the middlewareHandler

That's not an issue for the middleware, we already import it using an awaited import https://github.com/opennextjs/opennextjs-aws/blob/c2da3a8df638b8facdb896e22538a143b3425006/packages/open-next/src/core/routing/middleware.ts#L38

> The code is always parsed wether or not we return on line 41. Meaning that the perf difference could be explained by the global init.

I think the problem is not with the parsing itself, but with the v8 compilation. The whole file is parsed, but not all of it is compiled.

I did a simple experiment of returning a silly Response just after the await import and it added barely 100ms of cpu time, compared to the minimum 400 that i would get when returning something from the middleware layer (a redirect or any response from the middleware)

I did an experiment with v8 compile cache in aws and by precompiling the files, you can easily shave 200/300ms, but it did not fully remove the cold start. Things that are imported/required only inside a function doesn't get compiled until the function is actually called.
And i think that when the middleware is bundled, it is actually inside a big function that also contains every other `NextServer` related things, which probably explains the difference

packages/cloudflare/src/cli/templates/init.ts
vicb: Could you please add a comment?

conico974:
I'll fix it and rebase on main then.
One thing i think we should think about is relaxing some of the requirement in `ensure-cf-config`, maybe just show a warning instead of throwing an error. It could be useful for people that would want more customization

vicb:
> One thing i think we should think about is relaxing some of the requirement in `ensure-cf-config`, maybe just show a warning instead of throwing an error. It could be useful for people that would want more customization

+1, We could use `cloudflare.dangerousDisableConfigValidation` to switch form error to warnings

PR #666, opened 2025-05-15
Fix fetch inside `use cache` in ISR

conico974:
Using fetch (or I/O in general) inside of a `use cache` in an ISR/SSG route would make it crash with a `Cannot perform I/O on behalf of a different request` error during revalidation request if the instance is warm.

This is because by default it will use `AsyncLocalStorage.snapshot()` and it will bind everything to the initial request context. The downsides is that `use cache` function will have access to the full request ALS context from next (i.e. cookies, headers ...)

### DIFF
--- a/packages/cloudflare/src/cli/build/patches/plugins/use-cache.ts
+++ b/packages/cloudflare/src/cli/build/patches/plugins/use-cache.ts
@@ -3,6 +3,10 @@
  * server/app-render/async-local-storage.js file to an empty string.
  * This is necessary because the createSnapshot function is causing I/O issues for
  * ISR/SSG revalidation in Cloudflare Workers.
+ * This is because by default it will use AsyncLocalStorage.snapshot() and it will
+ * bind everything to the initial request context.
+ * The downsides is that use cache function will have access to the full request
+ * ALS context from next (i.e. cookies, headers ...)
  * TODO: Find a better fix for this issue.
  */
 import { patchCode } from "@opennextjs/aws/build/patch/astCodePatcher.js";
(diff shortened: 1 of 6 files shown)


conico974:
> Do you understand why we "perform I/O on behalf of a different request"?

I think i do, basically a function using `use cache` will be wrapped in this https://github.com/vercel/next.js/blob/62de494f3c28e3a6e949a434047ea3023121d239/packages/next/src/server/app-render/async-local-storage.ts#L55-L65. This is done only once at import time (so on the first request to the instance), subsequent request will then get bound to this initial AsyncLocalStorage context which is associated with the first request.

It might be possible to fix it at the workerd level so that ALS `bind` or `snapshot` is not actually bound to the request context, but this is likely tricky.

packages/cloudflare/src/cli/build/patches/plugins/use-cache.ts
```diff
@@ -0,0 +1,39 @@
+/**
+ * This patch will replace the createSnapshot function in the
+ * server/app-render/async-local-storage.js file to an empty string.
+ * This is necessary because the createSnapshot function is causing I/O issues for
+ * ISR/SSG revalidation in Cloudflare Workers.
+ * TODO: Find a better fix for this issue.
+ */
```
vicb, line 7: Could you please add more details from the PR discussion:

https://github.com/vercel/next.js/blob/62de494f/packages/next/src/server/app-render/async-local-storage.ts#L55-L65

> This is because by default it will use AsyncLocalStorage.snapshot() and it will bind everything to the initial request context. The downsides is that use cache function will have access to the full request ALS context from next (i.e. cookies, headers ...)

vicb (approved):
Thanks for the fix!

PR #674, opened 2025-05-20
Fix `blockConcurrencyWhile` in DO queue

conico974:
Right now there is an issue with `blockConcurrencyWhile` that will make it block for 30s and fail.
When there is too much revalidate request at a fast enough pace, it seems that there is multiple call to `blockConcurrencyWhile` and they seem to block each other.

This PR fixes it by removing the `blockConcurrencyWhile` and just waiting for `ongoingRevalidation` to go below `maxRevalidate`

### DIFF
--- a/packages/cloudflare/src/api/durable-objects/queue.ts
+++ b/packages/cloudflare/src/api/durable-objects/queue.ts
@@ -87,14 +87,14 @@ export class DOQueueHandler extends DurableObject<CloudflareEnv> {
       debug(
         `The maximum number of revalidations (${this.maxRevalidations}) is reached. Blocking until one of the revalidations finishes.`
       );
-      const ongoingRevalidations = this.ongoingRevalidations.values();
-      // When there is more than the max revalidations, we block concurrency until one of the revalidations finishes
-      // We still await the promise to ensure the revalidation is completed
-      // This is fine because the queue itself run inside a waitUntil
-      await this.ctx.blockConcurrencyWhile(async () => {
+      // TODO: need more investigation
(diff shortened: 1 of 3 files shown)


packages/cloudflare/src/api/durable-objects/queue.ts
```diff
@@ -84,25 +84,29 @@ export class DOQueueHandler extends DurableObject<CloudflareEnv> {
...
+      }
       debug(
         `The maximum number of revalidations (${this.maxRevalidations}) is reached. Blocking until one of the revalidations finishes.`
       );
-      const ongoingRevalidations = this.ongoingRevalidations.values();
-      // When there is more than the max revalidations, we block concurrency until one of the revalidations finishes
-      // We still await the promise to ensure the revalidation is completed
-      // This is fine because the queue itself run inside a waitUntil
-      await this.ctx.blockConcurrencyWhile(async () => {
+      // TODO: need more investigation
```
vicb, line 95: It would be great if you can have a minimal repro and create an issue on the workerd repro and link it here

conico974: I'll try to work on that. It's a bit tricky to reproduce locally

packages/cloudflare/src/api/durable-objects/queue.ts
```diff
@@ -73,6 +73,11 @@ export class DOQueueHandler extends DurableObject<CloudflareEnv> {
   }
 
   async revalidate(msg: QueueMessage) {
+    if (this.ongoingRevalidations.size > 2 * this.maxRevalidations) {
+      console.warn(
```
vicb, line 77: import warn?

PR #689, opened 2025-05-29
fix(preview): Ensure the protocol is correct in redirects from server actions during preview mode

### LINKED ISSUE #686: [BUG] NextAuth signOut request takes 15 seconds in preview mode
### Describe the bug

If you make a sign-out button with Auth.js/NextAuth and then use it while running your project with `npm run preview` (which uses Cloudflare wrangler and supports durable objects with rpc unlike `npm run dev`), the request to sign you out takes 15 seconds to complete, and this error shows up in the terminal:

```
✘ [ERROR] failed to get redirect response Error: Network connection lost.

      at async tz (home/macil/Coding/next-auth-test/.wrangler/tmp/dev-KJM7H1/worker.js:230571:29)
      at async tX (home/macil/Coding/next-auth-test/.wrangler/tmp/dev-KJM7H1/worker.js:230713:150)
      at async r9 (home/macil/Coding/next-auth-test/.wrangler/tmp/dev-KJM7H1/worker.js:231643:29)
      at async doRender
  (home/macil/Coding/next-auth-test/.wrangler/tmp/dev-KJM7H1/worker.js:127279:30)
      at async NextNodeServer.renderToResponseWithComponentsImpl
  (home/macil/Coding/next-auth-test/.wrangler/tmp/dev-KJM7H1/worker.js:127494:32)
      at async NextNodeServer.renderPageComponent
  (home/macil/Coding/next-auth-test/.wrangler/tmp/dev-KJM7H1/worker.js:127902:24)
      at async NextNodeServer.renderToResponseImpl
  (home/macil/Coding/next-auth-test/.wrangler/tmp/dev-KJM7H1/worker.js:127937:32)
      at async NextNodeServer.pipeImpl
  (home/macil/Coding/next-auth-test/.wrangler/tmp/dev-KJM7H1/worker.js:126742:29)
      at async NextNodeServer.handleCatchallRenderRequest
  (home/macil/Coding/next-auth-test/.wrangler/tmp/dev-KJM7H1/worker.js:241609:17)
      at async NextNodeServer.handleRequestImpl
  (home/macil/Coding/next-auth-test/.wrangler/tmp/dev-KJM7H1/worker.js:126633:17) {
    retryable: true
  }


[wrangler:info] POST / 303 See Other (15019ms)
```

This issue does not affect applications in production deployed to Cloudflare Workers or run locally with `npm run dev`. I'm a little surprised that `npm run preview` is the odd one out since generally it seems like `npm run preview` is the development option that works most closely to Cloudflare Workers.

### Steps to reproduce

A simple example project reproducing this bug is at https://github.com/Macil/next-auth-test. The project is a fresh OpenNext Cloudflare template with a straight-forward usage of NextAuth from following NextAuth's documentation. I've kept the git history clean so you can easily see exactly what was done.

The `signOut()` function is called in a server action in [`components/SignOut.tsx`](https://github.com/Macil/next-auth-test/blob/main/components/SignOut.tsx), following the example given in NextAuth's docs at https://authjs.dev/getting-started/session-management/login.

### Expected behavior

Calling NextAuth's `signOut()` function in a server action should be quick and not log any errors when used under `npm run preview`.

1.0.4

### Wrangler version

4.17.0

### next info output

```shell
Operating System:
  Platform: linux
  Arch: x64
  Version: #1 SMP Tue Nov 5 00:21:55 UTC 2024
  Available memory (MB): 31694
  Available CPU cores: 32
Binaries:
  Node: 23.6.1
  npm: 11.0.0
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  next: 15.3.2 // Latest available version is detected (15.3.2).
  eslint-config-next: 15.3.2
  react: 19.1.0
  react-dom: 19.1.0
  typescript: 5.8.3
Next.js Config:
  output: N/A
```

### Additional context

It might be worth noting that there is a separate OpenNext/NextAuth bug affecting signout (https://github.com/opennextjs/opennextjs-cloudflare/issues/606) that this repository has a workaround in place for in [`open-next.config.ts`](https://github.com/Macil/next-auth-test/blob/main/open-next.config.ts). I do not think that bug is directly related to the 15 second delay but I might be wrong. The delay is present with or without this workaround.

sommeeeer:
Closes #686. 

**Important** to note this didn't only affect `next-auth`. All redirects from server actions was affected in preview mode. `wrangler dev` would hang and give an error as it would try to send a request to `https`. 

Through my research I figured out that Next internally uses `process.env.__NEXT_PRIVATE_ORIGIN` in standalone mode. They set it during `startServer` on this [line](https://github.com/vercel/next.js/blob/6b1e48080e896e0d44a05fe009cb79d2d3f91774/packages/next/src/server/lib/start-server.ts#L275). To align with this behavior in preview mode this should be set to the origin during the initialization. 

Affected lines in Next:
- https://github.com/vercel/next.js/blob/6b1e48080e896e0d44a05fe009cb79d2d3f91774/packages/next/src/server/app-render/action-handler.ts#L177-L184
- https://github.com/vercel/next.js/blob/6b1e48080e896e0d44a05fe009cb79d2d3f91774/packages/next/src/server/app-render/action-handler.ts#L307-L308

`proto` would always be `https` here, thus the `fetch` below would recieve the wrong protocol in its `fetchUrl`.

### DIFF
--- a/packages/cloudflare/src/cli/templates/init.ts
+++ b/packages/cloudflare/src/cli/templates/init.ts
@@ -132,6 +132,14 @@ function populateProcessEnv(url: URL, env: CloudflareEnv) {
       port: url.port,
     },
   });
+
+  /* We need to set this environment variable to make redirects work properly in preview mode.
+   * Next sets this in standalone mode during `startServer`. Without this the protocol would always be `https` here:
+   * https://github.com/vercel/next.js/blob/6b1e48080e896e0d44a05fe009cb79d2d3f91774/packages/next/src/server/app-render/action-handler.ts#L307-L316
+   */
+  if (url.hostname === "localhost") {
+    process.env.__NEXT_PRIVATE_ORIGIN = url.origin;
+  }
 }
 
 /* eslint-disable no-var */
--- a/packages/cloudflare/src/cli/templates/init.ts
+++ b/packages/cloudflare/src/cli/templates/init.ts
@@ -137,9 +137,7 @@ function populateProcessEnv(url: URL, env: CloudflareEnv) {
    * Next sets this in standalone mode during `startServer`. Without this the protocol would always be `https` here:
    * https://github.com/vercel/next.js/blob/6b1e48080e896e0d44a05fe009cb79d2d3f91774/packages/next/src/server/app-render/action-handler.ts#L307-L316
    */
-  if (url.hostname === "localhost") {
-    process.env.__NEXT_PRIVATE_ORIGIN = url.origin;
-  }
+  process.env.__NEXT_PRIVATE_ORIGIN = url.origin;
 }
 
 /* eslint-disable no-var */
--- a/.changeset/hip-bags-argue.md
+++ b/.changeset/hip-bags-argue.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+fix: Ensure the protocol is correct for redirects in server actions during preview mode


vicb (commented):
Thanks for the fix!

Could we drop the condition `url.hostname === "localhost"` and always set `process.env.__NEXT_PRIVATE_ORIGIN`?

sommeeeer:
> Could we drop the condition `url.hostname === "localhost"` and always set `process.env.__NEXT_PRIVATE_ORIGIN`?

In theory that shouldn't be a problem. `fetchUrl` computed by Next would be correct if the protocol is `https`. This fix is only necessary when the protocol is `http`. Its up to you.

packages/cloudflare/src/cli/templates/init.ts
```diff
@@ -132,6 +132,14 @@ function populateProcessEnv(url: URL, env: CloudflareEnv) {
...
     },
   });
+
+  /* We need to set this environment variable to make redirects work properly in preview mode.
+   * Next sets this in standalone mode during `startServer`. Without this the protocol would always be `https` here:
+   * https://github.com/vercel/next.js/blob/6b1e48080e896e0d44a05fe009cb79d2d3f91774/packages/next/src/server/app-render/action-handler.ts#L307-L316
+   */
+  if (url.hostname === "localhost") {
+    process.env.__NEXT_PRIVATE_ORIGIN = url.origin;
+  }
```
vicb: Let's keep the change simple and consistent across dev/prod.

```suggestion
  process.env.__NEXT_PRIVATE_ORIGIN = url.origin;
```

vicb (approved):
LGTM, thanks!

I have updated the code to unconditionally apply the fix.

PR #696, opened 2025-06-03
Add the `routePreloadingBehavior` to the options of `defineCloudflareConfig`

vicb:
@conico974 what do you think about this PR?

When investigating perf issue with https://github.com/LubomirGeorgiev/cloudflare-workers-nextjs-saas-template, I measure the preload to be 200-300ms.

This change could save some CPU cycles.

### DIFF
--- a/packages/cloudflare/src/api/config.ts
+++ b/packages/cloudflare/src/api/config.ts
@@ -48,10 +48,8 @@ export type CloudflareOverrides = {
   enableCacheInterception?: boolean;
 
   /**
-   * Set the route preloading behavior.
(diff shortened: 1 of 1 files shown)


conico974 (approved):
Yeah i think it's a quick fix, but ideally, we should move the preload in the `waitUntil` after the request is done. This way we'll barely consume any cpu cycle from the current request and really execute everything in the background.
It should be done in AWS though

james-elicx (approved):

vicb:
Thanks for the reviews!

PR #702, opened 2025-06-04
perf: drop `babel` to reduce the server bundle size

vicb:
Next 15 pulls ~4MB of Babel dependency. 

Adding a patch to drop that.

See the effect on the `playground15` example

Before

image

After

image

### DIFF
--- a/.changeset/fast-spiders-hammer.md
+++ b/.changeset/fast-spiders-hammer.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+perf: drop `babel` to reduce the server bundle size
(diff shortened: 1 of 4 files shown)


conico974 (commented):
I think we'd want this to be in aws. It has far less effect there, but it's still something we'd want removed directly there

vicb:
> I think we'd want this to be in aws. It has far less effect there, but it's still something we'd want removed directly there

I could work on moving it there.

Don't you use `runEdgeFunction` though?

edit: probably not actually as you run the edge functions on Cloudflare

vicb:
Moved to https://github.com/opennextjs/opennextjs-aws/pull/891

PR #706, opened 2025-06-04
refactor: move cache patches to the Next Server astgrep based patches

vicb:

PR #705, opened 2025-06-04
refactor: group patches for next-server.js

vicb:
Group patches for `next-server.js` - inlining the middleware manifest is migrated to astgrep.

---
This is **part 1 of 3 in a stack** made with GitButler:
- <kbd>&nbsp;3&nbsp;</kbd> #707 
- <kbd>&nbsp;2&nbsp;</kbd> #706 
- <kbd>&nbsp;1&nbsp;</kbd> #705 👈

PR #707, opened 2025-06-04
refactor: convert require.resolve patches to astgrep

vicb:
---
This is **part 3 of 3 in a stack** made with GitButler:
- <kbd>&nbsp;3&nbsp;</kbd> #707 👈 
- <kbd>&nbsp;2&nbsp;</kbd> #706 
- <kbd>&nbsp;1&nbsp;</kbd> #705


conico974 (approved):
LGTM, i think in the future we should find a better solution than this.
It's weird to have to patch OpenNext itself

PR #710, opened 2025-06-05
perf: drop babel to reduce the server bundle size

vicb:
Integrate https://github.com/opennextjs/opennextjs-aws/pull/891 (supersedes #702).

~TODO: Update the aws version after 891 is released~


vicb:
Thanks Nico!

PR #712, opened 2025-06-05
[TODO] Drop the patch for NextServer#getMiddlewareManifest

vicb:
There is no need to patch this unused method.

We can rely on aws patching the method to return null. It will patch regardless of the version when https://github.com/opennextjs/opennextjs-aws/pull/892 gets merged. 

TODO:

- [ ] update package.json after opennextjs/opennextjs-aws#892 is released

### DIFF
--- a/.changeset/wild-readers-remain.md
+++ b/.changeset/wild-readers-remain.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+Drop the patch for NextServer#getMiddlewareManifest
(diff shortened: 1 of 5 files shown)


vicb:
Just rebased the PR on HEAD.
It is ready for review.
Thanks!

.changeset/wild-readers-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+Drop the patch for NextServer#getMiddlewareManifest
```
james-elicx: Should we mention that it was moved to aws instead of dropped completely?

vicb: Good point, I will clarify.

To be exact aws has been patching the body to `return null;` for some time but we were still applying a patch to inline the content of the manifest.

But this method is not called in OpenNext so we stop patching here and rely on the `return null`

Thanks for the review

vicb:
```suggestion
Drop the patch for NextServer#getMiddlewareManifest (taken care of by was)
```

PR #713, opened 2025-06-06
Make the require patch optional

vicb:
The "require" patch could be mark as optional - It triggered an error when I was playing with minification settings.

Also:
- Drop a debug `console.error`
- Make `patchCodeWithValidations` less verbose as it is now less used.


conico974 (approved):

vicb:
Thanks Nico!

PR #714, opened 2025-06-06
perf: minify the generated bundle by default

vicb:
Turns on minifying by default as bundle size could have an impact on performance.

This PR stays on the safe side by only minifying whitespaces and syntax. Identifier are explicitly not renamed to avoid runtime issues.

Some result experimenting with a sample app.

The original code size was 7.7MB before bundling.

Before this PR, the output would be 10.2MB
Using the most agressive minification (`minify: true`) -> 6.7MB
Using the safe minification implemented in this PR -> 7.4MB

That a close to 30% improvement.

### DIFF
--- a/.changeset/giant-books-dream.md
+++ b/.changeset/giant-books-dream.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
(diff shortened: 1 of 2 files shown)


vicb:
Thanks Nico!

PR #720, opened 2025-06-09
feat: pass cli arguments not used by opennext to wrangler

james-elicx:


.changeset/six-moons-shine.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+feat: pass cli arguments not used by opennext to wrangler
```
vicb:
```suggestion
feat: pass cli arguments not used by `opennextjs-cloudflare` to wrangler
```

packages/cloudflare/src/cli/args.ts
```diff
@@ -25,32 +26,33 @@ export type Arguments = (
     }
 ) & { outputDir?: string };
 
+const config = {
```
vicb, line 30: Could you please add a one line comment on what this is.

james-elicx, line 30:
```suggestion
// Config for parsing CLI arguments
const config = {
```

packages/cloudflare/src/cli/args.ts
```diff
@@ -25,32 +26,33 @@ export type Arguments = (
...
-  const passthroughArgs = getPassthroughArgs();
-
   switch (positionals[0]) {
     case "build":
       return {
         command: "build",
         outputDir,
         skipNextBuild:
-          values.skipBuild || ["1", "true", "yes"].includes(String(process.env.SKIP_NEXT_APP_BUILD)),
+          !!values.skipBuild || ["1", "true", "yes"].includes(String(process.env.SKIP_NEXT_APP_BUILD)),
```
vicb, line 54: `!!` should not be needed as we have default value ?

packages/cloudflare/src/cli/args.ts
```diff
@@ -81,9 +84,27 @@ export function getArgs(): Arguments {
...
-  return passthroughPos === -1 ? [] : process.argv.slice(passthroughPos + 1);
+export function getPassthroughArgs<T extends ParseArgsConfig>(args: string[], { options = {} }: T) {
+  const passthroughArgs: string[] = [];
+
+  args.forEach((fullArg, idx) => {
+    const [, name] = /^--?(\w[\w-_]*)(=.+)?$/.exec(fullArg) ?? [];
+    if (name && !(name in options)) {
+      passthroughArgs.push(fullArg);
+
+      for (let i = idx + 1; i < args.length; i++) {
```
vicb: not sure to understand why we need this nested loop?

packages/cloudflare/src/cli/args.ts
james-elicx: Disabling strict mode to stop failures on invalid params changes the return type to be less strict

packages/cloudflare/src/cli/args.ts
james-elicx: This was to append strings that follow the flag

vicb: I think what make this hard to follow is that the args processed in this loop will be processed again in the outer loop.

I think there should also be an early exit when `--` is found (and following args should be pushed to `passthroughArgs`).

So I'm mostly ok with the behavior but I think the implementation could be more readable.

james-elicx: I've changed it to use a for loop, with an inner while loop which increments the index in the for loop, as well as adding the early return

packages/cloudflare/src/cli/args.ts
```diff
@@ -81,9 +84,26 @@ export function getArgs(): Arguments {
...
+    if (args[i] === "--") {
+      passthroughArgs.push(...args.slice(i + 1));
+      return passthroughArgs;
+    }
+
+    const [, name] = /^--?(\w[\w-_]*)(=.+)?$/.exec(args[i]!) ?? [];
+    if (name && !(name in options)) {
+      passthroughArgs.push(args[i]!);
+
+      while (!args[i + 1]?.startsWith("-")) {
```
vicb: Thanks for the update. 
I find the new version much more readable.

Still wondering why we have a while here.

I get the `--flagName flagValue` thing but why more strings after `flagValue` ?

james-elicx: i was assuming array arguments might be able to be provided without specifying the key each time, e.g. `--flagName flagValue1 flagValue2`. yargs calls it greedy arrays - https://github.com/yargs/yargs-parser/blob/main/README.md#greedy-arrays

james-elicx: @vicb are you okay with that?

packages/cloudflare/src/cli/args.ts
```diff
@@ -81,9 +84,26 @@ export function getArgs(): Arguments {
...
+export function getPassthroughArgs<T extends ParseArgsConfig>(args: string[], { options = {} }: T) {
+  const passthroughArgs: string[] = [];
+
+  for (let i = 0; i < args.length; i++) {
+    if (args[i] === "--") {
+      passthroughArgs.push(...args.slice(i + 1));
+      return passthroughArgs;
+    }
+
+    const [, name] = /^--?(\w[\w-_]*)(=.+)?$/.exec(args[i]!) ?? [];
```
vicb: adding a comment, also `\w` includes `_`
```suggestion
    // look for `--arg(=value)`, `-arg(=value)`    
    const [, name] = /^--?(\w[\w-]*)(=.+)?$/.exec(args[i]!) ?? [];
```

packages/cloudflare/src/cli/args.ts
vicb: Adding a comment and preventing an infinite loop at the end
```suggestion
      // Array args can have multiple values  
      // ref https://github.com/yargs/yargs-parser/blob/main/README.md#greedy-arrays     
      while (i < args.length - 1 && !args[i + 1]?.startsWith("-")) {
```

.changeset/six-moons-shine.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+feat: pass cli arguments not used by `opennextjs-cloudflare` to wrangler
```
vicb, line 5: Could you add an example of the new allowed syntax vs the previous syntax.

vicb (approved):
LGTM with a few minor comments. Thanks!

(and sorry it totally slipped my mind)

.changeset/six-moons-shine.md
james-elicx:
```suggestion
feat: pass cli arguments not used by `opennextjs-cloudflare` to wrangler

Previously, arguments had to be provided after `--` e.g. `opennextjs-cloudflare preview -- --port 12345`. This is no longer necessary, and they can be provided normally, e.g. `opennextjs-cloudflare preview --port 12345`.
```

vicb:
Thanks James, that will ease the life of many users!

Issue #661, opened 2025-05-12
[BUG] Error Could not resolve when npm run deploy


ton11797:
### Describe the bug

Error when try to run npm run deploy

```
Next.js version : 15.3.1
@opennextjs/cloudflare version: 1.0.2
@opennextjs/aws version: ^3.6.1

┌─────────────────────────────────┐
│ OpenNext — Building Next.js app │
└─────────────────────────────────┘


> cf-bug@0.1.0 build
> next build

Using vars defined in .dev.vars
   ▲ Next.js 15.3.1
   - Environments: .env

Using vars defined in .dev.vars
   Creating an optimized production build ...
Using vars defined in .dev.vars
Using vars defined in .dev.vars
Using vars defined in .dev.vars
 ✓ Compiled successfully in 0ms
 ✓ Linting and checking validity of types    
 ✓ Collecting page data    
 ✓ Generating static pages (6/6)
 ✓ Collecting build traces    
 ✓ Finalizing page optimization    

Route (app)                                 Size  First Load JS
┌ ○ /                                    5.63 kB         107 kB
├ ○ /_not-found                            977 B         102 kB
└ ƒ /test                                  136 B         101 kB
+ First Load JS shared by all             101 kB
  ├ chunks/4bd1b696-67ee12fb04071d3b.js  53.2 kB
  ├ chunks/684-71b0b1461a73105d.js       45.9 kB
  └ other shared chunks (total)          1.96 kB


○  (Static)   prerendered as static content
ƒ  (Dynamic)  server-rendered on demand


┌──────────────────────────────┐
│ OpenNext — Generating bundle │
└──────────────────────────────┘

Bundling middleware function...
Bundling static assets...
Bundling cache assets...
Building server function: default...
Applying code patches: 3.695s
# copyPackageTemplateFiles
⚙️ Bundling the OpenNext server...

X [ERROR] Could not resolve "@libsql/isomorphic-ws"

    .open-next/server-functions/default/node_modules/@libsql/hrana-client/lib-esm/index.js:1:26:
      1 │ import { WebSocket } from "@libsql/isomorphic-ws";
        ╵                           ~~~~~~~~~~~~~~~~~~~~~~~

  The module "./web.mjs" was not found on the file system:

    .open-next/server-functions/default/node_modules/@libsql/isomorphic-ws/package.json:24:19:
      24 │         "workerd": "./web.mjs",
         ╵                    ~~~~~~~~~~~

[...]

PR #723, opened 2025-06-10
fix deploying Cloudflare:sockets and cloudflare email workers

franz101:
Based on https://github.com/opennextjs/opennextjs-cloudflare/issues/661

### DIFF
--- a/packages/cloudflare/src/cli/build/bundle-server.ts
+++ b/packages/cloudflare/src/cli/build/bundle-server.ts
@@ -104,7 +104,7 @@ export async function bundleServer(buildOpts: BuildOptions): Promise<void> {
       // Apply updater updates, must be the last plugin
       updater.plugin,
     ] as Plugin[],
-    external: ["./middleware/handler.mjs"],
+    external: ["./middleware/handler.mjs", "cloudflare:sockets", "cloudflare:email"],
     alias: {
       // Note: it looks like node-fetch is actually not necessary for us, so we could replace it with an empty shim
       //       but just to be safe we replace it with a module that re-exports the native fetch


packages/cloudflare/src/cli/build/bundle-server.ts
```diff
@@ -104,7 +104,7 @@ export async function bundleServer(buildOpts: BuildOptions): Promise<void> {
       // Apply updater updates, must be the last plugin
       updater.plugin,
     ] as Plugin[],
-    external: ["./middleware/handler.mjs"],
+    external: ["./middleware/handler.mjs", "cloudflare:sockets", "cloudflare:email"],
```
vicb: Thanks!

Would `cloudflare:*` work here ?

vicb:
Thanks for the change, could you please add a patch changeset (using `pnpm changeset`)

packages/cloudflare/src/cli/build/bundle-server.ts
SamyPesse: Yes, it should, also should it have `node:*`?

vicb: > Yes, it should, also should it have `node:*`?

That is taken care of by [`platform: "node",`](https://esbuild.github.io/api/#platform:~:text=All%20built%2Din%20node%20modules)

But we should have tests for `cloudflare:*` :)

franz101:
Currently running issues, when running

import 'cloudflare:email' in an API file


Maybe I forgot server-only or could be related to the next config:
https://github.com/vercel/next.js/discussions/50177#discussioncomment-6006702

franz101:
@vicb @vicb 

need some support. it looks like opennext just bundles it into the following:
`{var e5=Error("Cannot find module 'cloudflare:email'");throw e5.code="MODULE_NOT_FOUND",e5})` even with external libraries set


https://github.com/vercel/next.js/discussions/50177#discussioncomment-9409065

sommeeeer:
What if you add it to [`serverExternalPackages`](https://nextjs.org/docs/app/api-reference/config/next-config-js/serverExternalPackages) in your next config?

packages/cloudflare/src/cli/build/bundle-server.ts
franz101: how would you do this test?
open-next build worker.ts?

```
import "cloudflare:email";

export async function handler({
    return {}
}

```

franz101:
> What if you add it to [`serverExternalPackages`](https://nextjs.org/docs/app/api-reference/config/next-config-js/serverExternalPackages) in your next config?

Same error:
https://github.com/franz101/opennext-cf-email

franz101:
ok I was able to run it locally. based on this:
https://github.com/cloudflare/next-on-pages/blob/3d84ab0096258febb88193c4b9d269ecbaf1b19b/packages/next-on-pages/src/buildApplication/buildWorkerFile.ts#L88

franz101:
Requires this plugin:
```

export const builtInModulesPlugin = {
	name: 'built-in:modules',
	setup(build) {
		build.onResolve(
			{ filter: /^(node:|cloudflare:|async_hooks)/ },
			({ kind, path }) => {
				/**
				 * This plugin converts `require("<PREFIX>:*")` calls, those are the only ones that need
				 * updating (esm imports to "<PREFIX>:*" are totally valid), so here we tag with the
				 * built-in-modules namespace only imports that are require calls.
				 */
				return kind === 'require-call'
					? { path, namespace: 'built-in-modules' }
					: undefined;
			},
		);

		/**
		 * We convert the imports we tagged with the built-in-modules namespace so that instead of
		 * `require("<PREFIX>:*")` they import from `export * from "<PREFIX>:*";`
		 */
		build.onLoad(
			{ filter: /.*/, namespace: 'built-in-modules' },
			({ path }) => {
				return {
					contents: `export * from '${path}'`,
					loader: 'js',
				};
			},
		);
	},
};
```

```
  webpack: (config) => {
      config.externals.push('cloudflare:email');
      config.externals.push('cloudflare:socket');
      config.externalsType = 'commonjs';
    return config;
  }
};```

franz101:
how would you design the test? it needs to be run with wrangler dev afaik

vicb:
See https://github.com/opennextjs/opennextjs-cloudflare/blob/main/examples/playground15/app/api/cloudflare/route.ts for an example of using `cloudflare:*` packages

PR #725, opened 2025-06-10
test: use diff rather than raw output to test patches

vicb:


vicb:
Thanks for the review James!

PR #730, opened 2025-06-16
test: add tests for remotePatterns

vicb:
- use the compiled patterns from the image manifest instead of re-computing them
- add unit and integration tests for remote patterns
- extract the code out of init to images
- 404 -> 400 to conform with Next.js behavior

/cc @penalosa

### DIFF
--- a/packages/cloudflare/src/cli/templates/images.ts
+++ b/packages/cloudflare/src/cli/templates/images.ts
@@ -16,7 +16,7 @@ export type RemotePattern = {
 export function fetchImage(fetcher: Fetcher | undefined, imageUrl: string) {
   // https://github.com/vercel/next.js/blob/d76f0b1/packages/next/src/server/image-optimizer.ts#L208
   if (!imageUrl || imageUrl.length > 3072 || imageUrl.startsWith("//")) {
-    return new Response("Unsupported image config", { status: 400 });
+    return getUrlErrorResponse();
   }
 
   // Local
@@ -26,10 +26,10 @@ export function fetchImage(fetcher: Fetcher | undefined, imageUrl: string) {
       const url = new URL(imageUrl, "http://n");
       pathname = decodeURIComponent(url.pathname);
     } catch {
-      return new Response("Unsupported image config", { status: 400 });
+      return getUrlErrorResponse();
     }
     if (/\/_next\/image($|\/)/.test(pathname)) {
-      return new Response("Unsupported image config", { status: 400 });
+      return getUrlErrorResponse();
     }
 
     return fetcher?.fetch(`http://assets.local${imageUrl}`);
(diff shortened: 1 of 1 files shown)


packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -0,0 +1,90 @@
+export type RemotePattern = {
+  protocol?: "http" | "https";
+  hostname: string;
+  port?: string;
+  pathname: string;
```
james-elicx, line 6: I believe pathname can be optional here now that we're no longer setting it to a fallback, or does this get defaulted in the manifest?

vicb: Good question!
It's actually an always set regexp in the manifest - I'll add a comment

vicb, line 6:
```suggestion
  // pathname is always set in the manifest (to `makeRe(pathname ?? '**', { dot: true }).source`)
  pathname: string;
```

vicb:
Thanks for the review James!

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -0,0 +1,91 @@
...
+ * Local images (starting with a '/' as fetched using the passed fetcher).
+ * Remote images should match the configured remote patterns or a 404 response is returned.
+ */
+export function fetchImage(fetcher: Fetcher | undefined, imageUrl: string) {
+  // https://github.com/vercel/next.js/blob/d76f0b1/packages/next/src/server/image-optimizer.ts#L208
+  if (!imageUrl || imageUrl.length > 3072 || imageUrl.startsWith("//")) {
+    return new Response("Not Found", { status: 404 });
+  }
+
+  // Local
```
conico974, line 22: We should at least add a comment here to say that we don't do anything with the local pattern for now

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -0,0 +1,91 @@
...
+  if (imageUrl.startsWith("/")) {
+    let pathname: string;
+    try {
+      const url = new URL(imageUrl, "http://n");
+      pathname = decodeURIComponent(url.pathname);
+    } catch {
+      return new Response("Not Found", { status: 404 });
```
conico974: Should it be a 500 ? Or another code than 404 at least

conico974 (approved):

packages/cloudflare/src/cli/templates/images.ts
vicb: Note: local patterns are only supported on Next canary and we could fetch them directly from assets (without using `/_next/image`)

packages/cloudflare/src/cli/templates/images.ts
vicb: Good point.
Next returns 400, I'll update to do the same here

packages/cloudflare/src/cli/templates/images.ts
vicb: https://github.com/opennextjs/opennextjs-cloudflare/issues/731

PR #740, opened 2025-06-19
feat(images): Implement local patterns for images

### LINKED ISSUE #731: implement local patterns for images
Similar to https://github.com/opennextjs/opennextjs-cloudflare/pull/727 but for local pattern.

Not high priority as assets are accessible directly (without usgin `/_next/image?url=/...`) and it is a canary only feature

https://jira.cfdata.org/browse/DEVX-2015

sommeeeer:
fixes #731

### DIFF
--- a/packages/cloudflare/src/cli/templates/images.spec.ts
+++ b/packages/cloudflare/src/cli/templates/images.spec.ts
@@ -1,13 +1,12 @@
-/**
- * See https://github.com/vercel/next.js/blob/64702a9/test/unit/image-optimizer/match-remote-pattern.test.ts
- */
-
 import pm from "picomatch";
 import { describe, expect, it } from "vitest";
 
 import type { LocalPattern } from "./images.js";
 import { matchLocalPattern, matchRemotePattern as mRP } from "./images.js";
 
+/**
+ * See https://github.com/vercel/next.js/blob/64702a9/test/unit/image-optimizer/match-remote-pattern.test.ts
+ */
 describe("matchRemotePattern", () => {
   it("should match literal hostname", () => {
     const p = { hostname: pm.makeRe("example.com") } as const;
@@ -363,6 +362,9 @@ describe("matchRemotePattern", () => {
   });
 });
 
+/**
+ * See https://github.com/vercel/next.js/blob/64702a9/test/unit/image-optimizer/match-local-pattern.test.ts
+ */
 describe("matchLocalPattern", () => {
   const mLP = (p: LocalPattern, urlPathAndQuery: string) =>
     matchLocalPattern(p, new URL(urlPathAndQuery, "http://n"));
--- a/packages/cloudflare/src/cli/templates/images.ts
+++ b/packages/cloudflare/src/cli/templates/images.ts
@@ -1,4 +1,4 @@
-type RemotePattern = {
+export type RemotePattern = {
   protocol?: "http" | "https";
   hostname: string;
   port?: string;
@@ -42,11 +42,10 @@ export function fetchImage(fetcher: Fetcher | undefined, imageUrl: string) {
     }
 
     // If localPatterns are not defined all local images are allowed.
-    if (__IMAGES_LOCAL_PATTERNS__.length === 0) {
-      return fetcher?.fetch(`http://assets.local${imageUrl}`);
-    }
-
-    if (!__IMAGES_LOCAL_PATTERNS__.some((p: LocalPattern) => matchLocalPattern(p, url))) {
+    if (
+      __IMAGES_LOCAL_PATTERNS__.length > 0 &&
+      !__IMAGES_LOCAL_PATTERNS__.some((p: LocalPattern) => matchLocalPattern(p, url))
+    ) {
       return getUrlErrorResponse();
     }
 


examples/playground14/e2e/cloudflare.spec.ts
```diff
@@ -19,14 +19,38 @@ test.describe("playground/cloudflare", () => {
     await expect(res.json()).resolves.toEqual(expect.objectContaining({ PROCESS_ENV_VAR: "process.env" }));
   });
 
-  test("fetch an image allowed by remotePatterns", async ({ page }) => {
-    const res = await page.request.get("/_next/image?url=https://avatars.githubusercontent.com/u/248818");
-    expect(res.status()).toBe(200);
-    expect(res.headers()).toMatchObject({ "content-type": "image/jpeg" });
+  test.describe("remotePatterns", () => {
```
vicb, line 22: nit: we usually use `describe` and import it

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -83,6 +96,21 @@ export function matchRemotePattern(pattern: RemotePattern, url: URL): boolean {
   return true;
 }
 
+export function matchLocalPattern(pattern: LocalPattern, url: URL): boolean {
+  // https://github.com/vercel/next.js/blob/d76f0b1/packages/next/src/shared/lib/match-local-pattern.ts
+  if (pattern.search !== undefined) {
+    if (pattern.search !== url.search) {
+      return false;
+    }
+  }
```
vicb: nit: I removed the unneeded nested if in remote match too
```suggestion
  if (pattern.search !== undefined && pattern.search !== url.search) {
    return false;
  }
```

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -83,6 +96,21 @@ export function matchRemotePattern(pattern: RemotePattern, url: URL): boolean {
...
+
+  if (!new RegExp(pattern.pathname).test(url.pathname)) {
+    return false;
+  }
+
+  return true;
```
vicb:
```suggestion
  return new RegExp(pattern.pathname).test(url.pathname))
```

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -22,16 +28,23 @@ export function fetchImage(fetcher: Fetcher | undefined, imageUrl: string) {
...
       pathname = decodeURIComponent(url.pathname);
     } catch {
       return getUrlErrorResponse();
     }
+
     if (/\/_next\/image($|\/)/.test(pathname)) {
       return getUrlErrorResponse();
     }
 
+    if (!__IMAGES_LOCAL_PATTERNS__.some((p: LocalPattern) => matchLocalPattern(p, url))) {
```
vicb: `[].some(v => true) === false`

Wouldn't that be an issue with empty local patterns

edit: as no local patterns mean allow everything (not the same as remote Patterns) - maybe we should add a comment both here and for RP

vicb:
image

not a 🦊 ??

vicb:
Also would be nice to add tests from https://github.com/vercel/next.js/blob/canary/test/unit/image-optimizer/match-local-pattern.test.ts in images.spec.ts as we have for remote patterns.
Thx

examples/playground14/e2e/cloudflare.spec.ts
sommeeeer: From `@playwright/test`? Its not really there tbh.

packages/cloudflare/src/cli/templates/images.ts
sommeeeer: Yup your right. Fix incoming! :)

examples/playground14/e2e/cloudflare.spec.ts
vicb: You're absolutely right, was thinking vitest.
Sorry for the confusion.

sommeeeer:
Alright, so I added a fix commit. Sorry about the confusion!

edit2: I guess we wrap `pathname` and `hostname` with `makeRe()` in our test cause Next uses the same when writing the manifest [here](https://github.com/vercel/next.js/blob/23ce1e9e66a68b7ddef71b50f44b0f01f6083a21/packages/next/src/build/index.ts#L585-L597)? However, why does Next use it again inside `matchRemotePattern()` and `matchLocalPattern()`? They are also using the arrays from the config.

Say Hello to Snipp! She is the cutest most active cat on this planet. Unfortunently she got sold to a happy customer in London! I hope she is doing fine there (my family owns a cat breeding business 🐈).

vicb:
> However, why does Next use it again inside `matchRemotePattern()` and `matchLocalPattern()`?

Those methods are call from the image optimizer that [passes in the next config](https://github.com/vercel/next.js/blob/0ecd6b59a9803067e562617bc60ee1296db1a99e/packages/next/src/server/image-optimizer.ts#L222-L223) and not the manifest

Snipp is so lovely 🥰

packages/cloudflare/src/cli/templates/images.spec.ts
vicb, line 1: Could you please move l1-3 to l12 and add a similar ref for local pattern tests?

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -1,4 +1,4 @@
-export type RemotePattern = {
+type RemotePattern = {
```
vicb: why is export removed here?
(also see l10)

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -22,16 +28,28 @@ export function fetchImage(fetcher: Fetcher | undefined, imageUrl: string) {
...
     }
 
+    // If localPatterns are not defined all local images are allowed.
+    if (__IMAGES_LOCAL_PATTERNS__.length === 0) {
+      return fetcher?.fetch(`http://assets.local${imageUrl}`);
+    }
+
+    if (!__IMAGES_LOCAL_PATTERNS__.some((p: LocalPattern) => matchLocalPattern(p, url))) {
+      return getUrlErrorResponse();
+    }
```
vicb:
```suggestion
    if (__IMAGES_LOCAL_PATTERNS__.length > 0 && !__IMAGES_LOCAL_PATTERNS__.some((p: LocalPattern) => matchLocalPattern(p, url))) {
      return getUrlErrorResponse();
    }
```

to avoid repeating `return fetcher?.fetch(`http://assets.local${imageUrl}`);` - if that piece need to be updated at some point, we will probably miss one of the 2

vicb (commented):
LGTM, thanks!

A couple more nits and should be good to 🚀

packages/cloudflare/src/cli/templates/images.ts
sommeeeer: I removed it, since its not really being consumed outside that module. You want me to add it back?

vicb: I usually export the types used for exported APIs but up to you.
(As long as we are consistent for remote and local)

vicb (approved):
Awesome, thanks!

PR #746, opened 2025-06-20
feat: add support for skew protection

vicb:
There are 2 main files in this PR:

`packages/cloudflare/src/cli/commands/skew-protection.ts` (build time)

This builds a mapping from `deploymentId` to the worker version at build time.
Note that because the worker you are building has no version yet, "current" is used instead.

`packages/cloudflare/src/cli/templates/skew-protection.ts` (runtime)

At runtime, if a particular version is requested and it is present in the mapping,
we'll fetch the result from a preview URL with a hostname of `<version>-<worker_name>.<domain>.workers.dev`

How to use it:

- set `cloudflare.skewProtection.enabled` to `true` in your OpenNext config
- provide values for the new env vars added to `packages/cloudflare/src/api/cloudflare-context.ts`
- update your next config to set the `deployementId` - you can use the `getDeploymentId()` helper
- set [`run_worker_first` to `true`](https://developers.cloudflare.com/workers/static-assets/binding/#run_worker_first)

Docs PR https://github.com/opennextjs/docs/pull/164

TODO:
- [x] add tests
- [x] handle assets (will need to run the worker before the asset worker)
- [x] follow-up: move the skew protection logic in the aws/routing layer with pluggable overrides -> see #768


packages/cloudflare/src/cli/templates/skew-protection.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    if (requestDeploymentId in mapping) {
+      const version = mapping[requestDeploymentId];
+
+      if (!version || version === CURRENT_VERSION_ID) {
+        return undefined;
+      }
+
+      const versionDomain = version.split("-")[0];
+      const hostname = `${versionDomain}-${process.env.CF_WORKER_NAME}.${process.env.CF_PREVIEW_DOMAIN}.workers.dev`;
```
conico974: I don't think this one will work if people uses durable object, for these cases you could use `Cloudflare-Workers-Version-Overrides` headers with the worker binding.

vicb: `Cloudflare-Workers-Version-Overrides` can only access one of 2 versions in a gradual deployment :(

packages/cloudflare/src/cli/commands/skew-protection.ts
```diff
@@ -0,0 +1,258 @@
...
+): Promise<Record<string, string> | undefined> {
+  if (config.cloudflare?.skewProtectionEnabled !== true) {
+    return undefined;
+  }
+
+  const nextConfig = loadConfig(path.join(options.appBuildOutputPath, ".next"));
+  const deploymentId = nextConfig.deploymentId;
+
+  if (!deploymentId) {
+    logger.error("Deployment ID should be set int the Next config when skew protection is enabled");
```
james-elicx:
```suggestion
    logger.error("Deployment ID should be set in the Next config when skew protection is enabled");
```

james-elicx (commented):
I think this is looking pretty good on my initial read though. Going to have another look later on as well.

james-elicx (approved):
I think this looks good and the logic makes sense to me.

It might be nice for us to look at having an e2e that runs against a deployment at some point in the future, for testing features like this.

packages/cloudflare/src/cli/templates/skew-protection.ts
vicb: We will recommend to deploy DO to another worker when skew protection is enabled.

packages/cloudflare/src/cli/build/open-next/compile-skew-protection.ts
```diff
@@ -17,17 +17,17 @@ export async function compileSkewProtection(options: BuildOptions, config: OpenN
...
+				nodir: true,
+				absolute: false,
+				cwd: assetPath,
+			})
+		: [];
 	// All files located inside `_next/static` are static assets, no need to list them
-	const assetTree = filesToTree(
-		skewProtectionEnabled ? files.filter((path) => !path.startsWith("_next/static/")) : []
-	);
+	const assetTree = filesToTree(files.filter((path) => !path.startsWith("_next/static/")));
```
sommeeeer: Was it not possible to use `ignore` in `glob()` for this?

packages/cloudflare/src/cli/build/open-next/compile-skew-protection.ts
```diff
@@ -0,0 +1,77 @@
...
+		platform: "node",
+		define: {
+			__SKEW_PROTECTION_ENABLED__: JSON.stringify(skewProtectionEnabled),
+			__CF_ASSETS_TREE__: JSON.stringify(assetTree),
+		},
+	});
+}
+
+/**
+ * Converts a list a file to tree of `FolderNode`
```
sommeeeer: Converts a list of file paths into a tree of `FolderNode`

packages/cloudflare/src/cli/commands/skew-protection.ts
```diff
@@ -0,0 +1,258 @@
...
+					break;
+				}
+				versions.push({ id, createdOnMs });
+				if (versions.length >= maxNumberOfVersions) {
+					break;
+				}
+			}
+		}
+	} catch (e) {
+		if (e instanceof NotFoundError && e.status === 404) {
```
sommeeeer, line 255: (e as NotFoundError).status

packages/cloudflare/src/cli/templates/skew-protection.ts
```diff
@@ -0,0 +1,137 @@
...
+	if (path.startsWith("/_next/static/") || isFileInTree(path, __CF_ASSETS_TREE__)) {
+		return assets.fetch(request);
+	}
+}
+
+/**
+ * A node represents a folder in the file tree
+ */
+export type FolderNode = {
+	// List of file file in this folder
```
sommeeeer: List of files in this folder

packages/cloudflare/src/cli/build/open-next/compile-skew-protection.ts
vicb: Great tip, thanks!

packages/cloudflare/src/cli/commands/skew-protection.ts
vicb: Exceptions beside `NotFoundError` might be actual errors we want to surface (i.e. network error, ...)

vicb:
Thanks for the review @sommeeeer !

packages/cloudflare/src/cli/build/open-next/compile-skew-protection.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+		const paths = ["a/b/file1.txt", "c/b/file2.txt"];
+		const tree = filesToTree(paths);
+		expect(tree).toEqual({
+			f: [],
+			d: {
+				a: {
+					f: [],
+					d: {
+						b: {
+							f: ["file1.txt"],
```
conico974: This is horrible to look at with tab :smile:

packages/cloudflare/src/cli/build/open-next/compile-skew-protection.ts
```diff
@@ -0,0 +1,81 @@
...
+	});
+}
+
+/**
+ * Converts a list of file paths into a tree of `FolderNode`
+ *
+ * @param paths The list of path
+ * @returns The root node of the tree
+ */
+export function filesToTree(paths: string[]): FolderNode {
```
conico974: I think in the future here, we should create glob pattern whenever we can to avoid having a too big `FolderNode` . This could grow quite big if people have a lot of assets

packages/cloudflare/src/cli/build/open-next/compile-skew-protection.ts
```diff
@@ -0,0 +1,81 @@
...
+				posix: true,
+				nodir: true,
+				absolute: false,
+				cwd: assetPath,
+				// All files located inside `_next/static` are static assets, no need to list them
+				ignore: ["_next/static/**"],
+			})
+		: [];
+
+	console.log(assets);
```
conico974:
```suggestion
```

packages/cloudflare/src/cli/commands/helpers.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export type WorkerEnvVar = Record<keyof CloudflareEnv, string | undefined>;
+
+/**
+ * Return the string env vars from the worker.
+ *
+ * @param options Options to pass to `getPlatformProxy`, i.e. to set the environment
+ * @returns the env vars
+ */
+export async function getEnvFromPlatformProxy(options: GetPlatformProxyOptions) {
```
conico974, line 11: I think we should have only one instance of this, that we spawn at the beginning and that we dispose at the end. It's a bit wasteful to launch multiple workerd process.
And once we have support for remote bindings here, we could reuse it to populate the cache directly (which should be way faster)

packages/cloudflare/src/cli/commands/skew-protection.ts
```diff
@@ -0,0 +1,258 @@
...
+		afterTimeMs?: number;
+		maxNumberOfVersions?: number;
+	}
+): Promise<{ id: string; createdOnMs: number }[]> {
+	const versions = [];
+	const {
+		client,
+		accountId,
+		afterTimeMs = new Date().getTime() - MAX_VERSION_AGE_DAYS * 24 * 3600 * 1000,
+		maxNumberOfVersions = MAX_NUMBER_OF_VERSIONS,
```
conico974, line 233: I think this one should be overridable

packages/cloudflare/src/cli/templates/skew-protection.ts
```diff
@@ -0,0 +1,137 @@
...
+		return undefined;
+	}
+
+	let path = new URL(request.url).pathname;
+	const basePath = globalThis.__NEXT_BASE_PATH__;
+	if (basePath && path.startsWith(basePath)) {
+		path = path.slice(basePath.length);
+	}
+	if (path.startsWith("/_next/static/") || isFileInTree(path, __CF_ASSETS_TREE__)) {
+		return assets.fetch(request);
```
conico974: Does this one respect the `_headers` file ?

packages/cloudflare/src/cli/build/open-next/compile-skew-protection.ts
vicb: I'll create a follow up issue for that.

packages/cloudflare/src/cli/templates/skew-protection.ts
vicb: It should but I'll test that

vicb: Actually it doesn't, see https://developers.cloudflare.com/workers/static-assets/headers/#custom-headers

vicb: Follow up:
- Check what Next does add by default and apply that in the asset resolver
- Document how to override (using middleware or Next config)

vicb:
I have rebase the PR on top of #768 as the asset resolver will be used to resolve the assets from previous versions (when `run_worker_first=false` the worker does not get a chance to override the assets as they are served even before the worker is executed).

Docs plan (today/tomorrow)
- write some new doc for the asset resolver
- update the existing doc for the skew protection to sync with this PR

@dario-piotrowicz @james-elicx @conico974 please take a look at the PR when you get a chance.

Thanks!

examples/playground15/app/api/env/route.ts
```diff
@@ -4,5 +4,7 @@
 export const dynamic = "force-dynamic";
 
 export async function GET() {
-	return new Response(JSON.stringify(process.env));
+	return new Response(JSON.stringify(process.env, null, 2), {
```
conico974, line 7: Why this change ?

packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -65,6 +65,16 @@ declare global {
 		CACHE_PURGE_ZONE_ID?: string;
 		// The API token to use for the cache purge. It should have the `Cache Purge` permission
 		CACHE_PURGE_API_TOKEN?: string;
+
+		// The following variables must be provided when skew protection is enabled
+		// The name of the worker (as defined in the wrangler configuration)
+		CF_WORKER_NAME?: string;
```
conico974, line 74: It's the name with the environment right ?

packages/cloudflare/src/cli/commands/skew-protection.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+						versions: {
+							list: () => [],
+						},
+					},
+				},
+			};
+
+			const now = Date.now();
+
+			client.workers.scripts.versions.list = vi.fn().mockReturnValue([
```
conico974, line 22: Can the client return this in the wrong order ? This is not tested here

packages/cloudflare/src/cli/templates/skew-protection.ts
```diff
@@ -0,0 +1,73 @@
...
+ * @param request
+ * @returns
+ */
+export function maybeGetSkewProtectionResponse(request: Request): Promise<Response> | Response | undefined {
+	// no early return as esbuild would not treeshake the code.
+	if (__SKEW_PROTECTION_ENABLED__) {
+		const url = new URL(request.url);
+
+		// Skew protection is only active for the latest version of the app served on a custom domain.
+		if (url.hostname === "localhost" || url.pathname.endsWith(".workers.dev")) {
```
conico974:
```suggestion
		if (url.hostname === "localhost" || url.hostname.endsWith(".workers.dev")) {
```

packages/cloudflare/src/cli/templates/skew-protection.ts
```diff
@@ -0,0 +1,73 @@
...
+
+			const versionDomain = version.split("-")[0];
+			const hostname = `${versionDomain}-${process.env.CF_WORKER_NAME}.${process.env.CF_PREVIEW_DOMAIN}.workers.dev`;
+			url.hostname = hostname;
+			const requestToOlderDeployment = new Request(url!, request);
+
+			return fetch(requestToOlderDeployment);
+		}
+
+		return new Response("Bad Request", { status: 400 });
```
conico974: This means that older deploymentId (older than the one in mapping) won't work at all.
Maybe it should be an option to serve the current version

conico974 (approved):
Just a few nits/question, but other than that LGTM.
BTW do we want to allow users to override the assets resolver ?

examples/playground15/app/api/env/route.ts
vicb: I used this API for debugging (the env contains the build id, deployment id, and deployment mapping) and the response was easier to read when formatted this way.

packages/cloudflare/src/api/cloudflare-context.ts
vicb: That's right, the wrangler environment is appended to the name.
I'll add some examples.

packages/cloudflare/src/cli/commands/skew-protection.spec.ts
vicb: In my experience, the client always returns an ordered lists.

BUT I agree we should not rely on the result of a few experiments.

That's why I pay special attention to not order the input (~2 is before ~1) and you can see that the output of `listWorkerVersions` is ordered (~2 after ~1)

packages/cloudflare/src/cli/templates/skew-protection.ts
vicb: Th reasoning here is that versions that are too old might not be compatible with the latest version any more (i.e. because of a DB schema change) qnd we should not try to handle those requests.

But I'm open to suggestion - also we can refine after the PR lands.

james-elicx: I believe on Vercel an invalid deployment ID would just serve you the content from the current version of the app, so it would probably make sense to support that as an option. It could be configurable; fallback to current, or 400.

vicb: Update to serve the latest version on unknown deployment to match Vercel (and avoid an extra config)

vicb:
> BTW do we want to allow users to override the assets resolver ?

Yes we should but I agree that our current `defineCloudflareConfig` is not flexible enough:

```ts
export default {
	...defineCloudflareConfig({
		incrementalCache: r2IncrementalCache,
	}),
	cloudflare: {
		skewProtection: {
			enabled: false,
		},
	},
} satisfies OpenNextConfig;
```

As of today users can override the assets resolver but the syntax is not great.

I'll try to work on a better way to do that.

PR #748, opened 2025-06-22
chore: switch to tabs for indentation

### LINKED ISSUE #39: Switch to hard tabs for indentation
Everyone has their own opinions about how much visual indentation they want in their code. Some prefer 2 spaces, some prefer 4, and then some users may be visually impaired and need higher font sizes with smaller tab widths, etc.

There's been a tonne of debate over the years about which is better, and it's always come to down to arguments that are opinion based, but the one major and consequential difference between the two is this:

**Tabs allow individuals in the same codebase to select their indentation widths across every tool.**

To affirm the accessibility issue, I'd recommend reading through https://www.reddit.com/r/javascript/comments/c8drjo/nobody_talks_about_the_real_reason_to_use_tabs/:

> coworkers who unfortunately are highly visually impaired, and each has a different visual impairment:
> 
> * one of them uses tab-width 1 because he uses such a gigantic font-size
> * the other uses tab-width 8 and a really wide monitor
> * these guys have serious problems using codebases with spaces, they have to convert, do their work, and then unconvert before committing

Other references:

- https://www.reddit.com/r/javascript/comments/c8drjo/nobody_talks_about_the_real_reason_to_use_tabs/
- https://alexandersandberg.com/tabs-for-accessibility/
- https://adamtuttle.codes/blog/2021/tabs-vs-spaces-its-an-accessibility-issue/

Prettier is likely to switch to `useTabs` by default with the next major version too, for this exact reason: https://github.com/prettier/prettier/issues/7475

And other linters/formatters like Biome have already recognised this and default to hard tabs.

Other than the obvious code change differences, shipping an [`.editorconfig`](https://editorconfig.org/) in this repo for some sensible defaults would probably be a good idea too, to combat one of the most common counters to this, which is how some tools render tabs by default. This way, tools like GitHub can render tabs with a 4 character width by default instead of the usual 8, but folks can still customise this themselves if they so choose. Something like this:
```
# http://editorconfig.org
root = true

[*]
indent_style = tab
tab_width = 4
```

I'd like to avoid any debate around personal preference here and focus on the accessibility issue of continuing to indent code with spaces. Prior art:

- https://github.com/cloudflare/workers-sdk/issues/1298

james-elicx:
fixes #39


vicb (approved):

vicb (approved):
Thanks!

Issue #310, opened 2025-01-31
Test request: Add Playwright test for PPR


vicb:
The PPR samples was imported in #308 

We need to add a few Playwright tests.

PR #755, opened 2025-06-30
remove dependency on vercel api in tests

### LINKED ISSUE #368: Switch the `playground/isr` route to use an internal route
Switch from `https://api.vercel.app` to an internal route, see https://github.com/opennextjs/opennextjs-cloudflare/pull/367#discussion_r1950651047

james-elicx:
fixes #368 

Playground tested via e2es.
PPR tested locally after adding usage of Next's connection API.

### DIFF
--- a/examples/next-partial-prerendering/components/pricing.tsx
+++ b/examples/next-partial-prerendering/components/pricing.tsx
@@ -1,3 +1,4 @@
+import { connection } from 'next/server';
 import type { Product } from '#/types/product';
 import { Ping } from '#/components/ping';
 import { ProductEstimatedArrival } from '#/components/product-estimated-arrival';
@@ -13,6 +14,8 @@ import { cookies } from 'next/headers';
 import { getProduct } from '#/lib/products';
 
 async function AddToCartFromCookies() {
+	await connection();
+
 	// Get the cart count from the users cookies and pass it to the client
 	// AddToCart component
 	const cartCount = Number(cookies().get('_cart_count')?.value || '0');
--- a/examples/next-partial-prerendering/components/pricing.tsx
+++ b/examples/next-partial-prerendering/components/pricing.tsx
@@ -10,6 +10,7 @@ import { Suspense } from 'react';
 import { AddToCart } from '#/components/add-to-cart';
 import { delayShippingEstimate, withDelay } from '#/lib/delay';
 import { cookies } from 'next/headers';
+import { getProduct } from '#/lib/products';
 
 async function AddToCartFromCookies() {
 	// Get the cart count from the users cookies and pass it to the client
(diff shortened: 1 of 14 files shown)


james-elicx:
Sorry about that, I didn't realise we didn't have e2e tests to validate that PPR works.

I've added calls with the [`connection()` API](https://nextjs.org/docs/app/api-reference/functions/connection) in each of the 4 components so that it tells Next.js to only let them render at runtime.

Testing locally, this looks like it restores the correct behaviour - static shell, runtime'd components.

vicb:
> Sorry about that, I didn't realise we didn't have e2e tests to validate that PPR works.
> 
> I've added calls with the [`connection()` API](https://nextjs.org/docs/app/api-reference/functions/connection) in each of the 4 components so that it tells Next.js to only let them render at runtime.
> 
> Testing locally, this looks like it restores the correct behaviour - static shell, runtime'd components.

No problem, that's why we review PRs :)

Could you add comments explaining why you call `connection()` in the code. It could be helpful as it differs from the original example - nice fix btw.

_We have https://github.com/opennextjs/opennextjs-cloudflare/issues/310 to add test for PPR_

examples/next-partial-prerendering/components/pricing.tsx
```diff
@@ -10,8 +11,11 @@ import { Suspense } from 'react';
 import { AddToCart } from '#/components/add-to-cart';
 import { delayShippingEstimate, withDelay } from '#/lib/delay';
 import { cookies } from 'next/headers';
+import { getProduct } from '#/lib/products';
 
 async function AddToCartFromCookies() {
+	await connection();
```
james-elicx, line 18:
```suggestion
	// Tell Next.js to render dynamically at runtime instead of build-time
	await connection();
```

examples/next-partial-prerendering/components/recommended-products.tsx
```diff
@@ -1,19 +1,14 @@
+import { connection } from 'next/server';
 import { Product } from '#/types/product';
 import { ProductCard } from '#/components/product-card';
 import { delayRecommendedProducts, withDelay } from '#/lib/delay';
+import { getProducts } from '#/lib/products';
 
 export async function RecommendedProducts() {
+	await connection();
```
james-elicx, line 9:
```suggestion
	// Tell Next.js to render dynamically at runtime instead of build-time
	await connection();
```

examples/next-partial-prerendering/components/reviews.tsx
```diff
@@ -1,19 +1,14 @@
+import { connection } from 'next/server';
 import type { Review } from '#/types/review';
 import { ProductReviewCard } from '#/components/product-review-card';
 import { delayReviews, withDelay } from '#/lib/delay';
+import { getReviews } from '#/lib/reviews';
 
 export async function Reviews() {
+	await connection();
```
james-elicx, line 9:
```suggestion
	// Tell Next.js to render dynamically at runtime instead of build-time
	await connection();
```

examples/next-partial-prerendering/components/single-product.tsx
```diff
@@ -1,12 +1,16 @@
...
 import type { Product } from '#/types/product';
 import { ProductRating } from '#/components/product-rating';
 import Image from 'next/image';
+import { getProduct } from '#/lib/products';
 
 export async function SingleProduct() {
-	const product: Product = await fetch(
-		`https://app-router-api.vercel.app/api/products?id=1`,
-	).then((res) => res.json());
+	await connection();
```
james-elicx, line 10:
```suggestion
	// Tell Next.js to render dynamically at runtime instead of build-time
	await connection();
```

PR #768, opened 2025-07-07
feat: add support for serving assets from the routing layer

vicb:
Implement the [recently introduced `AssetResolver`](https://github.com/opennextjs/opennextjs-aws/pull/917)

Before this PR all assets had to be served before the app worker ([`run_worker_first=false`](https://github.com/opennextjs/opennextjs-aws/pull/917) which is the default value)

With this PR, when `run_worker_first` is set to `true` then the assets will be resolved during the routing phase, after the middleware. We will use that for skew protection as well.

(The code also supports when `use_worker_first` is set to an array of patterns)

### DIFF
--- a/examples/playground14/e2e/cloudflare.spec.ts
+++ b/examples/playground14/e2e/cloudflare.spec.ts
@@ -28,7 +28,10 @@ test.describe("playground/cloudflare", () => {
 
 		test("400 when fetching an image disallowed by remotePatterns", async ({ page }) => {
 			const res = await page.request.get("/_next/image?url=https://avatars.githubusercontent.com/u/248817");
-			expect(res.status()).toBe(400);
+			// The request should be blocked by either the remote patterns or the asset worker
+			const isBlockedByRemotePattern = res.status() === 400;
+			const isBlockedByAssetWorker = res.status() === 403;
+			expect(isBlockedByRemotePattern || isBlockedByAssetWorker).toBe(true);
 		});
 	});
 


packages/cloudflare/src/api/overrides/asset-resolver/index.ts
```diff
@@ -0,0 +1,104 @@
+import type { InternalEvent, InternalResult } from "@opennextjs/aws/types/open-next";
+import type { AssetResolver } from "@opennextjs/aws/types/overrides";
+
+import { getCloudflareContext } from "../../cloudflare-context.js";
+
+/**
+ * Serves assets when `run_worker_first` is set to true.
+ *
+ * When `tun_worker_first` is `false`, the assets are served directly bypassing Next routing.
```
james-elicx:
```suggestion
 * When `run_worker_first` is `false`, the assets are served directly bypassing Next routing.
```

packages/cloudflare/src/api/overrides/asset-resolver/index.ts
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * When it is `true`, assets are served from the routing layer. It should be used when assets
+ * should be behind the middleware or when skew protection is enabled.
+ *
+ * See https://developers.cloudflare.com/workers/static-assets/binding/#run_worker_first
+ */
+const resolver: AssetResolver = {
```
james-elicx, line 16: needs a changeset as well :)

vicb:
> Looks like the e2es are failing. Logic looks good though.

The value of `run_worker_first` was only pass to the server bundle and not to the middleware bundle.

I'll fix this by setting the value on `globalThis` in the init.

james-elicx (approved):

vicb:
Thanks for the review James!

PR #776, opened 2025-07-11
add support for images CSP, disposition, and allow SVG

vicb:
Kudos to @sommeeeer for the implementation!


packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -69,7 +69,43 @@ export function fetchImage(fetcher: Fetcher | undefined, imageUrl: string) {
...
+	const buffer = new ArrayBuffer(32);
+
+	try {
+		let contentType: string | undefined;
+		// body1 is eventually used for the response
+		// body2 is used to detect the content type
+		const [body1, body2] = imgResponse.body.tee();
+		const reader = body2.getReader({ mode: "byob" });
+		const { value } = await reader.read(new Uint8Array(buffer));
+		ctx.waitUntil(reader.cancel());
```
sommeeeer, line 87: A comment here to explain why you use `ctx.waitUntil` would be awesome.

PR #777, opened 2025-07-11
fix _next/image test

vicb:
Root cause is fixed at https://github.com/cloudflare/workers-sdk/pull/9935

This PR should not be merged until the fix is released to miniflare


vicb:
Tests look good after updating Miniflare (via wrangler)

vicb:
Thanks for the review

Issue #778, opened 2025-07-11
[BUG] Build with skew protection enabled fails with missing cloudflare dependency


alex-all3dp:
### Describe the bug

I try to follow the config steps for enabling skew protection.

When I run `opennextjs-cloudflare build` it fails with 

```bash
node:internal/modules/package_json_reader:256
  throw new ERR_MODULE_NOT_FOUND(packageName, fileURLToPath(base), null);
        ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'cloudflare' imported from /../node_modules/@opennextjs/cloudflare/dist/cli/commands/skew-protection.js
```

### Steps to reproduce

1. Upgrade to 1.5.0
2. Run `opennextjs-cloudflare build`

### Expected behavior

`opennextjs-cloudflare build` succeeds

### @opennextjs/cloudflare version

1.5.0

### Wrangler version

4.24.3

### next info output

```shell
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 23.6.0: Wed Jul 31 20:53:05 PDT 2024; root:xnu-10063.141.1.700.5~1/RELEASE_ARM64_T8112
  Available memory (MB): 16384
  Available CPU cores: 8
Binaries:
  Node: 22.17.0
  npm: 10.9.2
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  next: 14.2.30 // An outdated version detected (latest is 15.3.5), upgrade is highly recommended!
  eslint-config-next: 15.3.2
  react: 18.2.0
  react-dom: 18.2.0
  typescript: 5.8.2
Next.js Config:
  output: N/A
```

vicb:
That should be fixed with #779
I'm currently testing and will release if it looks ok

alex-all3dp:
I can confirm that it worked now with 1.5.1, thank you for the quick fix! :)

PR #779, opened 2025-07-11
move cloudflare from devDeps to dep since it is required at runtime

petebacondarwin:
fixes https://github.com/opennextjs/opennextjs-cloudflare/issues/778


vicb (approved):

vicb:
Tested successfully!

Thanks for the fix Pete 🎉

PR #781, opened 2025-07-11
fallback to the content-type header

vicb:
Thanks to @sommeeeer again for the suggestion 🙏


vicb:
Thanks for the reviews!

I'll make a release with the other fix

PR #789, opened 2025-07-15
fix: cancel the body stream when not consumed

vicb:
Unconsumed body stream might lead to stalled connections and be dropped with an exception

ref:
- https://github.com/cloudflare/workerd/issues/4471
- https://github.com/cloudflare/cloudflare-docs/pull/23661
- https://developers.cloudflare.com/workers/platform/limits/#simultaneous-open-connections


conico974 (approved):

vicb:
Thanks for the review!

PR #762, opened 2025-07-03
feat: add config flag support for deploy, preview, and upload commands

0ta-mjn:
Based on #761

Issue #761, opened 2025-07-03
[FEATURE] Add `config` options to `populateCache` command


0ta-mjn:
### Is your feature request related to a problem?

When using the `wrangler` CLI, it is common to specify the configuration file and environment. However, the current implementation of the `populateCache` command does not allow passing these options, which can lead to confusion and errors.

### Describe the solution you'd like

I would like to modify the `populateCache` command to accept `config` options, similar to how the `wrangler` CLI handles them. This will ensure consistency and make it easier to use the command in different environments.

### Describe alternatives you've considered

An alternative would be to manually set the environment and configuration file in the code, but this is not ideal as it reduces flexibility and reusability. By allowing these options to be passed, users can easily switch between different configurations and environments without modifying the code.

### @opennextjs/cloudflare version

1.3.1

### Before submitting

- [x] I have checked that there isn't already a similar feature request
- [x] This is a single feature (not multiple features in one request)

vicb:
Implemented in #791

PR #790, opened 2025-07-15
feat: add config flag support for deploy, preview, and upload commands

0ta-mjn:
Based on #761 .
Old: #762

### DIFF
--- a/packages/cloudflare/src/cli/commands/populate-cache.ts
+++ b/packages/cloudflare/src/cli/commands/populate-cache.ts
@@ -96,11 +96,14 @@ export function getCacheAssets(opts: BuildOptions): CacheAsset[] {
 
 async function populateR2IncrementalCache(
 	options: BuildOptions,
-	populateCacheOptions: { target: WranglerTarget; environment?: string }
+	populateCacheOptions: { target: WranglerTarget; environment?: string; config?: string }
 ) {
 	logger.info("\nPopulating R2 incremental cache...");
 
-	const config = unstable_readConfig({ env: populateCacheOptions.environment });
+	const config = unstable_readConfig({
+		env: populateCacheOptions.environment,
+		config: populateCacheOptions.config,
+	});
 
 	const binding = config.r2_buckets.find(({ binding }) => binding === R2_CACHE_BINDING_NAME);
 	if (!binding) {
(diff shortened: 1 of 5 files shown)


packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -96,11 +96,14 @@ export function getCacheAssets(opts: BuildOptions): CacheAsset[] {
 
 async function populateR2IncrementalCache(
 	options: BuildOptions,
-	populateCacheOptions: { target: WranglerTarget; environment?: string }
+	populateCacheOptions: { target: WranglerTarget; environment?: string; config?: string }
```
vicb, line 99: Could you please extract the type to top level?
(Maybe add a comment here that the `cacheChunkSize` is not used)

vicb (approved):
LGTM, thanks a lot for the PR 🙏

@james-elicx Could you please take a look when you have time

vicb:
@0ta-mjn We finally went with https://github.com/opennextjs/opennextjs-cloudflare/pull/791 that James created.

It makes the CLI more robust and more extensible and has a better UX.

We really appreciate the energy you put in this PR, sorry that we end up merging it :(

You [feature request](https://github.com/opennextjs/opennextjs-cloudflare/issues/761) should now be addressed.

alex-all3dp:
@vicb I tested 1.6.0 and tried to run `opennextjs-cloudflare preview --config wrangler.dev.jsonc` but it does not use the given config it appears, from the errors I get. When running `wrangler dev --config wrangler.dev.jsonc` it works as expected. Is `--config` really available for all commands now, as was intended with this PR?

james-elicx:
> @vicb I tested 1.6.0 and tried to run `opennextjs-cloudflare preview --config wrangler.dev.jsonc` but it does not use the given config it appears, from the errors I get. When running `wrangler dev --config wrangler.dev.jsonc` it works as expected. Is `--config` really available for all commands now, as was intended with this PR?

Hi @alex-all3dp, I've just run the following command:
```
pnpm --filter static-assets-incremental-cache exec opennextjs-cloudflare preview --config wrangler.jsonc
```
And the following is the command that we ran in response to that:
```
pnpm exec wrangler dev --config wrangler.jsonc
```

So it looks like the config arg is being passed along. Are you able to expand on what you're seeing?

james-elicx:
Are you referring to cache population instead?

Do you have an example of the errors?

vicb:
Oh I guess the pb might be with `initOpenNextCloudflareForDev();` called from the build.

@alex-all3dp would you mind creating a new issue with logs?

Thanks!

vicb:
^if that's confirmed we could add `NEXT_DEV_WRANGLER_CONFIG` (similar to `NEXT_DEV_WRANGLER_ENV`) and have the CLI set that when the flags are passed.

alex-all3dp:
Removing `initOpenNextCloudflareForDev()` from the next.config didn't change it. I'll look into creating a separate issue with more infos

PR #791, opened 2025-07-16
refactor: use yargs for the cli

james-elicx:
**Changes**
- Replaces our args parsing with yargs (out of the box commands, optionals, help menus, etc.)
- Removes `output` arg as it was never used.
- Moves the _setup_ part of the CLI to a reusable function called in each command, and commands now just accept args as their param.
- Moves wrangler config retrieval to the setup of the CLI.
- Reads the compiled OpenNext config in non-build commands.
- The flags `--env` and `--config` become proper first class citizens.

Replaces https://github.com/opennextjs/opennextjs-cloudflare/pull/790

**Commands Tested**
- [x] `build`
- [x] `preview`
- [x] `deploy`
- [x] `upload`
- [ ] `populateCache`

### DIFF
--- a/packages/cloudflare/src/cli/commands/utils.ts
+++ b/packages/cloudflare/src/cli/commands/utils.ts
@@ -55,7 +55,7 @@ export async function retrieveCompiledConfig() {
 	const configPath = path.join(nextAppDir, ".open-next/.build/open-next.config.edge.mjs");
 
 	if (!existsSync(configPath)) {
-		logger.error("Could not find compiled Open Next config");
+		logger.error("Could not find compiled Open Next config, did you run the build command?");
 		process.exit(1);
 	}
 


packages/cloudflare/src/cli/args.ts
```diff
@@ -1,127 +1,118 @@
...
+				withWranglerOptions(c)
+					.option("skipNextBuild", {
+						type: "boolean",
+						alias: ["skipBuild", "s"],
+						default: ["1", "true", "yes"].includes(String(process.env.SKIP_NEXT_APP_BUILD)),
+						desc: "Skip building the Next.js app",
+					})
+					.option("noMinify", {
+						type: "boolean",
+						alias: "s",
```
vicb: is it expected that all alias are "s"?

packages/cloudflare/src/cli/args.ts
```diff
@@ -1,127 +1,118 @@
...
-			};
-		default:
-			throw new Error(
-				"Error: invalid command, expected 'build' | 'preview' | 'deploy' | 'upload' | 'populateCache'"
-			);
-	}
+export function runCommand() {
+	return yargs(process.argv.slice(2))
+		.scriptName("opennextjs-cloudflare")
+		.parserConfiguration({ "unknown-options-as-args": true })
```
vicb: Would it be possible to do something like:

```ts
const y =  yargs(process.argv.slice(2))
		.scriptName("opennextjs-cloudflare")
		.parserConfiguration({ "unknown-options-as-args": true });

// Defined in command/<command>.ts
addBuildCommand(y);
addPreviewCommand(y);
...
```

packages/cloudflare/src/cli/args.ts
```diff
@@ -1,127 +1,118 @@
...
-	if (!dirStats.isDirectory()) {
-		throw new Error(`Error: the provided${argName ? ` "${argName}"` : ""} input is not a directory`);
-	}
+function withWranglerPassthroughArgs<
+	T extends yargs.ArgumentsCamelCase<{
+		config: string | undefined;
+		env: string | undefined;
+	}>,
+>(args: T) {
+	return { ...args, passthrough: getWranglerArgs(args) };
```
vicb: Would renaming "Passthrough(Arg)" to wranglerArg make things clearer?

packages/cloudflare/src/cli/commands/setup-cli.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export type WithWranglerArgs<T = unknown> = T & {
+	passthrough: string[];
+	config: string | undefined;
+	env: string | undefined;
+};
+
+const nextAppDir = process.cwd();
+
+export async function setupCLI(
```
vicb: Would be nice to add JSDoc (even one liners) the function

packages/cloudflare/src/cli/args.ts
james-elicx: oops!

james-elicx:
> Does it still work as expected when there is a `--`?

Unknown args and args after `--` will be added to the array of args to pass to wrangler.

If `--env` or `--config` are supplied in the `--`, they won't be picked up by OpenNext but will be picked up by the Wrangler CLI. Because I'm changing them to be first class citizens, we're no longer doing the searching through argv for them at any position, so it wouldn't extract them if they're after `--`.

vicb:
> If `--env` or `--config` are supplied in the `--`, they won't be picked up by OpenNext but will be picked up by the Wrangler CLI. Because I'm changing them to be first class citizens, we're no longer doing the searching through argv for them at any position, so it wouldn't extract them if they're after `--`.

Is it what we want?
IIUC `opennextjs-cloudflare deploy -- --env=...` would populate i.e. populate the cache and run `wrangler deploy` for different envs?

packages/cloudflare/src/cli/commands/build.ts
```diff
@@ -0,0 +1,48 @@
+import { compileOpenNextConfig } from "@opennextjs/aws/build/compileConfig.js";
+import type yargs from "yargs";
+
+import { build as buildImpl } from "../build/build.js";
+import { createOpenNextConfigIfNotExistent } from "../build/utils/create-config-files.js";
+import { setupCLI, withWranglerOptions, withWranglerPassthroughArgs } from "./setup-cli.js";
+
+async function buildCommand(args: {
```
vicb: It would be nice to add some JSDoc here.

i.e. does `wranglerArgs` contain config and env ?

Something like:

```ts
/**
 * Implementation of the `opennextjs- clouflare command`
 *
  * @param ...
  * @returns 
*/
```

An explicit return type would be nice too

packages/cloudflare/src/cli/commands/build.ts
```diff
@@ -0,0 +1,48 @@
...
+}) {
+	const { options, config, wranglerConfig, baseDir } = await setupCLI("build", args, async (baseDir) => {
+		await createOpenNextConfigIfNotExistent(baseDir);
+		return compileOpenNextConfig(baseDir, undefined, { compileEdge: true });
+	});
+
+	return buildImpl(options, config, { ...args, minify: !args.noMinify, sourceDir: baseDir }, wranglerConfig);
+}
+
+export function addBuildCommand<T extends yargs.Argv>(y: T) {
```
vicb, line 47:
```suggestion
/**
 * Add the `build` command to yargs configuration
 * 
 * Consumes 1 positional parameter
 */ 
export function addBuildCommand<T extends yargs.Argv>(y: T) {
```

packages/cloudflare/src/cli/commands/build.ts
```diff
@@ -0,0 +1,48 @@
+import { compileOpenNextConfig } from "@opennextjs/aws/build/compileConfig.js";
+import type yargs from "yargs";
+
+import { build as buildImpl } from "../build/build.js";
+import { createOpenNextConfigIfNotExistent } from "../build/utils/create-config-files.js";
+import { setupCLI, withWranglerOptions, withWranglerPassthroughArgs } from "./setup-cli.js";
+
+async function buildCommand(args: {
+	wranglerArgs: string[];
+	config: string | undefined;
```
vicb: nit: rename this to `configPath` (we also use `config` for the values in i.e. `populateR2IncrementalCache`)

packages/cloudflare/src/cli/commands/setup-cli.ts
```diff
@@ -0,0 +1,112 @@
...
+export type WithWranglerArgs<T = unknown> = T & {
+	wranglerArgs: string[];
+	config: string | undefined;
+	env: string | undefined;
+};
+
+const nextAppDir = process.cwd();
+
+/**
+ * Setup the CLI, print necessary messages, and retrieve various options and configs.
```
vicb: `setupCompiledAppCLI` has the same comment.

What's the difference, when to use which ?

vicb: IIUC the callback is to do some work in the middle of this function.

Another way could be do move this to the caller:

```ts
ShowHeader();
CompileConfig(); // for build
LoadConfig(); // for other commands
ExecuteCommand();
```
Not sure if this would make the code easier to follow?

packages/cloudflare/src/cli/commands/setup-cli.ts
```diff
@@ -0,0 +1,112 @@
...
+		}
+
+		const config = await import(configPath).then((mod) => mod.default);
+
+		// Note: buildDir is not used when an app is already compiled.
+		return { config, buildDir: baseDir };
+	});
+}
+
+export function withWranglerOptions<T extends yargs.Argv>(args: T) {
```
vicb: Some JSDoc would help for those 3 functions

packages/cloudflare/src/cli/index.ts
```diff
@@ -1,63 +1,25 @@
...
 
-const nextAppDir = process.cwd();
+export function runCommand() {
+	let y = yargs(process.argv.slice(2))
+		.scriptName("opennextjs-cloudflare")
+		.parserConfiguration({ "unknown-options-as-args": true });
 
-async function runCommand(args: Arguments) {
-	printHeader(`Cloudflare ${args.command}`);
+	y = addBuildCommand(y);
```
vicb: maybe this can return y for chaining?
Up to you

vicb (commented):
Looks good.
I think more comments might help maintaining this code in the future

packages/cloudflare/src/cli/commands/setup-cli.ts
james-elicx: I've changed it to have a single setupCLI function that takes an option for whether the config should be compiled or not instead. That should make it clearer what's going on.

packages/cloudflare/src/cli/commands/build.ts
james-elicx: I've added a bunch more jsdoc to various places and cleaned up the types, hope that helps.

packages/cloudflare/src/cli/index.ts
james-elicx: Not sure I follow - return it where?

james-elicx:
> > If `--env` or `--config` are supplied in the `--`, they won't be picked up by OpenNext but will be picked up by the Wrangler CLI. Because I'm changing them to be first class citizens, we're no longer doing the searching through argv for them at any position, so it wouldn't extract them if they're after `--`.
> 
> Is it what we want? IIUC `opennextjs-cloudflare deploy -- --env=...` would populate i.e. populate the cache and run `wrangler deploy` for different envs?

Yep, that's true, although people were complaining when we wanted people to use `--` instead of args being first-class citizens...

What would you think of us throwing an error when people pass `--env` or `--config` after the `--`? I really wanted to make the flag a first-class citizen in the CLI because we use it like one for things like cache population. It's not normal for CLIs to parse flags after the `--` and use them internally, which is why I was trying to get away from doing that.

packages/cloudflare/src/cli/index.ts
vicb: Nevermind... 
Do you need the y= or would the function mutate y?

packages/cloudflare/src/cli/index.ts
```diff
@@ -1,63 +1,25 @@
...
-import { getWranglerConfigFlag, getWranglerEnvironmentFlag } from "./utils/run-wrangler.js";
+import { addBuildCommand } from "./commands/build.js";
+import { addDeployCommand } from "./commands/deploy.js";
+import { addPopulateCacheCommand } from "./commands/populate-cache.js";
+import { addPreviewCommand } from "./commands/preview.js";
+import { addUploadCommand } from "./commands/upload.js";
 
-const nextAppDir = process.cwd();
+export function runCommand() {
+	let y = yargs(process.argv.slice(2))
```
vicb: Can we filter out -- here?

packages/cloudflare/src/cli/index.ts
james-elicx: it mutates it

packages/cloudflare/src/cli/index.ts
james-elicx: Anything after `--` are considered non-option arguments and put into `args._` by yargs

vicb:
I took a deeper look and have some comments:

First I really like the changes, I fell like the CLI is much nice now, thanks for working on this.

I do have a few comments:

1) `--`

We use to recommend [`opennextjs-cloudflare deploy -- --env=prod`](https://opennext.js.org/cloudflare/howtos/dev-deploy)

IIUC this command will now (maybe before also?) use a different env for populating the cache than for deploying.

2) yargs

I see

image

image

Would be nice to ba able to solve this

3) Open Next config / `setupCLI`

IIUC `setupCLI` is here to factor some code. It can retrieve or compile the config.

I wouldn't mind more duplication in each of the command to make the code easier to follow/extend

i.e. each command would:

```ts
printHeader();

compileConfig() or retieveCompiledConfig();

readWranglerConfig();
```

Each of this function could be helpers (i.e. `printHeader();` could print the helper and show the windows warning, ...)

Maybe one thing we could do then is to move `compileConfig()` back to `build.ts` because it really belongs there.

Thinking forward if we add new commands, we might have to add options to `setupCLI()` and it might makes code less readable.

james-elicx:
> I took a deeper look and have some comments:
> 
> First I really like the changes, I fell like the CLI is much nice now, thanks for working on this.
> 
> I do have a few comments:
> 
> 1. `--`
> 
> We use to recommend [`opennextjs-cloudflare deploy -- --env=prod`](https://opennext.js.org/cloudflare/howtos/dev-deploy)
> 
> IIUC this command will now (maybe before also?) use a different env for populating the cache than for deploying.

So would you like to preserve the old behaviour, or throw an error if it's used after the `--`?

> 2. yargs
> 
> I see
> 
> image image
> Would be nice to ba able to solve this

I don't get this - that looks like the dependencies aren't installed.

> 3. Open Next config / `setupCLI`
> 
> IIUC `setupCLI` is here to factor some code. It can retrieve or compile the config.
> 
> I wouldn't mind more duplication in each of the command to make the code easier to follow/extend
> 
> i.e. each command would:
> 
> ```ts
> printHeader();
> 
> compileConfig() or retieveCompiledConfig();
> 
> readWranglerConfig();
> ```
> 
> Each of this function could be helpers (i.e. `printHeader();` could print the helper and show the windows warning, ...)
> 
> Maybe one thing we could do then is to move `compileConfig()` back to `build.ts` because it really belongs there.
> 
> Thinking forward if we add new commands, we might have to add options to `setupCLI()` and it might makes code less readable.

Sure, I'll change it.

packages/cloudflare/src/cli/commands/utils.ts
```diff
@@ -0,0 +1,142 @@
...
+/**
+ * Retrieve a compiled OpenNext config, and ensure it is for Cloudflare.
+ *
+ * @returns OpenNext config.
+ */
+export async function retrieveCompiledConfig() {
+	const configPath = path.join(nextAppDir, ".open-next/.build/open-next.config.edge.mjs");
+
+	if (!existsSync(configPath)) {
+		logger.error("Could not find compiled Open Next config");
```
vicb:
```suggestion
		logger.error("Could not find compiled Open Next config, did you run the build command?");
```

vicb (approved):
Thank for this PR!

Great code and nice to have some help on the CLI now!

One nit about an error message when the config is not found.

Issue #794, opened 2025-07-16
[BUG] NextJS instrumentation hook error with NextJS 15.4.1


bapspatil:
### Describe the bug

In my monorepo (NextJS web app + Hono server), after I upgraded from NextJS 15.3.4 to 15.4.1, and deploy my web app with OpenNext Cloudflare, this is the error log I see on my Cloudflare Workers dashboard:
```
NextJS request failed. Error: An error occurred while loading the instrumentation hook
```
There is another error log right below this one, which also says:
```
Failed to prepare server Error: An error occurred while loading the instrumentation hook
```
When I open my NextJS app, I see the same error but in JSON format. If I revert back to NextJS 15.3.4, everything's working just fine again. 

Here is my `package.json` file:
```json
{
  "name": "web",
  "version": "0.1.0",
  "private": true,
  "scripts": {
    "dev": "next dev --turbopack --port=3001",
    "build": "next build",
    "start": "next start",
    "lint": "next lint",
    "preview": "opennextjs-cloudflare build && opennextjs-cloudflare preview",
    "deploy": "opennextjs-cloudflare build && opennextjs-cloudflare deploy",
    "upload": "opennextjs-cloudflare build && opennextjs-cloudflare upload",
    "cf-typegen": "wrangler types --env-interface CloudflareEnv cloudflare-env.d.ts"
  },
  "dependencies": {
    "@gsap/react": "^2.1.2",
    "@opennextjs/cloudflare": "^1.5.1",
    "@orpc/client": "^1.7.2",
    "@orpc/server": "^1.7.2",
    "@orpc/tanstack-query": "^1.7.2",
    "@radix-ui/react-checkbox": "^1.3.2",
    "@radix-ui/react-dropdown-menu": "^2.1.15",
    "@radix-ui/react-label": "^2.1.7",
    "@radix-ui/react-slot": "^1.2.3",
    "@radix-ui/react-tooltip": "^1.2.7",
    "@tanstack/react-form": "^1.14.1",
    "@tanstack/react-query": "^5.83.0",
    "better-auth": "^1.2.12",
    "class-variance-authority": "^0.7.1",
    "clsx": "^2.1.1",
    "gsap": "^3.13.0",
    "lucide-react": "^0.525.0",
    "motion": "^12.23.6",
    "next": "15.3.4",
    "next-themes": "^0.4.6",
    "react": "^19.1.0",
    "react-dom": "^19.1.0",
[...]

Issue #792, opened 2025-07-16
[BUG] Preview fails with Next.js 15.4.1 due to instrumentation hook error (works in 15.3.5)


rakibatomnixima:
### Describe the bug

**This is a blocking issue preventing production deployment.** The error occurs specifically during `opennextjs-cloudflare preview` command, while `next dev` and `next build` work perfectly.

### **Issue Summary**

- **Package**: `@opennextjs/cloudflare` v1.5.1
- **Next.js Version**: 15.4.1
- **Build/Dev**: ✅ Working perfectly
- **Preview/Deploy**: ❌ Failing with instrumentation hook error

### 🐛 **Error Details**

**Preview Response:**

```json
{
    "message": "Server failed to respond.",
    "details": {}
}
```

**Console Error:**

```
✘ [ERROR] Failed to prepare server Error: An error occurred while loading the instrumentation hook
      at Ke.loadInstrumentationModule
(Users/omnixima/Projects/Github/omnipixcraft/apps/web/.wrangler/tmp/dev-UApikN/worker.js:2219:27549)
at async Ke.prepare
(Users/omnixima/Projects/Github/omnipixcraft/apps/web/.wrangler/tmp/dev-UApikN/worker.js:1123:15311)
{
[cause]: Error: Dynamic require of "/.next/server/instrumentation.js" is not supported
at <unknown>
(Users/omnixima/Projects/Github/omnipixcraft/apps/web/.wrangler/tmp/dev-UApikN/worker.js:1:273)
at getInstrumentationModule
(Users/omnixima/Projects/Github/omnipixcraft/apps/web/.wrangler/tmp/dev-UApikN/worker.js:1138:2752052)
at Ke.loadInstrumentationModule
(Users/omnixima/Projects/Github/omnipixcraft/apps/web/.wrangler/tmp/dev-UApikN/worker.js:2219:27444)
at Ke.prepare
(Users/omnixima/Projects/Github/omnipixcraft/apps/web/.wrangler/tmp/dev-UApikN/worker.js:1123:15322)
at Ke.makeRequestHandler
(Users/omnixima/Projects/Github/omnipixcraft/apps/web/.wrangler/tmp/dev-UApikN/worker.js:2219:34383)
at Ke.getRequestHandler
(Users/omnixima/Projects/Github/omnipixcraft/apps/web/.wrangler/tmp/dev-UApikN/worker.js:2219:34231)
at Ke.getRequestHandlerWithMetadata
[...]

Issue #667, opened 2025-05-15
Page Router api route fails on latest canary


conico974:
Adding page router api route to the experimental e2e app make them fail with dynamic require of instrumentation.

If i had to guess it's probably because page router api route are the first one to "migrate" to the adapter API.
It probably means that all patch will have to be rewritten to some extent  (**Needs some investigation**)

neilenns:
I rolled to Canary today to play with `'use cache'` and got the same errors:

```
 Failed to prepare server Error: An error occurred while loading the instrumentation hook
      at NextNodeServer.loadInstrumentationModule
  (workspaces/sea-resource-guides/.wrangler/tmp/dev-iohvRn/worker.js:138572:78)
      at async NextNodeServer.prepare
  (workspaces/sea-resource-guides/.wrangler/tmp/dev-iohvRn/worker.js:70592:86) {
    [cause]: Error: Dynamic require of "/.next/server/instrumentation.js" is not supported
```

(I'm using app router, not page router, but got the same error)

Dramelac:
I just got the same error but with the **latest official stable** release `15.4.1` ! But i'm not sure if this is cache releated because I dont have `use cache`...

In my workers logs in got thoses errors:
- `Failed to prepare server Error: An error occurred while loading the instrumentation hook`
- `NextJS request failed. Error: An error occurred while loading the instrumentation hook`

<img width="2110" height="393" alt="Image" src="https://github.com/user-attachments/assets/4a74c931-db99-4ef0-bd02-7a22871a953d" />

And I receive this response from cloudflare worker (if it can help):
`200 OK`
```
{
  "message": "Server failed to respond.",
  "details": {}
}
```

As a workarround, I rollback next.js to version `15.3.5` and it worked again.

This critical bug is serious now that this affect the latest stable release !

Thanks for your help

sommeeeer:
https://github.com/vercel/next.js/blob/f1b88b4cfa63a6560a04baa209c09065c419969c/packages/next/src/server/lib/router-utils/instrumentation-globals.external.ts#L12-L43 . These lines needs to be patched.

`loadManifest` also needs to be patched. You will get an error on path `required-server-files.json` and `.next/BUILD_ID`. (`✘ [ERROR] ⨯ Error: Unexpected loadManifest(/.next/required-server-files.json) call!`)

[...]

PR #795, opened 2025-07-16
fix(patches): Update patchInstrumentation and loadManifest to work with Next 15.4

sommeeeer:
For #667, #794 and #792.

[This](https://github.com/vercel/next.js/blob/f1b88b4cfa63a6560a04baa209c09065c419969c/packages/next/src/server/lib/router-utils/instrumentation-globals.external.ts#L12-L43) code in Next needs to be patched, and also `loadManifest` also needs to be patched. It will get a path for `required-server-files.json` and `.next/BUILD_ID`.

Tested on `next@15.4.2-canary.1`

Update: This will make instrumenation and loadManifest work on `>=15.4.2`, but there is more patches that need update. Incremental Cache is not getting applied and possibly more issues. Needs more investigation.

### DIFF
--- a/packages/cloudflare/src/cli/build/patches/plugins/instrumentation.spec.ts
+++ b/packages/cloudflare/src/cli/build/patches/plugins/instrumentation.spec.ts
@@ -104,7 +104,7 @@ describe("getInstrumenationModule (Next154)", () => {
         return cachedInstrumentationModule;
       }
       try {
-        cachedInstrumentationModule = (0, _interopdefault.interopDefault)(await require(_nodepath.default.join(projectDir, distDir, "server", \`\${_constants.INSTRUMENTATION_HOOK_FILENAME}.js\`\)));
+        cachedInstrumentationModule = (0, _interopdefault.interopDefault)(await require(_nodepath.default.join(projectDir, distDir, "server", \`\${_constants.INSTRUMENTATION_HOOK_FILENAME}.js\`)));
         return cachedInstrumentationModule;
       } catch (err) {
         if ((0, _iserror.default)(err) && err.code !== "ENOENT" && err.code !== "MODULE_NOT_FOUND" && err.code !== "ERR_MODULE_NOT_FOUND") {
--- a/packages/cloudflare/src/cli/build/patches/plugins/instrumentation.spec.ts
+++ b/packages/cloudflare/src/cli/build/patches/plugins/instrumentation.spec.ts
@@ -1,7 +1,7 @@
 import { patchCode } from "@opennextjs/aws/build/patch/astCodePatcher.js";
 import { describe, expect, test } from "vitest";
 
-import { getNext14Rule, getNext15Rule } from "./instrumentation.js";
+import { getNext14Rule, getNext15Rule, getNext154Rule } from "./instrumentation.js";
 
 describe("LoadInstrumentationModule (Next15)", () => {
 	const code = `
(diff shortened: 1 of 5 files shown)


packages/cloudflare/src/cli/build/patches/plugins/instrumentation.ts
```diff
@@ -10,6 +10,16 @@ import { normalizePath } from "../../utils/normalize-path.js";
 export function patchInstrumentation(updater: ContentUpdater, buildOpts: BuildOptions): Plugin {
 	const builtInstrumentationPath = getBuiltInstrumentationPath(buildOpts);
 
+	updater.updateContent("patch-instrumentation-next15-4", [
+		{
+			field: {
+				filter: /\.(js|mjs|cjs|jsx|ts|tsx)$/,
```
conico974: This isn't right, you're replacing any async function that is called `getInstrumentationModule` in any file in the project (including node_modules)

packages/cloudflare/src/cli/build/patches/plugins/instrumentation.ts
```diff
@@ -36,6 +46,29 @@ export function patchInstrumentation(updater: ContentUpdater, buildOpts: BuildOp
 	};
 }
 
+export function getNext154Rule(builtInstrumentationPath: string | null) {
+	return `
+rule:
+  kind: function_declaration
+  any:
+    - has: {field: name, regex: ^getInstrumentationModule$}
```
conico974: Try to be more specific in your fix, the only part that needs to be changed is this one line, try to only modify this

packages/cloudflare/src/cli/build/patches/plugins/load-manifest.ts
```diff
@@ -61,6 +63,9 @@ function loadManifest($PATH, $$$ARGS) {
 		},
 		fix: `
 function loadManifest($PATH, $$$ARGS) {
+ if ($PATH === "/.next/BUILD_ID") {
```
vicb, line 66:
```suggestion
   $PATH = $PATH.replaceAll(${JSON.stringify(sep)}, ${JSON.stringify(posix.sep)});
  if ($PATH === "/.next/BUILD_ID") {
  return process.env.NEXT_BUILD_ID;
	}
```

packages/cloudflare/src/cli/build/patches/plugins/load-manifest.ts
```diff
@@ -61,6 +63,9 @@ function loadManifest($PATH, $$$ARGS) {
...
+  return process.env.NEXT_BUILD_ID;
+	}
   $PATH = $PATH.replaceAll(${JSON.stringify(sep)}, ${JSON.stringify(posix.sep)});
```
vicb:
```suggestion
  $PATH = $PATH.replaceAll(${JSON.stringify(sep)}, ${JSON.stringify(posix.sep)});
  if ($PATH === "/.next/BUILD_ID") {
  return process.env.NEXT_BUILD_ID;
	}
```

packages/cloudflare/src/cli/build/patches/plugins/instrumentation.spec.ts
```diff
@@ -96,3 +96,59 @@ describe("prepareImpl (Next14)", () => {
     `);
 	});
 });
+
+describe("getInstrumenationModule (Next154)", () => {
```
vicb:
```suggestion
describe("getInstrumenationModule (Next15.4)", () => {
```

packages/cloudflare/src/cli/build/patches/plugins/instrumentation.ts
```diff
@@ -4,12 +4,28 @@ import { join } from "node:path";
...
+	updater.updateContent("patch-instrumentation-next15-4", [
+		{
+			field: {
+				filter: getCrossPlatformPathRegex(
+					String.raw`/server/lib/router-utils/instrumentation-globals.external\.js$`,
+					{
+						escape: false,
+					}
+				),
+				contentFilter: /async function getInstrumentationModule\(/,
```
vicb: Use a quick heuristic unaffected by formatting change
```suggestion
				contentFilter: /getInstrumentationModule\(/,
```

packages/cloudflare/src/cli/build/patches/plugins/instrumentation.ts
```diff
@@ -36,6 +52,35 @@ export function patchInstrumentation(updater: ContentUpdater, buildOpts: BuildOp
...
+    stopBy: end
+    has:
+      all:
+        - has: {kind: return_statement, pattern: return cachedInstrumentationModule}
+    inside:
+      kind: statement_block
+      inside:
+        kind: function_declaration
+        all:
+         - has: {field: name, pattern: getInstrumentationModule}
```
vicb: Maybe the rule could be simplified, i.e.

```ts
rule:
  kind: expression_statement
  has:
    kind: assignment_expression
    all:
      - has: { pattern: "cachedInstrumentationModule" }
      - has: { kind: call_expression, regex: "INSTRUMENTATION_HOOK_FILENAME"}
  inside:
    kind: try_statement
    stopBy: end
    has: { regex: "return cachedInstrumentationModule" }
    inside:
      kind: function_declaration
      stopBy: end
      has: { field: name, pattern: getInstrumentationModule }
```

vicb (commented):
Thanks!

Does that fix at least some of the examples updated to 15.4.2?

It would be nice to add that info to the PR description

packages/cloudflare/src/cli/build/patches/plugins/instrumentation.ts
sommeeeer: Yeah I agree, this is much better! Thanks

sommeeeer:
Thanks for the review Victor!

> Does that fix at least some of the examples updated to 15.4.2?

It does fix these patches for `15.4.2`, but there is more patches that needs update. For the cache handler. You will get an error during runtime when you visit any route:

`✘ [ERROR] Failed to update prerender cache for /index Error: [unenv] fs.mkdir is not implemented yet!`
image

I havent figured out which part yet, but for some reason the incremental cache is not getting applied and I think [this](https://github.com/vercel/next.js/blob/719bc5e71bdcf4881e71ec54e1cdffa19e32f455/packages/next/src/lib/multi-file-writer.ts#L56) code from Next gets run. Needs further investigation.

vicb:
> I havent figured out which part yet, but for some reason the incremental cache is not getting applied and I think [this](https://github.com/vercel/next.js/blob/719bc5e71bdcf4881e71ec54e1cdffa19e32f455/packages/next/src/lib/multi-file-writer.ts#L56) code from Next gets run. Needs further investigation.

I think @conico974 told me some patches need to be updated for 15.4 to patch `server/route-modules/route-module.ts` instead.

Thanks for adding the info the PR description for now!

vicb (approved):
Thanks @sommeeeer !

PR #800, opened 2025-07-21
wrangler should not try to handle .env files

vicb:
wrangler will start loading .env files after [this PR](https://github.com/cloudflare/workers-sdk/pull/9914) is released

Next environments are [limited to `production`, `development`, and `test`](https://nextjs.org/docs/pages/guides/environment-variables#test-environment-variables) and they might not match the wrangler environment.

The way to define which of the .env file to load is to use the `NEXTJS_ENV` env variable.

### DIFF
--- a/.changeset/tasty-teeth-do.md
+++ b/.changeset/tasty-teeth-do.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+wrangler should not load .env files


james-elicx (approved):
probably want a changeset as well

vicb:
> probably want a changeset as well

I didn't include a changeset initially because there is actually no change in behavior and that feature has not been released yet. But you're right it still might be helpful.

Thanks for the review

Issue #803, opened 2025-07-21
[BUG] D1_ERROR: Wrong number of parameter bindings for SQL query.


alex-all3dp:
### Describe the bug

I have a worker deployed with the following open-next config:

```typescript
import {defineCloudflareConfig} from '@opennextjs/cloudflare'
import r2IncrementalCache from '@opennextjs/cloudflare/overrides/incremental-cache/r2-incremental-cache'
import {withRegionalCache} from '@opennextjs/cloudflare/overrides/incremental-cache/regional-cache'
import memoryQueue from '@opennextjs/cloudflare/overrides/queue/memory-queue'
import d1NextTagCache from '@opennextjs/cloudflare/overrides/tag-cache/d1-next-tag-cache'

const openNextConfig = {
  ...defineCloudflareConfig({
    incrementalCache: withRegionalCache(r2IncrementalCache, {
      mode: 'long-lived',
      shouldLazilyUpdateOnCacheHit: true,
    }),
    queue: memoryQueue,
    tagCache: d1NextTagCache,
  }),
  cloudflare: {
    skewProtection: {
      enabled: true,
      maxNumberOfVersions: 10, // Maximum number of previous versions to use.
      maxVersionAgeDays: 20, // Age of the oldest version to use (from the last deployment date)
    },
  },
}

export default openNextConfig
```

The app appears to be working normally but in the logs I can see the following error logs
```json
{
  "source": {
    "level": "error",
    "message": "Error: D1_ERROR: Wrong number of parameter bindings for SQL query.",
    "$cloudflare": {
      "$metadata": {
        "id": "01K0PF84ZKVSYVCJPAPP5Y4YCJ",
        "type": "cf-worker",
        "error": "Error: D1_ERROR: Wrong number of parameter bindings for SQL query.",
        "messagePattern": "Error: D1_ERROR: Wrong number of parameter bindings for SQL query."
      }
    }
  },
  "dataset": "cloudflare-workers",
  "timestamp": "2025-07-21T12:27:29.876Z",
  "$workers": {
    "truncated": false,
    "event": {
      "request": {
        "url": "https://<domain></path>?_rsc=1a9qi",
        "method": "GET",
        "path": "</path>",
        "search": {
          "_rsc": "1a9qi"
        }
      }
    },
    "outcome": "ok",
    "scriptName": "<workerName>",
    "eventType": "fetch",
    "executionModel": "stateless",
    "scriptVersion": {
[...]

PR #804, opened 2025-07-21
Fix missing SQL parameter bindings in `D1NextModeTagCache.getLastRevalidated()`

alex-all3dp:
Closes [803](https://github.com/opennextjs/opennextjs-cloudflare/issues/803)

I also noticed that `D1NextModeTagCache` was missing a unit test, so I had Copilot draft one up. If you don't want that included, let me know and I'll drop that commit.

### DIFF
--- a/.changeset/empty-garlics-lead.md
+++ b/.changeset/empty-garlics-lead.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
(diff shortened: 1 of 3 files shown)


vicb:
Thanks @alex-all3dp 
Could you please add a patch changeset (via `pnpm changeset`)

alex-all3dp:
> Thanks @alex-all3dp Could you please add a patch changeset (via `pnpm changeset`)

Done ✅

PR #807, opened 2025-07-22
Error when Next >= 15.4

vicb:
See #667 

We'll remove that when 15.4 is supported for now it will avoid having to closing too many issues as duplicates


conico974 (approved):

vicb:
Thanks for the review!

Issue #808, opened 2025-07-22
[BUG] Skew Protection not working for Simple form action


dario-piotrowicz:
### Describe the bug

After following all the instruction for skew protection from the official docs: https://opennext.js.org/cloudflare/howtos/skew

It looks to me like skew protection does not work for a simple server action triggered via `useActionState`.

### Steps to reproduce

You can find all the necessary steps to reproduce the issue here:
https://github.com/dario-piotrowicz/open-next-skew-protection-minimal-test/tree/main

### Expected behavior

This behavior should be equal to that of the application when deployed on Vercel.

### @opennextjs/cloudflare version

1.5.3

### Wrangler version

4.25.1

### next info output

```shell
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.5.0: Tue Apr 22 19:54:49 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T6000
  Available memory (MB): 32768
  Available CPU cores: 10
Binaries:
  Node: 22.17.1
  npm: 10.9.2
  Yarn: 1.22.22
  pnpm: 9.12.2
Relevant Packages:
  next: 15.3.5 // There is a newer version (15.4.3) available, upgrade recommended! 
  eslint-config-next: N/A
  react: 19.1.0
  react-dom: 19.1.0
  typescript: 5.8.3
Next.js Config:
  output: N/A
 ⚠ There is a newer version (15.4.3) available, upgrade recommended! 
   Please try the latest canary version (`npm install next@canary`) to confirm the issue still exists before creating a new issue.
   Read more - https://nextjs.org/docs/messages/opening-an-issue
```

PR #810, opened 2025-07-23
fix skew protection handling of POST requests

vicb:
fixes http://github.com/opennextjs/opennextjs-cloudflare/issues/808


vicb:
Thanks for the review and tests Dario!

Issue #801, opened 2025-07-21
Environment Variables discrepancy in Skew Protection


dario-piotrowicz:
In the [Skew Protection HowTo documentation](https://opennext.js.org/cloudflare/howtos/skew) there is a mention about environment variables that need to be set:

<img width="865" height="338" alt="Image" src="https://github.com/user-attachments/assets/12164177-24d0-4d83-a944-6f53877c1576" />

However trying to run
```sh
CF_WORKERS_SCRIPTS_API_TOKEN="<TOKEN>" pnpm cloudflare:deploy
```
does not work as the variable doesn't get detected

I think that this is due to the fact that the variables are taken from `getPlatformProxy`:
https://github.com/opennextjs/opennextjs-cloudflare/blob/d18cd9d2071248ebe5560599f88fa0552940d77a/packages/cloudflare/src/cli/commands/deploy.ts#L15-L18

If I add the variable to a `.dev.vars` file (or in the wrangler config file) it however works as expected.

So I think that either:
 - if this is expected, the docs they should clearly specify what they mean with environment variables
 - if this is unexpected then also the variables in `process.env` should be taken into account here

alex-all3dp:
I also ran into this when trying to access the api token from the worker secrets. Unfortunately it only worked when adding them to the wrangler `vars` list. 
Accounting for variables from `process.env` would be nice, since this would allow to store the token in the worker secrets?

vicb:
PRs (doc + code) welcome!

PR #818, opened 2025-07-25
fix `deploy` command not taking into account environment variables

dario-piotrowicz:
Fixes: #801

### DIFF
--- a/packages/cloudflare/src/cli/commands/deploy.ts
+++ b/packages/cloudflare/src/cli/commands/deploy.ts
@@ -27,10 +27,12 @@ export async function deployCommand(args: WithWranglerArgs<{ cacheChunkSize: num
 
 	const wranglerConfig = readWranglerConfig(args);
 
-	const envVars = await getEnvFromPlatformProxy({
+	const platformProxyEnvVars = await getEnvFromPlatformProxy({
 		configPath: args.configPath,
 		environment: args.env,
 	});
+
+	const envVars = { ...platformProxyEnvVars, ...process.env };
 
 	const deploymentMapping = await getDeploymentMapping(options, config, envVars);
 
--- a/.changeset/legal-bags-agree.md
+++ b/.changeset/legal-bags-agree.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+fix `deploy` command not taking into account environment variables


packages/cloudflare/src/cli/commands/deploy.ts
```diff
@@ -27,11 +27,13 @@ export async function deployCommand(args: WithWranglerArgs<{ cacheChunkSize: num
 
 	const wranglerConfig = readWranglerConfig(args);
 
-	const envVars = await getEnvFromPlatformProxy({
+	const platformProxyEnvVars = await getEnvFromPlatformProxy({
 		configPath: args.configPath,
 		environment: args.env,
 	});
 
+	const envVars = { ...platformProxyEnvVars, ...process.env };
```
vicb: `upload.ts` should be updated as well.

May be we could add a `includeProcessEnv: Boolean` to `getEnvFromPlatformProxy` so that we could document the purpose on the JSDoc of that functon.

I'm wondering if it should `{ ...platformProxyEnvVars, ...process.env }` or `{ ...process.env, ...platformProxyEnvVars}` - I would tend to prefer the latest as proxy vars are more specific as they could change with the environment but I don't have a very strong opinion on that. What do you think?

vicb: Actually I'm wondering if we should rather update

https://github.com/opennextjs/opennextjs-cloudflare/blob/c46eeee8f70032708f71dfb37a40fe8b1b805507/packages/cloudflare/src/cli/commands/skew-protection.ts#L68-L86

to read from either `process.env` of `envVars`.

That would be a more targeted change.

What are your thoughts?

dario-piotrowicz: > `upload.ts` should be updated as well.

Updated 👍 

> May be we could add a `includeProcessEnv: Boolean` to `getEnvFromPlatformProxy` so that we could document the purpose on the JSDoc of that functon.

What about renaming the function `getEnvironmentVariables` and have it always include values from both `platformProxy` and `process.env`? would there be an issue with that?

> I'm wondering if it should `{ ...platformProxyEnvVars, ...process.env }` or `{ ...process.env, ...platformProxyEnvVars}` - I would tend to prefer the latest as proxy vars are more specific as they could change with the environment but I don't have a very strong opinion on that. What do you think?

I purposely put `...process.env` as last since I think system environment variables should have precedence over those in the config files, for example if I run something line:
```sh
$ CF_ACCOUNT_ID="123" npm run deploy
```
I would personally expect that "123" would override a potential `CF_ACCOUNT_ID` value present in the config file.

I do think that this is the common behavior for this sort of things. Having the system environment variables also adds more flexibility to users, that could for example very easily deploy their worker onto different accounts (silly example):
```sh
$ CF_ACCOUNT_ID="123" npm run deploy
$ CF_ACCOUNT_ID="456" npm run deploy 
```

What would the benefits in having the `platformProxy` have precedence over `process.env` be?

dario-piotrowicz: > Actually I'm wondering if we should rather update
> 
> https://github.com/opennextjs/opennextjs-cloudflare/blob/c46eeee8f70032708f71dfb37a40fe8b1b805507/packages/cloudflare/src/cli/commands/skew-protection.ts#L68-L86
> 
> to read from either `process.env` of `envVars`.
> 
> That would be a more targeted change.
> 
> What are your thoughts?

I think that generally it would make more sense to get the environment variables from both `getPlatformProxy` and `process.env`, actually generally `process.env` makes more sense to me if we're talking about build time variables, since the variables in `getPlatformProxy` are, in my opinion, runtime variables.

But if you prefer I'll update the code to limit the change to the skew protection logic.

vicb: > What about renaming the function getEnvironmentVariables and have it always include values from both platformProxy and process.env? would there be an issue with that?

From the discussion, I think `getEnvFromPlatformProxy` and have it returns only the the env. This is the original use case and what we usually want - i.e. when we access the cache at build time, we want to use the config that will be used at runtime.

So I think what we want is to pass the `getEnvFromPlatformProxy()` and `getDeployementMapping()` (i.e. no change from what we do today).

Inside this function you could add `env = {...envVars, ...process.env};` with a comment that process.env has priority for those variables (agreed that `process.env` should have priority)

dario-piotrowicz: done, hopefully this is what you had in mind: https://github.com/opennextjs/opennextjs-cloudflare/pull/818/commits/945f849d014ecaf6837297df5e119741c471036a

vicb (approved):
Thanks!

Could you please comment the change at https://opennext.js.org/cloudflare/howtos/skew#:~:text=%7D-,Environment%20variables,-The%20following%20environment

dario-piotrowicz:
> Thanks!
> 
> Could you please comment the change at https://opennext.js.org/cloudflare/howtos/skew#:~:text=%7D-,Environment%20variables,-The%20following%20environment

How would you like me to update the docs there? I think that they are correct already? would you like me directly specify that environment variables are taken from both the platform proxy and the system env variables?

vicb:
> > Thanks!
> > Could you please comment the change at https://opennext.js.org/cloudflare/howtos/skew#:~:text=%7D-,Environment%20variables,-The%20following%20environment
> 
> How would you like me to update the docs there? I think that they are correct already? would you like me directly specify that environment variables are taken from both the platform proxy and the system env variables?

Yes.

The doc says "environment variables", we should clarify what this means wrt "process environment variables" and "wrangler environment variables".

PR #820, opened 2025-07-25
Fix regional cache for the DOShardedTagCache

### LINKED ISSUE #819: Incorrect regional tag cache for DOShardedTagCache
The regional tag cache is incorrect right now.
It is storing a boolean to know if an entry has been revalidated. 
It cannot work properly as it indirectly link a single entry in the incremental cache to a tag cache entry.

Multiple incremental cache entry accessed in the same datacenter, accesed around the same time could end up being considered valid while they're stale.

conico974:
Fix regional cache for the DOShardedTagCache.
I open it as a draft PR for now.
Fix #819 

This implementation will store every single tag in their own cache entry.

~~I think we should add an option to the tag cache.~~
~~It may make sense for some cases to store all the tag from `hasBeenRevalidated` in the same cache entry. It could be faster, but will likely ends up with less cache hit.~~
After some thought, i don't think it's worth it, the only times where it could be useful would be if you don't use (or barely) the data cache, and that you have very few ISR/SSG cache entry. Maybe an improvement to do in another PR.

### DIFF
--- a/packages/cloudflare/src/api/overrides/tag-cache/do-sharded-tag-cache.ts
+++ b/packages/cloudflare/src/api/overrides/tag-cache/do-sharded-tag-cache.ts
@@ -5,8 +5,8 @@ import { IgnorableError } from "@opennextjs/aws/utils/error.js";
 
 import type { OpenNextConfig } from "../../../api/config.js";
 import { getCloudflareContext } from "../../cloudflare-context";
-import { debugCache, purgeCacheByTags } from "../internal";
 import { DOShardedTagCache } from "../../durable-objects/sharded-tag-cache.js";
+import { debugCache, purgeCacheByTags } from "../internal";
 
 export const DEFAULT_WRITE_RETRIES = 3;
 export const DEFAULT_NUM_SHARDS = 4;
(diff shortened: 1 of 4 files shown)


packages/cloudflare/src/api/overrides/tag-cache/do-sharded-tag-cache.ts
```diff
@@ -294,28 +294,29 @@ class ShardedDOTagCache implements NextModeTagCache {
...
-							debug("Error while parsing cached value", e);
-							// If we can't parse the cached value, we should just ignore it and go to the durable object
-						}
+					const cachedValue = await this.getFromRegionalCache({ doId, tags });
+					// If all the value were found in the regional cache, we can just return the max value
+					if (cachedValue.length === tags.length) {
+						return Math.max(...cachedValue.map((item) => item.time));
 					}
+					// Otherwise we need to check the durable object on the ones that were not found in the cache
+					const filteredTags = tags.filter((tag) => !cachedValue.some((item) => item.tag === tag));
```
vicb:
```suggestion
					const filteredTags = deduplicatedTags.filter((tag) => !cachedValue.some((item) => item.tag === tag));
```

vicb (approved):
LGTM, one nit
Thanks for the fix!

vicb:
Could you please add a changeset

PR #824, opened 2025-07-27
Add option to gen sourcemaps for handler.mjs

cozzbp:


packages/cloudflare/src/cli/build/bundle-server.ts
```diff
@@ -81,6 +81,7 @@ export async function bundleServer(buildOpts: BuildOptions): Promise<void> {
 		minifySyntax: !debug,
 		legalComments: "none",
 		metafile: true,
+		sourcemap: getOpenNextConfig(buildOpts).cloudflare?.sourcemap,
```
vicb, line 84: May be we can always generate the source maps.

The Next build dominates the build time so the build time should not change a lot

```suggestion
		sourcemap: true,
```

brian-wemod: I'm just interested in figuring out a way to have server side sourcemaps for readable stack traces on cloudflare.  Trying to figure out how to make this happen.

FriendlyCityJohnstown:
Measured data on this exact gap — including the `handler.mjs` sourcemap that `OPEN_NEXT_DEBUG` *does* generate, and what it costs — is now in #1352: https://github.com/opennextjs/opennextjs-cloudflare/issues/1352#issuecomment-5671274412

Short version: the flag **does** produce the map (100 → 712 sources, real `lib/**` modules) for only **+2.0% gzip**. But it cannot be used as a build-only switch: `@opennextjs/aws` binds `minify` and `sourcemap` to `debug` (`build/helper.js:24,34,91,122`), so it also switches production into debug logging and adds `X-OpenNext-Version` / `X-OpenNext-RequestId` to every response.

PR #825, opened 2025-07-28
Add option to persist missing tags in regional cache

conico974:
Introduce a configuration option to allow persisting missing tags in the regional tag cache.
This could help reduce the load for the tag cache, but should only be used when the regional cache is properly purged on tag write.


vicb (approved):
Thanks!

Could you please update the doc repo

conico974:
Thanks for the review, I'll update the docs later today

PR #830, opened 2025-07-31
detect more image types

vicb:
Sync of <https://github.com/vercel/next.js/pull/82118>

### DIFF
--- a/packages/cloudflare/src/cli/templates/images.ts
+++ b/packages/cloudflare/src/cli/templates/images.ts
@@ -167,12 +167,16 @@ const AVIF = "image/avif";
 const WEBP = "image/webp";
 const PNG = "image/png";
 const JPEG = "image/jpeg";
+const JXL = "image/jxl";
+const JP2 = "image/jp2";
+const HEIC = "image/heic";
 const GIF = "image/gif";
 const SVG = "image/svg+xml";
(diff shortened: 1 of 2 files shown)


packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -167,12 +167,16 @@ const AVIF = "image/avif";
...
+const JXL = "image/jxl";
+const JP2 = "image/jp2";
+const HEIC = "image/heic";
 const GIF = "image/gif";
 const SVG = "image/svg+xml";
 const ICO = "image/x-icon";
 const ICNS = "image/x-icns";
 const TIFF = "image/tiff";
 const BMP = "image/bmp";
+const PDF = "application/pdf";
```
sommeeeer, line 184: PDF should not be rendered(i.e it should error). Currently it will be rendered if it gets detected in `detectContentType`. Perhaps add a small comment above it here aswell.

vicb: Good catch, for now Next blocks non "image/..." types.
I'll rework the logic.

vicb:
> LGTM, just make sure PDF is not accepted

Good catch, I have updated the PR, PTAL

sommeeeer (approved):
LGTM

PR #832, opened 2025-08-01
fix: --noMinify stop minifying the bundle

vicb:
`pnpm opennextjs-cloudflare build --noMinify` will stop minifying the bundle.

It helps with debugging

/cc @sommeeeer


vicb:
Thanks for the review @sommeeeer !

PR #833, opened 2025-08-01
Update the patches for Next 15.4

### LINKED ISSUE #667: Page Router api route fails on latest canary

sommeeeer:
fixes #667 

Patch Next `>= 15.4` for [`getIncrementalCache`](https://github.com/vercel/next.js/blob/c8c9bef3a1dd1a95434c1865dbd7f69985ea7d54/packages/next/src/server/route-modules/route-module.ts#L389-L437) in `/next/src/server/route-modules/route-module.ts`

Its pretty much the same as in the `next-server` patch, we just needed to update the rule and filter. These files were affected:
```
/next/dist/compiled/next-server/app-page.runtime.prod.js
/next/dist/compiled/next-server/pages.runtime.prod.js
/next/dist/compiled/next-server/app-route.runtime.prod.js
```

We might also have to make a patch for `loadCustomCacheHandlers` in the same files, however from my testing it seems to be already initialized therefor it will early return and log `cache handlers already initialized` if you enable `NEXT_PRIVATE_DEBUG_CACHE=1`. From what I can see it gets initialized in NextServer's `loadCustomCacheHandlers` beforehand.

### DIFF
--- a/.changeset/rude-sloths-march.md
+++ b/.changeset/rude-sloths-march.md
@@ -2,4 +2,4 @@
 "@opennextjs/cloudflare": patch
 ---
 
-add(patch): Patch getIncrementalCache in route-module.ts for ISR support in Next.js 15.4"
+Update the patches to support Next.js 15.4"


packages/cloudflare/src/cli/build/patches/plugins/route-module.ts
```diff
@@ -0,0 +1,61 @@
+/**
+ * Patch for `next/src/server/route-modules/route-module.ts`
+ * https://github.com/vercel/next.js/blob/c8c9bef3a1dd1a95434c1865dbd7f69985ea7d54/packages/next/src/server/route-modules/route-module.ts#L389-L437
```
vicb: That should be enough
```suggestion
 * https://github.com/vercel/next.js/blob/c8c9bef/packages/next/src/server/route-modules/route-module.ts#L389-L437
```

packages/cloudflare/src/cli/build/build.ts
```diff
@@ -99,11 +99,11 @@ function ensureNextjsVersionSupported(options: buildHelper.BuildOptions) {
 	}
 	// TODO: remove when 15.4 is supported
```
vicb: We can remove that when the PR becomes stable

.changeset/rude-sloths-march.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+add(patch): Patch getIncrementalCache in route-module.ts for ISR support in Next.js 15.4"
```
vicb: We should update the Next version of the e2e catalog to the latest 15.4

packages/cloudflare/src/cli/build/patches/plugins/route-module.ts
```diff
@@ -0,0 +1,61 @@
...
+ */
+export function getIncrementalCacheRule(handlerPath: string) {
+	return `
+rule:
+  pattern: "let $A, { cacheHandler: $B } = $C"
+  inside:
+    kind: method_definition
+    has:
+      field: name
+      regex: getIncrementalCache
```
vicb: Why not using an anchored regexp?

packages/cloudflare/src/cli/build/patches/plugins/route-module.ts
```diff
@@ -0,0 +1,61 @@
...
+ * cache handler and create a new instance of it.
+ *
+ * This is problematic in workerd due to the dynamic import of the file that is not known from
+ * build-time. Therefore, we have to manually override the default way that the cache handler is
+ * instantiated with a dynamic require that uses a string literal for the path.
+ */
+export function getIncrementalCacheRule(handlerPath: string) {
+	return `
+rule:
+  pattern: "let $A, { cacheHandler: $B } = $C"
```
vicb: Maybe `$A` -> `$CACHE_HANDLER` and `$B` -> `$HANDLER_PATH`

packages/cloudflare/src/cli/build/patches/plugins/route-module.ts
sommeeeer: Yeah it should be anchored.

sommeeeer:
This one is ready to go. There is atleast one more patch that needs update. If you build the `examples/playground15` with `next@15.4` and this PR you will get this error when you visit `http://localhost:8787/isr/hello/dynamic`:
`✘ [ERROR] TypeError: Class constructor Cache cannot be invoked without 'new`

It wont render the correct 500 html page either.

packages/cloudflare/src/cli/build/patches/plugins/route-module.ts
```diff
@@ -0,0 +1,61 @@
...
+import { patchCode } from "@opennextjs/aws/build/patch/astCodePatcher.js";
+import type { ContentUpdater, Plugin } from "@opennextjs/aws/plugins/content-updater.js";
+import { getCrossPlatformPathRegex } from "@opennextjs/aws/utils/regex.js";
+
+import { normalizePath } from "../../utils/index.js";
+
+export function patchRouteModules(updater: ContentUpdater, buildOpts: BuildOptions): Plugin {
+	return updater.updateContent("route-module", [
+		{
+			filter: getCrossPlatformPathRegex(String.raw`/next/dist/compiled/next-server/.*\.runtime\.prod\.js$`, {
```
vicb: nit: lazy match
```suggestion
			filter: getCrossPlatformPathRegex(String.raw`/next/dist/compiled/next-server/.*?\.runtime\.prod\.js$`, {
```

.changeset/rude-sloths-march.md
vicb:
```suggestion
Update the patches to support Next.js 15.4"
```

vicb (approved):
LGTM!

Thanks @sommeeeer

packages/cloudflare/src/cli/build/bundle-server.ts
```diff
@@ -149,6 +149,8 @@ export async function bundleServer(buildOpts: BuildOptions, projectOpts: Project
 			"process.env.TURBOPACK": "false",
 			// This define should be safe to use for Next 14.2+, earlier versions (13.5 and less) will cause trouble
 			"process.env.__NEXT_EXPERIMENTAL_REACT": `${needsExperimentalReact(nextConfig)}`,
+			// Fix `res.validate` in Next 15.4 (together with the `route-module` patch)
+			"process.env.__NEXT_TRUST_HOST_HEADER": "true",
```
sommeeeer, line 153: Did you set this one just in case? It didn't really have any effect.

vicb: Yep

PR #843, opened 2025-08-04
perf: drop the babel dependency

vicb:
The [former patch of next server is not enough](https://github.com/opennextjs/opennextjs-aws/blob/e1048ac9bfffa79d32b1cab810619c41a66187c3/packages/open-next/src/build/patch/patches/patchNextServer.ts#L94).

The code was changed sometimes in Next 15.

We should upstream that to aws.

Before (e2e/experimental)

image

After

image

/cc @conico974 @sommeeeer

### DIFF
--- a/.changeset/neat-jars-listen.md
+++ b/.changeset/neat-jars-listen.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+perf: drop the babel dependency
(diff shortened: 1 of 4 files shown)


sommeeeer (approved):
LGTM, does this get rid of that pesky switch case warning? Should you only apply it on Next >= 15.4?

vicb:
> LGTM, does this get rid of that pesky switch case warning? 

Yes, that warning is removed.

> Should you only apply it on Next >= 15.4?

I think somebody reported it for 15.3.<I don't remember what> last week

vicb:
> > Should you only apply it on Next >= 15.4?
> 
> I think somebody reported it for 15.3.<I don't remember what> last week

actually it was 15.2.5

vicb:
Oh and thanks for another review 🙏

conico974 (approved):
LGTM! And yeah we should add it to aws as well

conico974:
You guys are too fast :smile:

LubomirGeorgiev:
@vicb 

This helped a lot. It significantly reduced the worker size.

Before:
Total Upload: 25229.97 KiB / gzip: 6136.94 KiB

After:
Total Upload: 22279.11 KiB / gzip: 5475.65 KiB


Do you have in mind other similar optimizations like this one?

PR #849, opened 2025-08-13
chore: enforce the use of extensions for all the imports in src files

dario-piotrowicz:
Most of the imports in src files do include the `.js` extension, only some don't for some reason (those are likely unintentional), for consistency I think all imports should include `.js` extensions

### DIFF
--- a/packages/cloudflare/src/cli/commands/skew-protection.ts
+++ b/packages/cloudflare/src/cli/commands/skew-protection.ts
@@ -32,7 +32,7 @@ import type { VersionGetResponse } from "cloudflare/resources/workers/scripts/ve
 
 import type { OpenNextConfig } from "../../api/index.js";
 import { CURRENT_VERSION_ID, DEPLOYMENT_MAPPING_ENV_NAME } from "../templates/skew-protection.js";
-import type { WorkerEnvVar } from "./helpers";
+import type { WorkerEnvVar } from "./helpers.js";
 
 /** Maximum number of versions to list */
 const MAX_NUMBER_OF_VERSIONS = 20;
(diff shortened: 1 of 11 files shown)


packages/cloudflare/src/cli/commands/skew-protection.ts
```diff
@@ -28,11 +28,11 @@ import { loadConfig } from "@opennextjs/aws/adapters/config/util.js";
...
 import logger from "@opennextjs/aws/logger.js";
 import { Cloudflare, NotFoundError } from "cloudflare";
-import type { VersionGetResponse } from "cloudflare/resources/workers/scripts/versions";
+import type { VersionGetResponse } from "cloudflare/resources/workers/scripts/versions.js";
 
-import type { OpenNextConfig } from "../../api";
+import type { OpenNextConfig } from "../../api/index.js";
 import { CURRENT_VERSION_ID, DEPLOYMENT_MAPPING_ENV_NAME } from "../templates/skew-protection.js";
-import type { WorkerEnvVar } from "./helpers.js";
+import type { WorkerEnvVar } from "./helpers";
```
james-elicx: surprised this one isn't failing the linting

dario-piotrowicz: yeah... 😕

dario-piotrowicz: Ah I see, `import type` imports don't seem to be detected... I'll have a look if there's an option for that 👀

dario-piotrowicz: Fixed, I needed to set the `checkTypeImports` option 🙂 , thanks for spotting this 🙏

dario-piotrowicz:
Thanks for the reviews 😄 🙏

PR #850, opened 2025-08-13
Add option for regional cache to skip tagCache on cache hits

dario-piotrowicz:


packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
```diff
@@ -38,6 +38,15 @@ type Options = {
 	 * @default `false` for the `short-lived` mode, and `true` for the `long-lived` mode.
 	 */
 	shouldLazilyUpdateOnCacheHit?: boolean;
+
+	/**
+	 * Whether on cache hits the tagCache should be skipped or not. Skipping the tagCache allows requests to be
+	 * handled faster, the downside of this is that you need to make sure that the cache gets correctly purged
+	 * either by enabling the auto cache purging feature or doing that manually.
+	 *
+	 * @default `true` is the auto cache purging is enabled, `false` otherwise.
```
conico974:
```suggestion
	 * @default `true` if the auto cache purging is enabled, `false` otherwise.
```

packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
```diff
@@ -65,6 +74,17 @@ class RegionalCache implements IncrementalCache {
 		this.opts.shouldLazilyUpdateOnCacheHit ??= this.opts.mode === "long-lived";
 	}
 
+	get #bypassTagCacheOnCacheHit(): boolean {
+		if (this.opts.bypassTagCacheOnCacheHit) {
```
conico974: This means that people cannot disable it if they're using the automatic cache purge.
We should probably allow that, especially for preview domain

conico974 (approved):
LGTM, maybe just add some comment about the need for a custom domain

packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
dario-piotrowicz: yeah I also just noticed this! thanks for spotting it! 🙏

packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
dario-piotrowicz: sorry yeah my bad, the check was meant to be `this.opts.bypassTagCacheOnCacheHit !== undefined`, I'll update this, thanks a lot for noticing the mistake! 🙏

dario-piotrowicz: fixed 🙂

packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
```diff
@@ -65,6 +74,17 @@ class RegionalCache implements IncrementalCache {
 		this.opts.shouldLazilyUpdateOnCacheHit ??= this.opts.mode === "long-lived";
 	}
 
+	get #bypassTagCacheOnCacheHit(): boolean {
+		if (this.opts.bypassTagCacheOnCacheHit !== undefined) {
+			// If the bypassTagCacheOnCacheHit option is set we return that one
+			return this.opts.bypassTagCacheOnCacheHit;
+		}
+
+		// Otherwise we default to wether the automatic cache purging is enabled or not
```
james-elicx:
```suggestion
		// Otherwise we default to whether the automatic cache purging is enabled or not
```

packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
```diff
@@ -103,7 +126,7 @@ class RegionalCache implements IncrementalCache {
 				this.putToCache({ key, cacheType, entry: { value, lastModified } })
 			);
 
-			return { value, lastModified };
+			return { value, lastModified, shouldBypassTagCache: this.#bypassTagCacheOnCacheHit };
```
conico974: It should not be here, otherwise it means that it will not revalidate properly for call to `revalidateTag`

dario-piotrowicz: ok, removed, thanks! 🙏

packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
```diff
@@ -91,7 +111,10 @@ class RegionalCache implements IncrementalCache {
 					);
 				}
 
-				return cachedResponse.json();
+				return {
+					...cachedResponse.json(),
```
conico974: That's why the test are failing. We should await that

dario-piotrowicz: yes of course 🤦, not sure how I made this mistake, thanks 🙏

PR #858, opened 2025-08-25
bump @opennextjs/aws to 3.7.6

vicb:


conico974 (approved):

james-elicx (approved):

vicb:
Thanks for the reviews!
Look like there is a legit issue, I'll investigate

dario-piotrowicz (approved):

sommeeeer:
https://github.com/opennextjs/opennextjs-cloudflare/blob/c882339/examples/e2e/app-router/app/api/after/ssg/route.ts

This test might need to change into a page. The Cache Interceptor will return the prerendered data for a static route everytime which is correct according to Next docs. 

Currently it should not be supported to revalidate a route handler that is `force-static`. If you `next start` the `app-router` you will see that its not possible to revalidate that route by sending a `POST` to `api/after/revalidate`. 

By reading [here](https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#dynamic) they are claiming under `force-static` that it should only be possible to revalidate pages or layouts. Maybe we should turn that test into being a `page.tsx` instead?

conico974:
It's a bug here https://github.com/opennextjs/opennextjs-aws/blob/9a83eb754ca82f830b1cc06d2d1470cd19173e0b/packages/open-next/src/core/routing/cacheInterceptor.ts#L231
We should check route type here as well

vicb:
The issue was fixed in https://github.com/opennextjs/opennextjs-aws/pull/959.

We will bump to the next aws release when available

vicb:
Updated the PR to use aws 3.7.6

vicb:
Thank everybody for the review/help getting this in!

PR #862, opened 2025-08-31
feat: support for a custom OpenNext config path

james-elicx:
adds a cli flag that lets you use a custom config path (e.g. if you had a cli that internally called opennext, you could have your own build dir have the config path instead of it leaking out).

outcome of #853


vicb (changes requested):
Let's discuss the open point in https://github.com/opennextjs/opennextjs-aws/pull/972 before we merge this

vicb:
@james-elicx could you please update the PR description to include:

- dependencies - Is https://github.com/opennextjs/opennextjs-aws/pull/972 still a dependency needing to be merged/released before we can merge this
- the list of changes in this PR (i.e. support for custom Open Next config path, deprecation of `--configPath`, anything else?)

We can keep #871 around until we remove `--configPath` in a coming major release (or close it and create a separate issue)

Thanks!

james-elicx:
Yes, that's a dependency for absolute paths to work with the cloudflare cli. Updated the PR desc.

vicb:
@james-elicx I have added a proposal as a fixup commit.

Main changes:
- Added a hack to avoid having to wait for the aws PR to be merged
- Do not offer to create a config file when a custom path is used (assuming it's user error then)

Looking at the comments/changes in `compileConfig` is probably the best way grasp the changes.

What do you think?

packages/cloudflare/src/cli/commands/utils.ts
```diff
@@ -33,14 +33,27 @@ export function printHeaders(command: string) {
 }
 
 /**
- * Compile the OpenNext config, and ensure it is for Cloudflare.
+ * Compile the OpenNext config.
+ *
+ * When users do not specify a custom config file (using `----openNextConfigPath`),
```
james-elicx:
```suggestion
 * When users do not specify a custom config file (using `--openNextConfigPath`),
```

james-elicx:
> * Added a hack to avoid having to wait for the aws PR to be merged

This is why I used to keep asking when we'd get a monorepo :(

Personally, I prefer not adding hacks simply because we're using polyrepos, even if that means it might take longer to ship something.

> * Do not offer to create a config file when a custom path is used (assuming it's user error then)

Makes sense, looks fine.

vicb:
Thanks for working on this James!

> This is why I used to keep asking when we'd get a monorepo :(
> 
> Personally, I prefer not adding hacks simply because we're using polyrepos, even if that means it might take longer to ship something.

It's not always easy to balance priority across features, fixes, customer wishes, and resources.

The hack is tested and should not be here for long so I think it's safe to merge this for now.

As for the mono-repo, definitely something we should think about before jumping into the Adapters API effort.
It might be the right time.

PR #863, opened 2025-09-01
Use  `patchNodeEnvironment` from aws

vicb:
`patchNodeEnvironment` was introduced in #843 to drop the babel dependency. It was upstream into aws via https://github.com/opennextjs/opennextjs-aws/pull/948. This PR drops the local version to use the upstream version.

I haven't added any changeset as there is no behavior change.


conico974 (approved):

sommeeeer (approved):

vicb:
Thanks for the reviews!

PR #865, opened 2025-09-01
add(patch): Ensure request.url is with correct protocol in route handlers

sommeeeer:
This is for https://github.com/opennextjs/opennextjs-aws/issues/969

We patch NextServer's `attachRequestMeta` as it will be called in `handleRequestImpl`. In theory we can also just comment out that line as we pass in `initURL` to `nextServer.getRequestHandlerWithMetadata` arguments already [here](https://github.com/opennextjs/opennextjs-aws/blob/fe913bb/packages/open-next/src/core/requestHandler.ts#L259) when we process the request through NextServer.

However, I went with this approach. Has been tested on Next 14 aswell. 

I did update the tests that relied on this aswell.

### DIFF
--- a/packages/cloudflare/src/cli/build/patches/plugins/next-server.ts
+++ b/packages/cloudflare/src/cli/build/patches/plugins/next-server.ts
@@ -42,6 +42,8 @@ export function patchNextServer(updater: ContentUpdater, buildOpts: BuildOptions
 				// Node middleware are not supported on Cloudflare yet
 				contents = patchCode(contents, disableNodeMiddlewareRule);
 
+				contents = patchCode(contents, attachRequestMetaRule);
+
 				return contents;
 			},
 		},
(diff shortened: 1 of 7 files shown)


.changeset/afraid-sloths-shave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+add(patch): Ensure request.url is with correct protocol in route handlers
```
vicb:
```suggestion
fix: Ensure request.url reflects the actual host protocol in route handlers
```

packages/cloudflare/src/cli/build/patches/plugins/next-server.ts
```diff
@@ -118,3 +120,42 @@ fix: |-
...
+ *
+ * initUrl will always be with `https` cause this.fetchHostname && this.port is undefined in our case.
+ * this.nextConfig.experimental.trustHostHeader is also true.
+ *
+ * Callstack: handleRequest-> handleRequestImpl -> attachRequestMeta
+ *
+ * We actually set this `initURL` already in nextServer.getRequestHandlerWithMetadata(metadata) as seen here:
+ * https://github.com/opennextjs/opennextjs-aws/blob/fe913bb/packages/open-next/src/core/requestHandler.ts#L259
+ *
+ */
```
vicb:
```suggestion
/**
 * https://github.com/vercel/next.js/blob/ea08bf27/packages/next/src/server/next-server.ts#L1916-L1923
 *
 * `initUrl` always uses `https` because `this.fetchHostname && this.port` is `undefined` in our case.
 * `this.nextConfig.experimental.trustHostHeader` is also true.
 *
 * Callstack: `handleRequest` -> `handleRequestImpl` -> `attachRequestMeta`
 *
 * We actually set this `initURL` already in `nextServer.getRequestHandlerWithMetadata(metadata)` as seen here:
 * https://github.com/opennextjs/opennextjs-aws/blob/fe913bb/packages/open-next/src/core/requestHandler.ts#L259
 */
```

vicb: Could you please add to this comment to describe the before/after code of this rule?

vicb (approved):
Thanks a lot @sommeeeer for fixing this long standing issue.

I have added a couple of minor comments inline. The PR is good to go when those are addressed.

sommeeeer:
Thanks for the review Victor!

PR #867, opened 2025-09-02
Log cache purge errors as errors

vicb:
Cache purge errors when only logs when debugging Next cache.
They are now log unconditionally as errors.

Other minor changes:
- disable `cachePurge` in the `app-router` e2e (no token / zone id is provided)
- clarify the inline documentation


conico974 (approved):

vicb:
Thanks for the review Nico!

PR #872, opened 2025-09-03
Fix check for missing `CACHE_PURGE_ZONE_ID`

alex-all3dp:
The credentials check in `internalPurgeCacheByTags` only skips the operation if both required env vars are missing. However when an API token is provided but `CACHE_PURGE_ZONE_ID` is missing (or vice versa) it will proceed to execute a failing purge request.

Log example with missing zone id but existing token:

Bildschirmfoto 2025-09-03 um 12 11 53


vicb (approved):
Good catch, thanks!

PR #873, opened 2025-09-03
fix(e2e): Ensure the response event listener is registered before navigation

sommeeeer:
The event listener on these tests should be registered before the navigation. Before this PR these tests would never invoke the callback. 

When you have enabled the cache interception, the rewrite status code from middleware would not be propogated to the response. There is a follow-up PR coming in AWS to take care of this problem.

Also added a test for `x-opennext-requestid`.

Depends on https://github.com/opennextjs/opennextjs-aws/pull/974

**DO NOT MERGE BEFORE ^**

Note: As you can see in the [job](https://github.com/opennextjs/opennextjs-cloudflare/actions/runs/17442182322/job/49527671671#step:7:320) on this PR this test fails, cause we need to propogate the rewrite status code properly in the cache interceptor. We can rerun the E2E on this PR, once the fix in AWS is released and we have bumped to it.

### DIFF
--- a/examples/e2e/app-router/e2e/middleware.rewrite.test.ts
+++ b/examples/e2e/app-router/e2e/middleware.rewrite.test.ts
@@ -1,4 +1,4 @@
-import { expect, test } from "@playwright/test";
+import { expect, test, Response as PwResponse } from "@playwright/test";
 import { validateMd5 } from "../../utils";
 
 /*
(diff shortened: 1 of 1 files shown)


examples/e2e/app-pages-router/e2e/headers.test.ts
```diff
@@ -0,0 +1,20 @@
+import { expect, test } from "@playwright/test";
+
+test("should test if poweredByHeader adds the correct headers ", async ({ page }) => {
```
vicb:
```suggestion
test("poweredByHeader adds the correct headers ", async ({ page }) => {
```

examples/e2e/app-pages-router/e2e/headers.test.ts
```diff
@@ -0,0 +1,20 @@
...
+	const result = await page.goto("/ssg");
+	expect(result).toBeDefined();
+	expect(result?.status()).toBe(200);
+	const headers = result?.headers();
+
+	// This header should not be defined even when cache interception happens in the external middleware
+	expect(headers?.["x-opennext-requestid"]).toBeFalsy();
```
vicb: Would the following be more in line with what we want to test?

(same below and in examples/e2e/pages-router/e2e/header.test.ts)

```suggestion
	expect(headers?.["x-opennext-requestid"]).toBeUndefined();
```

examples/e2e/app-pages-router/e2e/headers.test.ts
```diff
@@ -0,0 +1,20 @@
...
+
+	await page.waitForTimeout(2000); // Wait 2s
```
vicb: Why do we need to wait here?

vicb (approved):
Thanks!

I added a few minor comments.

examples/e2e/app-pages-router/e2e/headers.test.ts
sommeeeer: Good catch. Actually, this should be `does not set x-opennext-requestid header on cache interceptor response` instead. Will come in a review commit shortly.

examples/e2e/app-pages-router/e2e/headers.test.ts
sommeeeer: I agree, this is more precise! Thanks

examples/e2e/app-pages-router/e2e/headers.test.ts
sommeeeer: Thanks again! 

We dont. I'll remove those lines. It will respond with the cached response on first request as long as you have populated the cache. When I was testing this yesterday, I didnt populate the cache first.

sommeeeer:
Thanks Victor!

vicb:
I have rebased the PR.
I was expecting it to fail but it passes.

Could you please clarify when you have time @sommeeeer ?

vicb:
Rebased on main in the last force push

Issue #879, opened 2025-09-09
No need to lazy refresh the regional cache when cache purge is used


vicb:
When i.e. "long-lived" R2 cache is used, the regional cache is refreshed from R2 after cache hits:

https://github.com/opennextjs/opennextjs-cloudflare/blob/3be30a4d33d978b14c42b8c295617e17bf7978f3/packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts#L103-L111

This is needed when cache purge is enabled.

That is `shouldLazilyUpdateOnCacheHit` should default to false when cache purge is in use.

Note that in the source code, we already check if the cache purge is enabled via:

https://github.com/opennextjs/opennextjs-cloudflare/blob/3be30a4d33d978b14c42b8c295617e17bf7978f3/packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts#L84

But this check is not correct. Having a DO binding does not mean that purge is enabled:
- there could be a DO but no cache purge override configured - in this case we assume that cache purge is used but it is not
- there could be no DO and a direct cache purge configure - in this case we assume that cache purge is not used but it actually is

A more robust way to detect if cache purge is enabled would be via `globalThis.openNextConfig`

/cc @conico974 @dario-piotrowicz

vicb:
closed via https://github.com/opennextjs/opennextjs-cloudflare/pull/882

PR #881, opened 2025-09-09
fix: detection of cache purge

vicb:
relates to https://github.com/opennextjs/opennextjs-cloudflare/issues/879

Before this PR purge cache was detected by looking for the related DO but

- There could be a DO without cache purge being enabled (i.e. no override configured)
- Purge could be enabled in direct mode with no DO


conico974 (approved):

vicb:
Thanks for the review Nico!

PR #882, opened 2025-09-09
perf: skip lazy update on cache hit when cache purge is enabled

vicb:
Syncing the Cache API on hits is not required when cache purge is enabled (they is no risk of having a cache hit if the cache API is not in sync with R2 in this case as updated entries are purged).

The change reduces the numbers of R2 reads when cache purge is enabled.


vicb:
Thanks for the review Nico!

PR #888, opened 2025-09-13
fix: remote flag not working for preview command's cache population

### LINKED ISSUE #887: [BUG] preview cli doesn't --preview flag to wrangler command
### Describe the bug

When running `opennextjs-cloudflare preview --remote` with a preview_id defined in wrangler.toml, OpenNext attempts to populate the KV cache via `wrangler kv:bulk put`, but it does not pass the required `--preview` flag to Wrangler. So we get error of this

[ERROR] NEXT_INC_CACHE_KV has both a namespace ID and a preview ID. Specify "--preview" or "--preview false" to avoid writing data to the wrong namespace.

### Steps to reproduce

1. Add KV for cache in wrangler config

```
[[kv_namespaces]]
binding = "NEXT_INC_CACHE_KV"
id = "xxxxxxxxxxxxxx"
preview_id = "xxxxxxxxxx"
```

(preview_id seems required by wrangler, otherwise you will get error: 
In development, you should use a separate kv namespace than the one you'd use in production. Please create a new kv namespace with "wrangler kv namespace create <name> --preview" and add its id as preview_id to the kv_namespace "NEXT_INC_CACHE_KV" in your wrangler.toml file)

2. enable kvIncrementalCach in opennext config

```
});

```

3. run `opennextjs-cloudflare build && opennextjs-cloudflare preview --remote`

error get:
NEXT_INC_CACHE_KV has both a namespace ID and a preview ID. Specify "--preview" or "--preview false" to avoid writing data to the wrong namespace.

### Expected behavior

Opennext should set the --preview flag to wrangler put command, or provides a way for us to pass custom wrangler args from cli.
(I can't find how to pass through args in doc, please let me know if I missed it)

1.6.1

### Wrangler version

4.26

### next info output

```shell
Using vars defined in .dev.vars

Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.5.0: Tue Apr 22 19:54:29 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T6030
  Available memory (MB): 18432
  Available CPU cores: 12
Binaries:
  Node: 22.16.0
  npm: 10.9.2
  Yarn: 1.22.22
  pnpm: 10.14.0
Relevant Packages:
  next: 15.3.2 // There is a newer version (15.5.3) available, upgrade recommended! 
  eslint-config-next: 15.3.2
  react: 19.1.0
  react-dom: 19.1.0
  typescript: 5.8.3
Next.js Config:
  output: N/A
 ⚠ There is a newer version (15.5.3) available, upgrade recommended! 
```

### Additional context

[Cloudflare KV config doc](https://developers.cloudflare.com/workers/wrangler/configuration/#kv-namespaces)

james-elicx:
When passing the `--remote` flag to our `preview` command, it would fail to properly populate the cache due to not targeting the remote bindings and missing a `--preview` flag for Workers KV / D1.

fixes #887

**Changes:**

- Support both remote and local as modes for cache population in the preview command.
- Pass the preview flag for KV / D1 cache population for the preview command.

Tested that cache population was working as expected with the following commands for the `d1-tag-next` example:
- `opennextjs-cloudflare preview --config wrangler.e2e.jsonc`
- `opennextjs-cloudflare preview --config wrangler.e2e.jsonc --remote`
- `opennextjs-cloudflare deploy --config wrangler.e2e.jsonc`

### DIFF
--- a/packages/cloudflare/src/cli/commands/populate-cache.ts
+++ b/packages/cloudflare/src/cli/commands/populate-cache.ts
@@ -109,6 +109,7 @@ type PopulateCacheOptions = {
 	environment?: string;
 	wranglerConfigPath?: string;
 	cacheChunkSize?: number;
+	shouldUsePreviewNamespace?: boolean;
 };
 
 async function populateR2IncrementalCache(
@@ -197,12 +198,21 @@ async function populateKVIncrementalCache(
 
 		writeFileSync(chunkPath, JSON.stringify(kvMapping));
 
-		runWrangler(options, ["kv bulk put", quoteShellMeta(chunkPath), `--binding ${KV_CACHE_BINDING_NAME}`], {
-			target: populateCacheOptions.target,
-			environment: populateCacheOptions.environment,
-			configPath: populateCacheOptions.wranglerConfigPath,
-			logging: "error",
-		});
+		runWrangler(
+			options,
+			[
+				"kv bulk put",
+				quoteShellMeta(chunkPath),
+				`--binding ${KV_CACHE_BINDING_NAME}`,
+				...(populateCacheOptions.shouldUsePreviewNamespace ? ["--preview"] : ["--preview false"]),
+			],
+			{
+				target: populateCacheOptions.target,
+				environment: populateCacheOptions.environment,
+				configPath: populateCacheOptions.wranglerConfigPath,
+				logging: "error",
+			}
+		);
 
 		rmSync(chunkPath);
 	}
@@ -228,6 +238,7 @@ function populateD1TagCache(
 			"d1 execute",
 			D1_TAG_BINDING_NAME,
 			`--command "CREATE TABLE IF NOT EXISTS revalidations (tag TEXT NOT NULL, revalidatedAt INTEGER NOT NULL, UNIQUE(tag) ON CONFLICT REPLACE);"`,
+			...(populateCacheOptions.shouldUsePreviewNamespace ? ["--preview"] : ["--preview false"]),
 		],
 		{
 			target: populateCacheOptions.target,
--- a/packages/cloudflare/src/cli/commands/preview.ts
+++ b/packages/cloudflare/src/cli/commands/preview.ts
@@ -16,7 +16,9 @@ import {
  *
  * @param args
  */
-export async function previewCommand(args: WithWranglerArgs<{ cacheChunkSize: number }>): Promise<void> {
+export async function previewCommand(
+	args: WithWranglerArgs<{ cacheChunkSize: number; remote: boolean }>
+): Promise<void> {
 	printHeaders("preview");
 
 	const { config } = await retrieveCompiledConfig();
@@ -25,10 +27,11 @@ export async function previewCommand(args: WithWranglerArgs<{ cacheChunkSize: nu
 	const wranglerConfig = readWranglerConfig(args);
 
 	await populateCache(options, config, wranglerConfig, {
-		target: "local",
+		target: args.remote ? "remote" : "local",
 		environment: args.env,
 		wranglerConfigPath: args.wranglerConfigPath,
 		cacheChunkSize: args.cacheChunkSize,
+		shouldUsePreviewNamespace: args.remote,
 	});
 
 	runWrangler(options, ["dev", ...args.wranglerArgs], { logging: "all" });
@@ -43,7 +46,12 @@ export function addPreviewCommand<T extends yargs.Argv>(y: T) {
 	return y.command(
 		"preview",
 		"Preview a built OpenNext app with a Wrangler dev server",
-		(c) => withPopulateCacheOptions(c),
+		(c) =>
+			withPopulateCacheOptions(c).option("remote", {
+				type: "boolean",
+				default: false,
+				desc: "Run on the global Cloudflare network with access to production resources",
+			}),
 		(args) => previewCommand(withWranglerPassthroughArgs(args))
 	);
 }
--- a/.changeset/odd-colts-walk.md
+++ b/.changeset/odd-colts-walk.md
@@ -1,0 +1,10 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+fix: remote flag not working for preview command's cache population
+
+Previously, passing the `--remote` flag when running `opennextjs-cloudflare preview --remote` would not result in the remote preview binding being populated, and would throw errors due to a missing preview flag when populating Workers KV. The remote flag is now supported for the cache popoulation step when running the preview command.
+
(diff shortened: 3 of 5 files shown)


vicb:
Thanks for the fix @james-elicx 

@dario-piotrowicz could you please take a look as you worked on the remote bindings.

dario-piotrowicz (approved):
Looks good to me 🙂 

Although, tangential to the changes here, long term I would say that ideally [remote bindings](https://developers.cloudflare.com/workers/development-testing/#remote-bindings) should be used instead of running the worker with `--remote` (unless there's a benefit/reason to do so?)

vicb:
> Although, tangential to the changes here, long term I would say that ideally [remote bindings](https://developers.cloudflare.com/workers/development-testing/#remote-bindings) should be used instead of running the worker with --remote (unless there's a benefit/reason to do so?)

I think the plan is to remove `--remote` is the next major of wrangler @dario-piotrowicz right?

packages/cloudflare/src/cli/commands/preview.ts
```diff
@@ -25,10 +27,11 @@ export async function previewCommand(args: WithWranglerArgs<{ cacheChunkSize: nu
 	const wranglerConfig = readWranglerConfig(args);
 
 	await populateCache(options, config, wranglerConfig, {
-		target: "local",
+		target: args.remote ? "remote" : "local",
 		environment: args.env,
 		wranglerConfigPath: args.wranglerConfigPath,
 		cacheChunkSize: args.cacheChunkSize,
+		shouldUsePreviewNamespace: args.remote,
```
vicb: Should `shouldUsePreviewNamespace` just be `remote` / `useRemote`

packages/cloudflare/src/cli/commands/preview.ts
```diff
@@ -43,7 +46,12 @@ export function addPreviewCommand<T extends yargs.Argv>(y: T) {
 	return y.command(
 		"preview",
 		"Preview a built OpenNext app with a Wrangler dev server",
-		(c) => withPopulateCacheOptions(c),
+		(c) =>
+			withPopulateCacheOptions(c).option("remote", {
+				type: "boolean",
+				default: false,
```
vicb, line 53:
```
  -r, --remote                                     Run on the global Cloudflare network with access to production resources  [boolean] [default: false]
```

Add the "r" alias?

packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -228,6 +238,7 @@ function populateD1TagCache(
 			"d1 execute",
 			D1_TAG_BINDING_NAME,
 			`--command "CREATE TABLE IF NOT EXISTS revalidations (tag TEXT NOT NULL, revalidatedAt INTEGER NOT NULL, UNIQUE(tag) ON CONFLICT REPLACE);"`,
+			...(populateCacheOptions.shouldUsePreviewNamespace ? ["--preview"] : ["--preview false"]),
```
vicb: nit:

```suggestion
			...(populateCacheOptions.shouldUsePreviewNamespace ? ["--preview"] : []),
```

Because:
- it's a boolean arg
- to be consistent across the PR

vicb: Same comment applies to l 207

vicb (changes requested):
@james-elicx I added a few minor comments.

The one most important thing is to add the "r" alias for "remote" IMO.

Thanks!

packages/cloudflare/src/cli/commands/preview.ts
james-elicx: I don't think so. Both the preview and deploy commands can populate remote, but only the preview command wants to use preview namespaces when deploying remote.

packages/cloudflare/src/cli/commands/populate-cache.ts
james-elicx: This is intentional - Wrangler doesn't default to false for the flag with KV and specifically asks for this in the error message we were getting. It does default for D1, but I kept it the same as the KV flag for consistency.

packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -109,6 +109,7 @@ type PopulateCacheOptions = {
 	environment?: string;
 	wranglerConfigPath?: string;
 	cacheChunkSize?: number;
+	shouldUsePreviewNamespace?: boolean;
```
vicb: Nit: "Id" matches `preview_id` (kv) / `preview_database_id` (d1) while namespace does not apply to KV.

```suggestion
	shouldUsePreviewId?: boolean;
```

Could you please add JSDoc to this field explaining what it is about and when it is used (`wrangler preview` only).

vicb: We should also add default values.

(Bonus points to add docs for other fields too)

packages/cloudflare/src/cli/commands/populate-cache.ts
vicb: Maybe `--preview ${Boolean(populateCacheOptions.shouldUsePreviewNamespace ?? false)}`

vicb:
Sorry for the delay in the review, please see my comments.
The PR looks ready to go soon.
Thanks for your work on this James!

packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -105,10 +105,30 @@ export function getCacheAssets(opts: BuildOptions): CacheAsset[] {
...
 	environment?: string;
+	/**
+	 * Path to the Wrangler config file.
+	 */
 	wranglerConfigPath?: string;
-	cacheChunkSize?: number;
+	/**
+	 * Chunk sizes to use when populating KV cache. Ignored for R2.
+	 *
+	 * @default 25
```
vicb: I'm not sure to understand why has the `?` been dropped here?

There should be no default is the type `number` (and same for boolean below)?

james-elicx: It's defaulted in yargs - was never actually optional

vicb: I guess it would make sense not to default in yargs as we might want a different default for R2 (we this gets implemented).

That would have my pref, what do you think?

If you decide to keep the default there then it makes sense to update l199 below

```ts
const chunkSize = Math.max(1, populateCacheOptions.cacheChunkSize ?? 25);
```

james-elicx: That makes sense, i'll remove it from yargs

PR #891, opened 2025-09-18
fix error message when setting a key

vicb:


james-elicx (approved):

vicb:
Thanks for the reviews!

PR #896, opened 2025-09-21
Add support for passing multiple configs to wrangler

timReynolds:
This PR updates the opennextjs-cloudflare CLI to correctly pass through multiple configuration files to Wrangler, allowing multiple workers to be bound to a single dev session. This is helpful when using Durable Objects, which aren't bound to the worker when using Miniflare locally, as per the CLI warning.

CleanShot 2025-09-21 at 13 22 23@2x

This reduces the need to run two dev sessions locally and ensures you can use RPC with DO locally, as per the [supported binding docs](https://developers.cloudflare.com/workers/wrangler/api/#supported-bindings).

Developers can enable this by updating preview to:
```
pnpm run build && opennextjs-cloudflare preview -c wrangler.jsonc -c src/agents/wrangler.jsonc
``` 

You can do this today; however, you'll encounter an error when an array is passed through to unstable_readConfig unexpectedly, as yargs will return an array of strings for config without error, misaligned with its typing.

As the CLI has to read the config file used for OpenNext to support some features, I've made the choice to assume the first config passed is the one used to define the worker for Next.js. This might not be the best choice but felt more aligned with Wrangler than adding additional config flags.

Given this usage is probably an edge case and most users don't pass config, it seemed like a reasonable trade-off.

One item that could use more work but would require more code changes is ensuring the info message I added is below the header. I'm happy to make this change if people would consider merging this.

CleanShot 2025-09-21 at 14 03 24@2x

### DIFF
--- a/packages/cloudflare/src/cli/commands/utils.ts
+++ b/packages/cloudflare/src/cli/commands/utils.ts
@@ -15,7 +15,9 @@ import { createOpenNextConfigIfNotExistent, ensureCloudflareConfig } from "../bu
 export type WithWranglerArgs<T = unknown> = T & {
 	// Array of arguments that can be given to wrangler commands, including the `--config` and `--env` args.
 	wranglerArgs: string[];
-	wranglerConfigPath: string | undefined;
+	wranglerConfigPath: string[] | undefined;
+	// The first wrangler config path passed into the CLI, if any. Assumed to be the one used for OpenNext.
+	nextjsWranglerConfigPath: string | undefined;
 	env: string | undefined;
 };
 
@@ -101,7 +103,7 @@ export function getNormalizedOptions(config: OpenNextConfig, buildDir = nextAppD
  * @returns Wrangler config.
  */
 export function readWranglerConfig(args: WithWranglerArgs) {
-	return unstable_readConfig({ env: args.env, config: args.wranglerConfigPath });
+	return unstable_readConfig({ env: args.env, config: args.nextjsWranglerConfigPath });
 }
 
 /**
@@ -111,6 +113,7 @@ export function withWranglerOptions<T extends yargs.Argv>(args: T) {
 	return args
 		.option("config", {
 			type: "string",
+			array: true,
 			alias: "c",
 			desc: "Path to Wrangler configuration file",
 		})
@@ -128,7 +131,7 @@ export function withWranglerOptions<T extends yargs.Argv>(args: T) {
 
 type WranglerInputArgs = {
 	configPath: string | undefined;
-	config: string | undefined;
+	config: string[] | undefined;
 	env: string | undefined;
 };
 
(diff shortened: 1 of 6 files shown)


packages/cloudflare/src/cli/commands/utils.ts
```diff
@@ -166,9 +169,15 @@ function getWranglerArgs(args: WranglerInputArgs & { _: (string | number)[] }):
...
+	const wranglerConfigPath = args.config ?? (args.configPath ? [args.configPath] : undefined);
+	if (wranglerConfigPath && wranglerConfigPath.length > 1) {
+		logger.info("Multiple Wrangler config paths found, first config assumed as opennext config.");
+	}
+
 	return {
 		...args,
-		wranglerConfigPath: args.config ?? args.configPath,
+		wranglerConfigPath,
+		nextjsWranglerConfigPath: wranglerConfigPath?.[0],
```
vicb, line 180: If I understand correctly this will only work when the next app config is first which is not ideal.

Can't you run `opennextjs-cloudflare ...` (for the next app) and `wrangler dev ...` for other workers?

timReynolds: I agree this isn't ideal but I couldn't think of another solution. 

You can run them in separate dev sessions but you'll hit them limitation outline in the [documentation](https://developers.cloudflare.com/workers/wrangler/api/#supported-bindings); 

> If you are not using RPC with your Durable Object, you can run a separate Wrangler dev session alongside your framework development server.
Otherwise, you can build your application and run both Workers in the same Wrangler dev session.

I found this when trying to port the [agents starter](https://github.com/cloudflare/agents-starter) to opennext-cloudflare which is working when I link this change. 

I'm trying to run them locally in separate dev sessions now but its having an issue upgrading to web-sockets I've not debugged yet.

I'm open to other ways of solving this or I can just open an issue so it can be fixed upstream in wrangler somehow

vicb:
Sounds like inter process communication is coming soon.
I'll update the issue when I have more info

packages/cloudflare/src/cli/commands/utils.ts
mattvick: Seeing this now after creating a [similar PR](#997) (I should have checked first, sorry). When testing my PR all workers ran irrespective of the order that the config args were passed.

mattvick: @vicb do you have an update regarding the inter process communication?

vicb:
@timReynolds @mattvick I did say support was [coming soon](https://github.com/opennextjs/opennextjs-cloudflare/pull/896#issuecomment-3323846213) but the it looks like the implementation has changed and is still in flux. Could one of you please create an issue on https://github.com/cloudflare/workers-sdk describing your use case and why you need that?
Thanks!

PR #898, opened 2025-09-23
perf: add an index on the tag column in D1 tag cache

vicb:
Adding an index to speed up reads which are dominant on the tag cache.

(D1 now uses the same SQL structure/queries than what is used on with DO)


vicb:
oops, nevermind UNIQUE is already here 😅

Issue #899, opened 2025-09-23
[BUG] OpenNext AWS adapter code is included in Cloudflare bundle, breaking request serving at runtime


WilcoKruijer:
### Describe the bug

The [server-adapter.ts](https://github.com/opennextjs/opennextjs-aws/blob/main/packages/open-next/src/adapters/server-adapter.ts#L27) file from the OpenNext AWS adapter is included in a Cloudflare OpenNext build. This file contains a side-effect of setting `process.chdir(__dirname);`. 

The problem can be found in the bundle (after following reproduction steps): `.open-next/server-functions/default/index.mjs` (Search for: setNextjsServerWorkingDirectory). In the `.wrangler` directory the same function can be found, where we can observe that this becomes a call to `process.chdir("");`

This issue was hidden because Cloudflare's `nodejs_compat` did not correctly implement `chdir`, at least until `enable_nodejs_process_v2` was enabled by default on compatibility date [2025-09-15](https://developers.cloudflare.com/workers/configuration/compatibility-flags/#enable-process-v2-implementation).

Updating the compatibility date caused my site to break **at runtime**. `npm run dev` kept working, as well as building the project.

**Workaround:** set compatibility date to `2025-09-01` or earlier.

### Steps to reproduce

1. Create a new open-next Cloudflare project: `npm create cloudflare@latest -- my-next-app --framework=next`
2. Change the Wrangler compatibility date to `2025-09-15` (or later) OR add flag `enable_nodejs_process_v2`
3. `npm run preview`
4. Open the URL in your browser and observe the request erroring

### Expected behavior

No error :^)

### @opennextjs/cloudflare version

1.8.4

### Wrangler version

4.38

### next info output

```shell
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.5.0: Tue Apr 22 19:53:27 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T6041
  Available memory (MB): 49152
  Available CPU cores: 12
Binaries:
  Node: 22.13.1
  npm: 10.9.2
  Yarn: 1.22.22
  pnpm: 10.13.1
Relevant Packages:
  next: 15.4.6 // There is a newer version (15.5.3) available, upgrade recommended! 
  eslint-config-next: N/A
  react: 19.1.0
  react-dom: 19.1.0
  typescript: 5.9.2
Next.js Config:
  output: N/A
[...]

PR #903, opened 2025-09-23
fix: enable using workerd process v2

vicb:
fixes https://github.com/opennextjs/opennextjs-cloudflare/issues/899

`node:process` has a new implementation in workerd ("v2" enabled by default on 2025-09-15).
v2 supports `process.chdir()` so it now errors when a non existent directory is passed (the cloudflare adapter passes "").
(The `unenv` implementation used before was a no-op).

### DIFF
--- a/.changeset/stale-adults-divide.md
+++ b/.changeset/stale-adults-divide.md
@@ -1,0 +1,7 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+fix: enable using workerd process v2
+
+process v2 is an updated version of `node:process` active by default after 2025-09-15
--- a/examples/playground15/wrangler.jsonc
+++ b/examples/playground15/wrangler.jsonc
@@ -2,7 +2,7 @@
 	"$schema": "node_modules/wrangler/config-schema.json",
 	"main": ".open-next/worker.js",
 	"name": "playground15",
-	"compatibility_date": "2024-12-30",
+	"compatibility_date": "2025-09-15",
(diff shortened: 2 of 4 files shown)


vicb:
> It does work, but I don't think this is the right fix. We should not patch things in aws here, at least not like that. We should use this plugin https://github.com/opennextjs/opennextjs-aws/blob/main/packages/open-next/src/plugins/replacement.ts and just add an `#override` around it in aws

I agree it's not ideal and we should improve this.
One way would be to pass `additionalCodePatches` to `createServerBundle` and use the replacement plugin.
But:

- cloudflare already has its own version of `createServerBundle`
- having AWS specific code in the "common" code base is not the best thing either

So let's merge this PR as is for now and think about a better way to support "common" + variants when we work on the adapters API.

WDYT?

vicb:
> > So let's merge this PR as is for now and think about a better way to support "common" + variants when we work on the adapters API.
> > WDYT?
> 
> That's fine by me, but I still prefer the `#override` solution with the `additionalCodePatches`
> 
> What i really don't like about this one is that there is no way to know from the aws repo that we are modifying this function somewhere else. Very easy to forget about it and break a release this way

Agreed.

Hopefully we don't add more than what the `setNextjsServerWorkingDirectory` name says in this function. Also this is e2e tested in playground15.

I have create [this issue](https://github.com/opennextjs/opennextjs-cloudflare/issues/904) to improve the current PR but let's merge this now to avoid new projects fail build for now.

vicb:
Thanks all for the review and feedback.

As the comments say, we should improve the current implementation but I'll merge this to avoid breaking new projects.

PR #906, opened 2025-09-24
feat: add an experimental KV based tag cache

vicb:


packages/cloudflare/src/api/overrides/tag-cache/kv-next-tag-cache.ts
```diff
@@ -0,0 +1,95 @@
...
+	readonly mode = "nextMode" as const;
+	readonly name = NAME;
+
+	async getLastRevalidated(tags: string[]): Promise<number> {
+		const kv = this.getKv();
+		if (!kv) {
+			return 0;
+		}
+
+		console.log(`[KVNextModeTagCache] getLastRevalidated, tags=${tags}`);
```
sommeeeer: Debug?

vicb: Good catch, thanks!

examples/overrides/kv-tag-next/package.json
```diff
@@ -0,0 +1,29 @@
+{
+	"name": "d1-tag-next",
```
conico974:
```suggestion
	"name": "kv-tag-next",
```

packages/cloudflare/src/api/overrides/tag-cache/kv-next-tag-cache.ts
```diff
@@ -0,0 +1,94 @@
...
+export const NAME = "kv-next-mode-tag-cache";
+
+export const BINDING_NAME = "NEXT_TAG_CACHE_KV";
+
+/**
+ * Tag Cache based on a KV namespace
+ *
+ * Warning:
+ * This implementation is considered experimental for now.
+ * KV is eventually consistency and can take up to 60s to reflect the last write.
```
conico974:
```suggestion
 * KV is eventually consistent and can take up to 60s to reflect the last write.
```

packages/cloudflare/src/api/overrides/tag-cache/kv-next-tag-cache.ts
```diff
@@ -0,0 +1,94 @@
...
+ * This means that:
+ * - revalidations can take up to 60s to apply
+ * - when a page depends on multiple tags they can be inconsistent for up to 60s
```
conico974: It's actually a tiny bit more problematic than that, 2 pages that uses the same set of 2 tags could be revalidated at different times, leading to what supposed to be the same data with different data. 
This inconsistency at page level could stay forever because the page would have been revalidated after the date in the tag cache (`hasBeenRevalidated` would be false)

packages/cloudflare/src/api/overrides/tag-cache/kv-next-tag-cache.ts
```diff
@@ -0,0 +1,94 @@
...
+		}
+
+		const timeMs = String(Date.now());
+
+		await Promise.all(
+			tags.map(async (tag) => {
+				await kv.put(this.getCacheKey(tag), timeMs);
+			})
+		);
+		await purgeCacheByTags(tags);
```
conico974, line 68: This is already done in aws, we should not do it again here

conico974 (approved):
Just some small little change needed and a few thing maybe to add as comment

packages/cloudflare/src/api/overrides/tag-cache/kv-next-tag-cache.ts
vicb: So I guess it should be removed from D1 and DO too?

vicb: Not sure how/where this is done in aws though?
Can you point to the code?

conico974: Nevermind I was wrong, but we should do it there https://github.com/opennextjs/opennextjs-aws/blob/23ed1df6b887526dc8b573c0ea099b4009b410ec/packages/open-next/src/adapters/cache.ts#L331-L333

vicb: ~Before checking the length, right?~

vicb: Oh I see what you mean, we should implement a new method in the CDN invalidation.

So I guess this code is good to merge now (I'll add a TODO here and for D1 & DO).
(A follow-up I had was to only call `purgeCacheByTags` when there is a CDN invalidator so that we do not log errors when not).

I could create an issue in aws to add the method and reference it in this repo.

Do that sound good?

packages/cloudflare/src/api/overrides/tag-cache/kv-next-tag-cache.ts
vicb: I clarified the comment

packages/cloudflare/src/api/overrides/tag-cache/kv-next-tag-cache.ts
conico974: Yeah perfect

vicb:
Thanks for the review and feedback Nico!

PR #907, opened 2025-09-25
refactor: only try to purge the cache when invalidation is configured

vicb:
Before this PR, `purgeCacheByTags` is called whether or not the invalidation is configured.

This result in logging an error when the invalidation is not configured.

This PR removes the error when the invalidation is not configured (calling `purgeCacheByTags` is a noop anyway).


james-elicx (approved):

vicb:
Thanks James!

PR #911, opened 2025-09-26
add cache debug log

vicb:


packages/cloudflare/src/api/overrides/incremental-cache/static-assets-incremental-cache.ts
```diff
@@ -50,11 +50,11 @@ class StaticAssetsIncrementalCache implements IncrementalCache {
 	}
 
 	async set(): Promise<void> {
-		error("Failed to set to read-only cache");
+		error("StaticAssetsIncrementalCache: Failed to set to read-only cache");
```
conico974: Here I think we should add the key and cache type so that people knows what are triggering this

vicb: We should probably add that one layer up in AWS instead - a sha does not really help.
WDYT?

conico974: The sha is only if it's the data cache, and actually we don't have any more info in aws, the sha is generated by Next.
For pages they should get something readable, and they shouldn't really use the data cache with this anyway

vicb:
Thanks for the review and feedback Nico!

PR #918, opened 2025-10-01
refactor: account for empty tag list in tag cache

vicb:
We were not handling the case were the tag list is empty, because we assume it would not happen...
But https://github.com/opennextjs/opennextjs-aws/pull/988

So this PR adds an explicit check - it should not be needed after the aws is merged but defensive programming is good when the perf impact is insignificant


james-elicx (approved):

vicb:
Thanks for the reviews!

Issue #866, opened 2025-09-02
[BUG] Slow Wrangler file by file cache populating


grabmateusz:
### Describe the bug

Hi,
we're using @opennextjs/cloudflare in version 1.6.5.
We have ISR enabled and we use R2 for incremental cache.
During deployment we deploy ~500 items to the cache and there are 2 issues:
- it takes over 20 minutes,
- sometimes it randomly crashes on 500 / 502 HTTP response code during deploy.
Is there any way to speed up deployment process?

### Steps to reproduce

1. Create a Next.js app with ISR, add multiple HTTP response code to render process of pages that are subject of ISR
2. Deploy such OpenNext worker to Cloudflare using `opennextjs-cloudflare deploy` command

### Expected behavior

1. Deployment time should be reduced
2. Deployment should be deterministic and do not randomly fail in the middle

### @opennextjs/cloudflare version

1.6.5

### Wrangler version

4.33.1

### next info output

```shell
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.5.0: Tue Apr 22 19:54:49 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T6000
  Available memory (MB): 32768
  Available CPU cores: 10
Binaries:
  Node: 20.19.2
  npm: 10.8.2
  Yarn: 1.22.21
  pnpm: 10.15.0
Relevant Packages:
  next: 15.3.4 // There is a newer version (15.5.2) available, upgrade recommended! 
  eslint-config-next: 14.1.0
  react: 18.3.1
  react-dom: 18.3.1
  typescript: 5.3.3
Next.js Config:
  output: N/A
 ⚠ There is a newer version (15.5.2) available, upgrade recommended! 
   Please try the latest canary version (`npm install next@canary`) to confirm the issue still exists before creating a new issue.
   Read more - https://nextjs.org/docs/messages/opening-an-issue
```

makepanic:
I'm also seeing an issue when attempting to upload a opennext app with many static pages.
It's using ISR with `staticAssetsIncrementalCache`.

```
11:46:08.636 | 🌀 Building list of assets...
-- | --
11:46:08.941 | ✨ Read 16729 files from the assets directory /opt/buildhome/repo/.open-next/assets
11:46:18.054 | 🌀 Starting asset upload...
11:46:24.957 | 🌀 Found 9153 new or modified static assets to upload. Proceeding with upload...
```

[...]

PR #925, opened 2025-10-03
feat(cloudflare): add --rcloneBatch flag for faster R2 cache uploads using rclone

krzysztof-palka-monogo:
# Batch upload R2 cahce using rclone

This PR proposes adding support for optional batch uploading R2 cache using rclone.
Based on the solution discussed in [Issue #866](https://github.com/opennextjs/opennextjs-cloudflare/issues/866)

## Details

The `--rcloneBatch` flag enables faster R2 cache uploads using rclone batch mode. 
This flag is supported by the following commands:
- `populateCache` - Explicitly populate cache
- `deploy` - Deploy and populate cache
- `upload` - Upload version and populate cache
- `preview` - Preview and populate cache

### DIFF
--- a/.changeset/rclone-batch-upload.md
+++ b/.changeset/rclone-batch-upload.md
@@ -1,0 +1,34 @@
+---
+"@opennextjs/cloudflare": minor
+---
+
+feature: add --rcloneBatch flag for faster R2 cache uploads
+
+This adds a new optional `--rcloneBatch` flag that enables batch uploading to R2 using rclone instead of individual wrangler uploads. This significantly improves upload performance for large caches.
+
+**Supported commands:**
+
+- `populateCache` - Explicit cache population
+- `deploy` - Deploy with cache population
+- `upload` - Upload version with cache population
+- `preview` - Preview with cache population
+
+**Performance improvements:**
+
+- Creates staging directory with all cache files organized by R2 keys
+- Uses rclone's parallel transfer capabilities (32 transfers, 16 checkers)
+- Reduces API calls to Cloudflare
+
+**Usage:**
+
+```bash
+opennextjs-cloudflare deploy --rcloneBatch
+opennextjs-cloudflare populateCache remote --rcloneBatch
+```
+
+**Requirements:**
+
+- The `rclone.js` package (included as dependency) provides the binary
+- An rclone config file at `~/.config/rclone/rclone.conf` with R2 credentials (see README for setup instructions)
+
+The original wrangler-based upload remains the default behavior for backward compatibility.
--- a/packages/cloudflare/README.md
+++ b/packages/cloudflare/README.md
@@ -55,3 +55,54 @@ Deploy your application to production with the following:
   # or
   bun opennextjs-cloudflare build && bun opennextjs-cloudflare deploy
   ```
+
+## CLI Options
+
+### Batch Cache Population (rclone)
+
+The `--rcloneBatch` flag enables faster R2 cache uploads using rclone batch mode. This flag is supported by the following commands:
+
+- `populateCache` - Explicitly populate cache
+- `deploy` - Deploy and populate cache
+- `upload` - Upload version and populate cache
+- `preview` - Preview and populate cache
+
+**Usage:**
+
+```bash
+# Standalone cache population
+npx opennextjs-cloudflare populateCache local --rcloneBatch
+npx opennextjs-cloudflare populateCache remote --rcloneBatch
+
+# During deployment
+npx opennextjs-cloudflare deploy --rcloneBatch
+
+# During upload
+npx opennextjs-cloudflare upload --rcloneBatch
+
+# During preview
+npx opennextjs-cloudflare preview --rcloneBatch
+```
+
+**Requirements:**
+
+1. The `rclone.js` package (included as a dependency) provides the rclone binary automatically
+2. An rclone configuration file is required at `~/.config/rclone/rclone.conf` with your R2 credentials
+
+**rclone Configuration:**
+
+Create or update `~/.config/rclone/rclone.conf` with your R2 bucket configuration:
+
+```ini
+[r2]
+type = s3
+provider = Cloudflare
+access_key_id = YOUR_ACCESS_KEY_ID
+secret_access_key = YOUR_SECRET_ACCESS_KEY
+endpoint = https://YOUR_ACCOUNT_ID.r2.cloudflarestorage.com
+acl = private
+```
+
+See [Cloudflare's rclone documentation](https://developers.cloudflare.com/r2/examples/rclone/) for more details.
+
+**Default:** `false` (uses standard wrangler-based uploads)
--- a/packages/cloudflare/src/cli/commands/populate-cache.ts
+++ b/packages/cloudflare/src/cli/commands/populate-cache.ts
@@ -1,4 +1,5 @@
-import { cpSync, existsSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { spawnSync } from "node:child_process";
+import { copyFileSync, cpSync, existsSync, mkdirSync, readFileSync, rmSync, writeFileSync } from "node:fs";
 import path from "node:path";
 
 import type { BuildOptions } from "@opennextjs/aws/build/helper.js";
@@ -127,7 +128,53 @@ type PopulateCacheOptions = {
 	 * Instructs Wrangler to use the preview namespace or ID defined in the Wrangler config for the remote target.
 	 */
 	shouldUsePreviewId: boolean;
+	/**
+	 * Use rclone for batch uploading to R2 instead of individual wrangler uploads.
+	 *
+	 * @default false
+	 */
+	rcloneBatch?: boolean;
 };
+
+async function populateR2IncrementalCacheWithRclone(
+	options: BuildOptions,
+	bucket: string,
+	prefix: string | undefined,
+	assets: CacheAsset[]
+) {
+	logger.info("\nPopulating R2 incremental cache using rclone batch upload...");
+
+	// Create a staging dir with proper key paths
+	const stagingDir = path.join(options.outputDir, ".r2-staging");
+	rmSync(stagingDir, { recursive: true, force: true });
+	mkdirSync(stagingDir, { recursive: true });
+
+	for (const { fullPath, key, buildId, isFetch } of assets) {
+		const cacheKey = computeCacheKey(key, {
+			prefix,
+			buildId,
+			cacheType: isFetch ? "fetch" : "cache",
+		});
+		const destPath = path.join(stagingDir, cacheKey);
+		mkdirSync(path.dirname(destPath), { recursive: true });
+		copyFileSync(fullPath, destPath);
+	}
+
+	// Use rclone to sync the R2
+	const remote = `r2:${bucket}`;
+	const args = ["copy", stagingDir, remote, "--progress", "--transfers=32", "--checkers=16"];
+
+	const result = spawnSync("rclone", args, {
+		stdio: "inherit",
+		env: process.env,
+	});
+
+	if (result.status !== 0) {
+		throw new Error("rclone copy failed");
+	}
+
+	logger.info(`Successfully uploaded ${assets.length} assets to R2`);
+}
 
 async function populateR2IncrementalCache(
 	options: BuildOptions,
@@ -150,6 +197,10 @@ async function populateR2IncrementalCache(
 	const prefix = envVars[R2_CACHE_PREFIX_ENV_NAME];
 
 	const assets = getCacheAssets(options);
+
+	if (populateCacheOptions.rcloneBatch) {
+		return populateR2IncrementalCacheWithRclone(options, bucket, prefix, assets);
+	}
 
 	for (const { fullPath, key, buildId, isFetch } of tqdm(assets)) {
 		const cacheKey = computeCacheKey(key, {
@@ -329,7 +380,7 @@ export async function populateCache(
  */
 async function populateCacheCommand(
 	target: "local" | "remote",
-	args: WithWranglerArgs<{ cacheChunkSize?: number }>
+	args: WithWranglerArgs<{ cacheChunkSize?: number; rcloneBatch?: boolean }>
 ) {
 	printHeaders(`populate cache - ${target}`);
 
@@ -344,6 +395,7 @@ async function populateCacheCommand(
 		wranglerConfigPath: args.wranglerConfigPath,
 		cacheChunkSize: args.cacheChunkSize,
 		shouldUsePreviewId: false,
+		rcloneBatch: args.rcloneBatch,
 	});
 }
 
@@ -372,8 +424,14 @@ export function addPopulateCacheCommand<T extends yargs.Argv>(y: T) {
 }
 
 export function withPopulateCacheOptions<T extends yargs.Argv>(args: T) {
-	return withWranglerOptions(args).options("cacheChunkSize", {
-		type: "number",
-		desc: "Number of entries per chunk when populating the cache",
-	});
+	return withWranglerOptions(args)
+		.options("cacheChunkSize", {
+			type: "number",
+			desc: "Number of entries per chunk when populating the cache",
+		})
+		.options("rcloneBatch", {
+			type: "boolean",
+			desc: "Use rclone for batch uploading to R2 (faster for large caches)",
+			default: false,
+		});
 }
--- a/packages/cloudflare/package.json
+++ b/packages/cloudflare/package.json
@@ -57,6 +57,7 @@
 		"cloudflare": "^4.4.1",
 		"enquirer": "^2.4.1",
 		"glob": "catalog:",
+		"rclone.js": "^0.6.6",
 		"ts-tqdm": "^0.8.6",
 		"yargs": "catalog:"
 	},
--- a/packages/cloudflare/src/cli/commands/deploy.ts
+++ b/packages/cloudflare/src/cli/commands/deploy.ts
@@ -19,7 +19,9 @@ import {
  *
  * @param args
  */
-export async function deployCommand(args: WithWranglerArgs<{ cacheChunkSize?: number }>): Promise<void> {
+export async function deployCommand(
+	args: WithWranglerArgs<{ cacheChunkSize?: number; rcloneBatch?: boolean }>
+): Promise<void> {
 	printHeaders("deploy");
 
 	const { config } = await retrieveCompiledConfig();
@@ -40,6 +42,7 @@ export async function deployCommand(args: WithWranglerArgs<{ cacheChunkSize?: nu
 		wranglerConfigPath: args.wranglerConfigPath,
 		cacheChunkSize: args.cacheChunkSize,
 		shouldUsePreviewId: false,
+		rcloneBatch: args.rcloneBatch,
 	});
 
 	runWrangler(
--- a/packages/cloudflare/src/cli/commands/populate-cache.spec.ts
+++ b/packages/cloudflare/src/cli/commands/populate-cache.spec.ts
@@ -5,7 +5,7 @@ import type { BuildOptions } from "@opennextjs/aws/build/helper.js";
 import mockFs from "mock-fs";
 import { afterAll, beforeAll, describe, expect, test } from "vitest";
 
-import { getCacheAssets } from "./populate-cache.js";
+import { getCacheAssets, withPopulateCacheOptions } from "./populate-cache.js";
 
 describe("getCacheAssets", () => {
 	beforeAll(() => {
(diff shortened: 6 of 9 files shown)


vicb:
@krzysztof-palka-monogo thank you so much for this PR, it will please many of our users!

I only took a quick look and I'll do a full review on Monday.

I have some preliminary comments/thoughts:
- I think we should rename `--rcloneBatch`. First we do not want to expose the implementation (rclone) as it might change - we can keep mention to rclone and in the doc.  Ideally we can re-use `cacheChunkSize`
- `rclone.js` was last released 3y ago - Ideally we could find a more recent and maintained implementation (but we should not block the PR on that, it could be a follow-up or we can wait until we actually have an issue)
- We should info log when users are not using batch upload and have i.e. > 10 assets. Another option could be to batch upload by default and fallback to slow/safe on error (again happy if this addressed as a follow up)

ghost:
Good morning @vicb

Here’s the justification for keeping an explicit `--rcloneBatch` (or similarly named) flag and retaining the existing upload mechanism as the default:

1) Requires external configuration
Rclone is not a drop-in replacement — it depends on a separate config file (the rclone config, credentials, remotes, etc). For users who have never used rclone, they must explicitly create and supply that config. That means it is not “just work” out of the box.

2) Avoids a breaking change
If we made rclone-based behavior the default, we risk breaking setups for users who do not have configured rclone. Keeping the current behavior as default ensures backward compatibility.

3) Explicit intent
Having a named flag (e.g. --rcloneBatch) clearly signals the use of rclone and keeps any implementation coupling explicit (rather than hidden). It also gives future flexibility: if one day we swap in a different batch uploader (or abstraction), the explicit flag boundary helps isolate that change.

vicb:
> Here’s the justification for keeping an explicit `--rcloneBatch` (or similarly named) flag and retaining the existing upload mechanism as the default:

We definitely don't want "rclone" in the flag. We could use another mechanism, i.e. remote bindings.
 
> 1. Requires external configuration
>    Rclone is not a drop-in replacement — it depends on a separate config file (the rclone config, credentials, remotes, etc). For users who have never used rclone, they must explicitly create and supply that config. That means it is not “just work” out of the box.

IMO we should drop the need for an external config in favor of using env vars - the CLI could dynamically create a temp config file for rclone.js as it doesn't seem to support env vars.

> 2. Avoids a breaking change
>    If we made rclone-based behavior the default, we risk breaking setups for users who do not have configured rclone. Keeping the current behavior as default ensures backward compatibility.

We can use rclone when configured and fallback to the current mechanism - but log that there is faster way and link some docs.

> 3. Explicit intent
>    Having a named flag (e.g. --rcloneBatch) clearly signals the use of rclone and keeps any implementation coupling explicit (rather than hidden). It also gives future flexibility: if one day we swap in a different batch uploader (or abstraction), the explicit flag boundary helps isolate that change.

I disagree here.
Users want something that works (fast), they should not have to care about how it is implemented.

ghost:
Hi @vicb 

I've updated the PR to align with your suggestions:

- Removed `--rcloneBatch` flag and all rclone references from user-facing API
- Implemented environment variable configuration (R2_ACCESS_KEY_ID, R2_SECRET_ACCESS_KEY, R2_ACCOUNT_ID)
- Batch upload now activates automatically when credentials are detected
- Fallback to wrangler uploads when credentials aren't provided
- Temp config files are created dynamically with secure permissions and
auto-cleanup

Documentation now focuses on "batch upload" as an optional performance feature, not implementation details.

Please let me know if thats will meet the requirements and if so could we run "Publish prereleases" action to properly test the changes in some real applications 😄

ghost:
Hi @vicb 

I will be glad for code review.

.changeset/rclone-batch-upload.md
```diff
@@ -0,0 +1,42 @@
...
+- Reduced API calls to Cloudflare
+
+**Usage:**
+
+```bash
+# Enable batch upload by setting environment variables (recommended for large caches)
+export R2_ACCESS_KEY_ID=your_key
+export R2_SECRET_ACCESS_KEY=your_secret
+export R2_ACCOUNT_ID=your_account
+opennextjs-cloudflare deploy  # batch upload automatically used
```
vicb: were there a link to the docs here to generate the values before, might be nice to (re)add

packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -129,6 +131,127 @@ type PopulateCacheOptions = {
...
+}
+
+/**
+ * Create a temporary configuration file for batch upload from environment variables
+ * @returns Path to the temporary config file or null if env vars not available
+ */
+function createTempRcloneConfig(): string | null {
+	const accessKey = process.env.R2_ACCESS_KEY_ID;
+	const secretKey = process.env.R2_SECRET_ACCESS_KEY;
+	const accountId = process.env.R2_ACCOUNT_ID;
```
vicb: To avoid double access, maybe there could be a `retrieveR2CredentialsFromEnv` returning `{R2_ACCESS_KEY_ID, R2_SECRET_ACCESS_KEY, CLOUDFLARE_ACCOUNT_ID}` (I don't think the account as anything to do with R2)

ghost:
Hello @vicb 

I've addressed both points:
1) Renamed `R2_ACCOUNT_ID` to `CLOUDFLARE_ACCOUNT_ID`
2) The CLI now loads environment variables from a `.env` file in the project root.

This keeps the implementation simple while making local dev more convenient. WDYT?

packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -129,6 +131,127 @@ type PopulateCacheOptions = {
...
+		}
+
+		logger.info(`Successfully uploaded ${assets.length} assets to R2 using batch upload`);
+	} finally {
+		try {
+			// Cleanup temporary config file
+			rmSync(tempConfigPath);
+		} catch {
+			console.warn(`Failed to remove temporary config at ${tempConfigPath}`);
+		}
```
vicb, line 323: maybe use force and recursive instead of try/catch?

vicb:
> This keeps the implementation simple while making local dev more convenient. WDYT?

I don't think there is an ideal solution for `.env` vs a local config file.

That being said, Next.js uses .env files with [this loading order](https://nextjs.org/docs/app/guides/environment-variables#environment-variable-load-order) so we should make sure we are consistent.

I'll take a deeper look today and comment on how we could do that

ghost:
I’ll be away for a week starting tomorrow, so I need to wrap up this functionality today 😅
Since our ideas have been evolving, could we prioritize only essential changes now and handle the rest in a follow-up iteration?

.changeset/rclone-batch-upload.md
```diff
@@ -0,0 +1,55 @@
...
+
+This update adds optional batch upload support for R2 cache population, significantly improving upload performance for large caches when enabled via environment variables.
+
+**Key Changes:**
+
+1. **Optional Batch Upload**: Configure R2 credentials via environment variables to enable faster batch uploads:
+
+   - `R2_ACCESS_KEY_ID`
+   - `R2_SECRET_ACCESS_KEY`
+   - `CLOUDFLARE_ACCOUNT_ID`
```
vicb: Sorry for the back and forth here but it turns out that we already use `CF_ACCOUNT_ID` for skew protection. Could you please use the same name here?

Thanks

packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -151,6 +274,23 @@ async function populateR2IncrementalCache(
 
 	const assets = getCacheAssets(options);
 
+	// Use batch upload if R2 credentials are available (optional but recommended)
+	const canUseBatchUpload = hasR2EnvCredentials();
+
+	if (canUseBatchUpload) {
```
vicb: I think we can remove the call to `hasR2EnvCredentials` and the if and handle everything in `populateR2IncrementalCacheWithBatchUpload`

My current thinking is to update `getEnvFromPlatformProxy` to retrieve env from `process.env`, `.env*`, `.dev.vars`, and the wrangler wrangler config (currently the last 2 only are handled).

ghost: > I think we can remove the call to hasR2EnvCredentials and the if and handle everything in populateR2IncrementalCacheWithBatchUpload

I’m not sure I follow — the `if()` here is what decides whether the batch upload function should be called or not.
Could you clarify what you mean by handling everything directly in `populateR2IncrementalCacheWithBatchUpload`?

> My current thinking is to update getEnvFromPlatformProxy to retrieve env from process.env, .env*, .dev.vars, and the wrangler wrangler config (currently the last 2 only are handled).

I’m not entirely sure what `getEnvFromPlatformProxy()` currently does.
Could you maybe make the change you have in mind?

vicb: > I’m not sure I follow — the if() here is what decides whether the batch upload function should be called or not.
Could you clarify what you mean by handling everything directly in populateR2IncrementalCacheWithBatchUpload?

```ts
  try {
    batch(); // error if not properly configured
  } catch {
    sequential();
  }
```      

> I’m not entirely sure what getEnvFromPlatformProxy) currently does. Could you maybe make the change you have in mind?

We're having some discussions on the team and I'll update when we setlle on something.

ghost: Just pushed an update following your idea:
```typescript
try {
  batch(); // error if not properly configured
} catch {
  sequential();
}
```

That’s my final commit before a week away — fingers crossed it’s good to go 😅

packages/cloudflare/src/cli/commands/populate-cache.ts
ghost: This is path to the config file, not directory :)

packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -129,27 +131,142 @@ type PopulateCacheOptions = {
...
+			"copy",
+			stagingDir,
+			remote,
+			"--progress",
+			"--transfers=32",
+			"--checkers=16",
+			"--error-on-no-transfer", // Fail if no files transferred
+		];
+
+		const result = spawnSync("rclone", args, {
```
vicb: any reason not to use rclone.js API as it is installed anyway?

packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -129,27 +131,142 @@ type PopulateCacheOptions = {
...
-	const envVars = await getEnvFromPlatformProxy(populateCacheOptions);
-	const prefix = envVars[R2_CACHE_PREFIX_ENV_NAME];
+	const env = { ...process.env };
+	env.RCLONE_CONFIG = tempConfigPath;
+	logger.info("Using batch upload with R2 credentials from environment variables");
 
-	const assets = getCacheAssets(options);
+	try {
+		// Create a staging dir with proper key paths
+		const stagingDir = path.join(options.outputDir, ".r2-staging");
```
vicb: should `.r2-staging` be a temp dir instead or cleanup at the end ?

packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -177,6 +294,46 @@ async function populateR2IncrementalCache(
...
+	}
+
+	const envVars = await getEnvFromPlatformProxy(populateCacheOptions);
+	const prefix = envVars[R2_CACHE_PREFIX_ENV_NAME];
+
+	const assets = getCacheAssets(options);
+
+	try {
+		// Attempt batch upload first (using rclone)
+		return await populateR2IncrementalCacheWithBatchUpload(options, bucket, prefix, assets);
```
vicb: We should not do this for local R2 as the API is remote only

vicb:
https://github.com/opennextjs/opennextjs-cloudflare/pull/937 allow retrieving env vars from process.env in CI

packages/cloudflare/src/cli/commands/populate-cache.ts
ghost: Package is in ver 0.6.6 • Published 3 years ago
But it has nice feature (fetches newest rclone): https://github.com/sntran/rclone.js/blob/main/rclone/selfupdate.js

benchmarking/src/cloudflare.ts
ghost, line 1: Looks like this file was forgotten to be updated when `packages/cloudflare/src/api/cloudflare-context.ts` was introduced?

vicb: That's an env var used by wrangler, please revert

ghost: reverted

packages/cloudflare/src/cli/commands/populate-cache.ts
ghost: When target is local -> forced sequential

ghost:
> [Open in StackBlitz](https://pkg.pr.new/template/0696a09f-aa55-4206-8a8b-5216d4572847)
> 
> ```
> npm i https://pkg.pr.new/@opennextjs/cloudflare@925
> ```
> 
> _commit: [`0d53eac`](https://github.com/opennextjs/opennextjs-cloudflare/runs/52030594609)_

We did tests based on our website and results are amazing:

image

packages/cloudflare/src/cli/commands/populate-cache.ts
vicb: Doesn't it have a working version of copy?

.changeset/rclone-batch-upload.md
```diff
@@ -0,0 +1,55 @@
...
+opennextjs-cloudflare deploy  # batch upload automatically used
+```
+
+Or create a `.env` file in your project root (automatically loaded):
+
+```bash
+R2_ACCESS_KEY_ID=your_key
+R2_SECRET_ACCESS_KEY=your_secret
+CF_ACCOUNT_ID=your_account
+```
```
vicb:
```suggestion
Add the crendentials in a `.env`/`.dev.vars` file in your project root:

\`\`\`bash
R2_ACCESS_KEY_ID=your_key
R2_SECRET_ACCESS_KEY=your_secret
CF_ACCOUNT_ID=your_account
\`\`\`

You can also set the environment variables for CI builds.

```

packages/cloudflare/README.md
```diff
@@ -55,3 +55,36 @@ Deploy your application to production with the following:
   # or
   bun opennextjs-cloudflare build && bun opennextjs-cloudflare deploy
   ```
+
+### Batch Cache Population (Optional, Recommended)
+
+For improved performance with large caches, you can enable batch upload by providing R2 credentials via environment variables.
+
+You can either set environment variables directly:
```
vicb: Same suggestion as in the changeset

packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -81,6 +81,10 @@ declare global {
 		CF_WORKERS_SCRIPTS_API_TOKEN?: string;
 		// Cloudflare account id
 		CF_ACCOUNT_ID?: string;
+
+		// R2 API credentials for batch cache population (optional, enables faster uploads)
+		R2_ACCESS_KEY_ID?: string;
+		R2_SECRET_ACCESS_KEY?: string;
```
vicb:
```suggestion

		// Cloudflare account id - needed for skew protection and R2 batch population
		CF_ACCOUNT_ID?: string;

		// R2 API credentials for batch cache population (optional, enables faster uploads)
		R2_ACCESS_KEY_ID?: string;
		R2_SECRET_ACCESS_KEY?: string;
```

packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -204,27 +206,151 @@ type PopulateCacheOptions = {
...
+ */
+async function populateR2IncrementalCacheWithBatchUpload(
+	bucket: string,
+	prefix: string | undefined,
+	assets: CacheAsset[],
 	envVars: WorkerEnvVar
 ) {
-	logger.info("\nPopulating R2 incremental cache...");
+	// Ensure R2 credentials are available in env vars
+	const { accessKey, secretKey, accountId } = getR2CredentialsFromEnv(envVars);
```
vicb: We can probably destructurate envVars and remove the extra function?

packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -204,27 +206,151 @@ type PopulateCacheOptions = {
...
+
+	// Create temporary config from env vars - required for batch upload
+	const tempConfigPath = createTempRcloneConfig(accessKey, secretKey, accountId);
+	if (!tempConfigPath) {
+		throw new Error("Failed to create temporary rclone config for R2 batch upload.");
 	}
 
-	const prefix = envVars[R2_CACHE_PREFIX_ENV_NAME];
+	const env = { ...process.env };
+	env.RCLONE_CONFIG = tempConfigPath;
```
vicb: nit
```suggestion
	const env = { 
	  ...process.env,
	  RCLONE_CONFIG: tempConfigPath
	};
```

packages/cloudflare/src/cli/index.ts
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env node
 
+import "./load-envs.js";
+
```
vicb: I think this should be revert?

(same for `packages/cloudflare/src/cli/load-envs.ts` below)

packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -79,8 +79,12 @@ declare global {
 		CF_PREVIEW_DOMAIN?: string;
 		// Should have the `Workers Scripts:Read` permission
 		CF_WORKERS_SCRIPTS_API_TOKEN?: string;
-		// Cloudflare account id
+		// Cloudflare account id - needed for skew protection and R2 batch population
 		CF_ACCOUNT_ID?: string;
```
vicb, line 84:
```suggestion

		// Cloudflare account id - needed for skew protection and R2 batch population
CF_ACCOUNT_ID?: string;
```

ghost:
Hello @vicb I addressed all the comments.

vicb:
> Hello @vicb I addressed all the comments.

I think there is one opened comment about [rclone.promises](https://github.com/sntran/rclone.js/blob/8bd4e08afed96f44bd287e34068dfbf071095c9f/test/rclone-test.js#L133)

packages/cloudflare/src/cli/commands/populate-cache.ts
ghost: It is downloading rclone from official website after npm install
https://github.com/sntran/rclone.js/blob/main/package.json#L30

ghost:
@vicb could you please clarify what changes exactly you are requesting?
I already replied to your earlier comment about the rclone npm package, which is abandoned.

ghost:
I’m honestly starting to feel a bit sabotaged 😞

vicb:
> I already replied to your earlier comment about the rclone npm package, which is abandoned.

Either it's abandonned and we should not use it or we use it.

The alternative is to use any of the other more recent package on npm.

> I’m honestly starting to feel a bit sabotaged 😞

Sorry you feel that way.
That's a package we plan to maintain long-term so we take it the time it needs to review and merge the PRs.

packages/cloudflare/src/cli/commands/populate-cache.ts
ghost: Replaced rclone direct call with rclone.js promises.

packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -11,6 +12,7 @@ import type {
 } from "@opennextjs/aws/types/open-next.js";
 import type { IncrementalCache, TagCache } from "@opennextjs/aws/types/overrides.js";
 import { globSync } from "glob";
+import { promises as rclone } from "rclone.js";
```
vicb: Image

`import rclone from "rclone.js";` / `rclone.promises.copy` might work

packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -204,27 +206,140 @@ type PopulateCacheOptions = {
...
+			copyFileSync(fullPath, destPath);
+		}
+
+		// Use rclone.js to sync the R2
+		const remote = `r2:${bucket}`;
+
+		// Using rclone.js Promise-based API for the copy operation
+		await rclone.copy(stagingDir, remote, {
+			progress: true,
+			transfers: 32,
```
vicb: Could you please test with 16/8 instead of 32/16 to see if this makes a difference - you should delete your objects first to make sure the transfers actually take place and are not skipped by the checker because they already exist, or temp change the key for testing.

We can handle that in a follow up PR, the idea would be to set a reasonable default that can be overriden via `cacheChunkSize`

ghost: Updated

packages/cloudflare/src/cli/commands/populate-cache.ts
ghost: my bad, fixed

ghost:
Huge relief :) Thanks a lot for the approval and merge!

PR #934, opened 2025-10-08
[See #939] fix low-hanging fruits due to performance

anonrig:


packages/cloudflare/src/cli/build/patches/plugins/next-server.ts
vicb, line 1: I don't think this make any difference.

packages/cloudflare/src/cli/build/utils/extract-project-env-vars.ts
vicb, line 1: I don't think this make any difference.

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -13,6 +13,8 @@ export type LocalPattern = {
 	search?: string;
 };
 
+const NEXT_IMAGE_REGEXP = /\/_next\/image($|\/)/;
```
vicb: Should this be lazy rather than eager?

packages/cloudflare/src/cli/templates/init.ts
```diff
@@ -53,12 +53,6 @@ function init(request: Request, env: CloudflareEnv) {
 }
 
 function initRuntime() {
-	// Some packages rely on `process.version` and `process.versions.node` (i.e. Jose@4)
```
vicb, line 56: This should be reverted - it is not handled by neither unenv nor workerd for older compat date. We will be able to remove this at some point when we enforce a minimum compat date, not now.

packages/cloudflare/src/cli/templates/init.ts
```diff
@@ -83,10 +77,10 @@ function initRuntime() {
 				delete (init as { cache: unknown }).cache;
 				// https://github.com/cloudflare/workerd/issues/2746
 				// https://github.com/cloudflare/workerd/issues/3245
-				Object.defineProperty(init, "body", {
+				if (init.body instanceof stream.Readable) {
```
vicb, line 86: Please revert, this will fail with NextJS Proxy.

packages/cloudflare/src/cli/templates/skew-protection.ts
```diff
@@ -5,6 +5,8 @@ export const DEPLOYMENT_MAPPING_ENV_NAME = "CF_DEPLOYMENT_MAPPING";
 /** Version used for the latest worker */
 export const CURRENT_VERSION_ID = "current";
 
+let deploymentMappingEnvName: Record<string, string> | undefined;
```
vicb:
```suggestion
let deploymentMapping: Record<string, string> | undefined;
```

packages/cloudflare/src/cli/templates/skew-protection.ts
```diff
@@ -42,16 +44,16 @@ export function maybeGetSkewProtectionResponse(request: Request): Promise<Respon
 			return undefined;
 		}
 
-		const mapping = process.env[DEPLOYMENT_MAPPING_ENV_NAME]
+		deploymentMappingEnvName ??= process.env[DEPLOYMENT_MAPPING_ENV_NAME]
 			? JSON.parse(process.env[DEPLOYMENT_MAPPING_ENV_NAME])
 			: {};
 
-		if (!(requestDeploymentId in mapping)) {
+		if (!(requestDeploymentId in (deploymentMappingEnvName as Record<string, string>))) {
```
vicb: nit: the cast should be moved to l48 or probably adding `deploymentMapping &&` at the start of the condition is a better way

packages/cloudflare/src/cli/templates/skew-protection.ts
```diff
@@ -42,16 +44,16 @@ export function maybeGetSkewProtectionResponse(request: Request): Promise<Respon
...
 			// Unknown deployment id, serve the current version
 			return undefined;
 		}
 
-		const version = mapping[requestDeploymentId];
+		const version = deploymentMappingEnvName?.[requestDeploymentId];
```
vicb: is the `?` really needed given the previous if condition?

packages/cloudflare/src/cli/templates/init.ts
```diff
@@ -83,10 +83,10 @@ function initRuntime() {
...
 					// @ts-ignore
-					value: init.body instanceof stream.Readable ? ReadableStream.from(init.body) : init.body,
-				});
+					init.body = ReadableStream.from(init.body);
```
jasnell: Someone had mentioned that opennext was targeting node.js 18? If so, using `ReadableStream.from(...)` likely forces that to move up. Probably makes this a semver-major if that's the case

anonrig: But this is opennextjs-cloudflare?

packages/cloudflare/src/cli/templates/worker.ts
```diff
@@ -44,7 +44,7 @@ export default {
 				`${globalThis.__NEXT_BASE_PATH__}/_next/image${globalThis.__TRAILING_SLASH__ ? "/" : ""}`
 			) {
 				const imageUrl = url.searchParams.get("url") ?? "";
-				return await fetchImage(env.ASSETS, imageUrl, ctx);
+				return fetchImage(env.ASSETS, imageUrl, ctx);
```
jasnell: why remove the `await` here? It changes the error stack if this fails along with a few other smaller bits.

packages/cloudflare/src/cli/templates/init.ts
anonrig: And this line shouldn't be updated actually

PR #948, opened 2025-10-13
refactor: do not create a wrangler config when a custom one is passed

vicb:
Only ask whether to create a config file when no custom one is specified via `--config`.

Also the check is moved at the beginning of the build at the build aborts when the reply is No.

Both cases (no config, custom config) tested locally.

### DIFF
--- a/.changeset/thick-seas-walk.md
+++ b/.changeset/thick-seas-walk.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+refactor: do not create a wrangler config when a custom one is passed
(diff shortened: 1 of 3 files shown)


pnpm-workspace.yaml
```diff
@@ -35,7 +35,7 @@ catalog:
   typescript-eslint: ^8.37.0
   typescript: ^5.7.3
   vitest: ^2.1.1
-  wrangler: ^4.38.0
+  wrangler: https://pkg.pr.new/wrangler@10621
```
james-elicx: Code looks fine, just wanted to check if this was intentional before approving?

vicb: Good catch, that's my local test and should be reverted!

vicb:
```suggestion
  wrangler: ^4.38.0
```

vicb:
Thanks for the review and catching the wrong version James!

PR #955, opened 2025-10-20
add EsbuildOverride to defineCloudflareConfig (change build logging level)

bratvanov:
The `preview` script in my project (`opennextjs-cloudflare build && opennextjs-cloudflare preview`) was plagued by tens of `esbuild` warnings coming from third-party dependencies, and was extremely noisy.

Some of the warnings were coming from the `opennextjs-cloudflare preview` command – mostly `direct-eval` warnings. I was able to silence them by passing `--log-level error` to the `opennextjs-cloudflare preview` command.

However, I didn't find a way to configure logging in the `opennextjs-cloudflare build` command. Therefore, this PR adds the ability to completely override the esbuild config in the `open-next.config.ts`, like that:

```ts
import { defineCloudflareConfig } from "@opennextjs/cloudflare/config";

  esbuild: {
    bundleServer: (options) => ({
      ...options,
      logOverride: {
        ...options.logOverride,
        "impossible-typeof": "silent",
        "duplicate-object-key": "silent",
      },
    }),
  },
});
```

You should be able to override/re-config any aspect of esbuild, not just superficial logging overrides, e.g:

```ts
esbuild: {
  bundleServer: (options) => ({
    ...options,
    external: [
      ...(options.external ?? []),
      "sharp", // this package is huge; we let another runtime handle it instead
    ],
    define: {
      ...options.define,
      "process.env.SENTRY_RELEASE": JSON.stringify(
        process.env.SENTRY_RELEASE ?? "dev"
      ),
    },
  }),
},
```

Please let me know what you think. Thanks!

### DIFF
--- a/packages/cloudflare/src/cli/build/bundle-server.ts
+++ b/packages/cloudflare/src/cli/build/bundle-server.ts
@@ -7,7 +7,8 @@ import { type BuildOptions, getPackagePath } from "@opennextjs/aws/build/helper.
 import { ContentUpdater } from "@opennextjs/aws/plugins/content-updater.js";
 import { build, type BuildOptions as EsbuildBuildOptions, type Plugin } from "esbuild";
 
-import { getOpenNextConfig, type EsbuildOverride } from "../../api/config.js";
+import type { EsbuildOverride } from "../../api/config.js";
+import { getOpenNextConfig } from "../../api/config.js";
 import type { ProjectOptions } from "../project-options.js";
 import { patchVercelOgLibrary } from "./patches/ast/patch-vercel-og-library.js";
 import { patchWebpackRuntime } from "./patches/ast/webpack-runtime.js";


bratvanov:
I realized that even with the above changes patching `opennextjs-cloudflare build` – the warnings will not be suppressed when running the equivalent build command from wrangler (`wrangler build`). 

Therefore, I wonder if it wouldn't be better to propose and consider adding `--log-level` support for the `build` command into wrangler instead (and also to `wrangler deploy`, as it also surfaces esbuild warnings with no way to silence them)?

While the warnings are harmless, I have extensively integrated AI coding agents not just for local development, but also in CI/CD, and constantly burning through thousands of tokens re-processing the same benign warnings from external packages that I have no control over is adding a real financial cost.

As it stands, currently only `wrangler dev` (and by extension, `opennextjs-cloudflare preview`) supports the `--log-level` flag, which in my opinion, is insufficient.

For the time being, I've implemented custom build/deploy script wrappers  that filter the warnings from the original output, but this feels like a hack, as it's fragile and may break when/if the format of the warnings change. I'm open to suggestions on how to best tackle this. Thanks!

PR #963, opened 2025-10-27
feat: add serverExternals option to CloudflareOverrides and update build process

michalgrezel:


vicb:
Thanks for the PR!
Could you please add a description and tell of this was tested

Issue #826, opened 2025-07-28
[BUG] Windows: ERR_UNSUPPORTED_ESM_URL_SCHEME when previewing due to raw Windows paths in import()


cording12:
### Bug Description

Running `opennextjs-cloudflare preview` on Windows can fail with the following error:

```bash
Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
```

This occurs because the `retrieveCompiledConfig()` function in [cli/commands/utils.ts](https://github.com/opennextjs/opennextjs-cloudflare/blob/3de79624b63358af0881c21f1f93bbed1d9ca983/packages/cloudflare/src/cli/commands/utils.ts) can dynamically import a local file using a raw absolute Windows path (`C:\...`) rather than a proper `file://` URI on the line `const config = await import(configPath).then((mod) => mod.default);`.

This is potentially an issue only arising in monorepos, as I have not seen it happen elsewhere. The monorepo structure hoists `@opennext/cloudflare` to the `node_modules` in the project root.

This bug is also intermittent. The first time running the `preview` command can run successfully.

### Steps to reproduce

1. Create a turbo repo: `npx create-turbo@latest`
2. CD into `apps` and run `npm create cloudflare@latest -- my-next-app --framework=next --platform=workers`
3. Run `opennextjs-cloudflare preview`
4. Observe the error 

### Expected behavior

System is able to build and run correctly

### @opennextjs/cloudflare version

1.6.1

### Wrangler version

4.25.1

### next info output

```shell
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 10 Pro     
  Available memory (MB): 65365
  Available CPU cores: 16     
Binaries:
  Node: 22.17.0
  npm: 11.4.2
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  next: 15.3.0
  eslint-config-next: N/A
  react: 19.1.0
  react-dom: 19.1.0
  typescript: 5.8.3
Next.js Config:
  output: N/A
```

### Additional context

This can be temporarily patched for Windows users if you are running into this bug. A quick manual patch:

1. Locate the `utils.ts` file located in `root\node_modules\@opennextjs\cloudflare\dist\cli\commands`
2. Add `import { pathToFileURL } from "url";` to the top of the file
[...]

PR #968, opened 2025-10-30
fix: Compiled config is now imported using pathToFileURL to avoid crashes on Windows

rgembalik:
For the moment, this is the only bug preventing me from building and deploying with opennext to cloudflare. With this fix, build and publish passes without issues. 

The solution was described in #826 - after testing, I decided to file a PR for it. 

Feel free to suggest fixes if I missed anything. If there are other fixes needed that require windows testing, ping me, I might be able to test and fix other issues as well.

### DIFF
--- a/.changeset/clean-lions-tan.md
+++ b/.changeset/clean-lions-tan.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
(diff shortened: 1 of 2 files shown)


vicb (approved):
Thanks for the PR!

PR #973, opened 2025-11-04
Add standard Next template generated from create-next-app that will be used by create-cloudflare going forward

petebacondarwin:
This setup was tested locally by running the following commands:

```bash
cd create-cloudflare/next
npm i
npm run preview
```

Notes:

- npm is used here rather than pnpm since this package is not officially part of the monorepo
- `npm run preview` does the open next build and then runs the project with wrangler - the project can also be run with `npm run dev` to use the classic Next.js dev server

### DIFF
--- a/create-cloudflare/next/public/_headers
+++ b/create-cloudflare/next/public/_headers
@@ -1,3 +1,4 @@
 # https://developers.cloudflare.com/workers/static-assets/headers
+# https://opennext.js.org/cloudflare/caching#static-assets-caching
 /_next/static/*
   Cache-Control: public,max-age=31536000,immutable


.changeset/better-paws-shout.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/cloudflare": minor
+---
+
+Add standard Next template generated from create-next-app that will be used by create-cloudflare going forward
```
petebacondarwin, line 5: Not sure if this should be prefixed with something like `chore:`?

vicb: We do not enforce that in the repo (we sometimes use it for core features/fixes)

create-cloudflare/next/.gitignore
```diff
@@ -0,0 +1,44 @@
...
+# typescript
+*.tsbuildinfo
+next-env.d.ts
+
+# OpenNext
+/.open-next
+
+# wrangler files
+.wrangler
+!.dev.vars*
```
vicb: I think `.dev.vars*` should be excluded (we might not do this in all examples for testing purposes)

create-cloudflare/next/.dev.vars.example
vicb, line 1: rename to `.dev.vars.example` and exclude in `.gitignore`

create-cloudflare/next/next.config.ts
```diff
@@ -0,0 +1,11 @@
+import type { NextConfig } from "next";
+
+const nextConfig: NextConfig = {
+  /* config options here */
+};
+
+export default nextConfig;
+
+// added by create cloudflare to enable calling `getCloudflareContext()` in `next dev`
```
vicb:
```suggestion
// Enable calling `getCloudflareContext()` in `next dev`
// see https://opennext.js.org/cloudflare/bindings#local-access-to-bindings
```

create-cloudflare/next/package.json
```diff
@@ -0,0 +1,33 @@
+{
+	"name": "next",
+	"version": "0.1.0",
+	"private": true,
```
vicb, line 4: Should templates be private?

create-cloudflare/next/wrangler.jsonc
```diff
@@ -0,0 +1,18 @@
+/**
+ * For more details on how to configure Wrangler, refer to:
+ * https://developers.cloudflare.com/workers/wrangler/configuration/
+ */
+{
+	"$schema": "node_modules/wrangler/config-schema.json",
+	"name": "next",
+	"main": ".open-next/worker.js",
+	"compatibility_date": "2025-03-01",
```
vicb: We could probably bump to the latest supported date in the installed wrangler

create-cloudflare/next/.gitignore
petebacondarwin: Yeah the project generated by C3 had it ignored in `.gitignore` but then in the docs for Open Next Cloudflare, I saw that it suggested you add this so that Next runs in dev mode locally...

https://opennext.js.org/cloudflare/get-started#5-add-a-devvars-file

So I thought it might be nice to include that...

But perhaps what we should do is include that in the template (by hard adding it) but then exclude it in the `.gitignore` file of the template so that it doesn't get git commited?

create-cloudflare/next/package.json
petebacondarwin: I think so... this means that they are not deployable to npm, right?

vicb:
Thanks Pete,
LGTM
Could you please add details on how this was tested in the PR description.
You also need to run `pnpm fix`

create-cloudflare/next/.dev.vars.example
petebacondarwin: How about we leave it in the template but then ignore it in `.gitignore`?

create-cloudflare/next/wrangler.jsonc
petebacondarwin: I can change C3 to update this when it instantiates the template.

create-cloudflare/next/.dev.vars.example
vicb: Wouldn't that be a pain if we want to update the file at any point?

create-cloudflare/next/.gitignore
```diff
@@ -0,0 +1,45 @@
...
+/.pnp
+.pnp.js
+.yarn/install-state.gz
+
+# testing
+/coverage
+
+# next.js
+/.next/
+/out/
```
vicb, line 18: nit for consistency
```suggestion
/.next
/out
```

create-cloudflare/next/.gitignore
```diff
@@ -0,0 +1,45 @@
...
+
+# local env files
+.env*.local
+
+# vercel
+.vercel
+
+# typescript
+*.tsbuildinfo
+next-env.d.ts
```
vicb, line 42: we should move this to the next.js section

create-cloudflare/next/.gitignore
```diff
@@ -0,0 +1,45 @@
...
+
+# OpenNext
+/.open-next
+
+# wrangler files
+.wrangler
+.dev.vars*
+!.dev.vars.example
+!.env.example
```
vicb: we should also move this to the next.js section (wrangler is setup to not process `.env` files with the adapter)

create-cloudflare/next/package.json
vicb: Oh right, I was thinking about the instantiated templates.

create-cloudflare/next/wrangler.jsonc
vicb: A fixed date should be good for now IMO

create-cloudflare/next/.gitignore
petebacondarwin: Be aware that this file is basically what is generated by create-next-app. See https://github.com/opennextjs/opennextjs-cloudflare/blob/main/examples/create-next-app/.gitignore

The only modifications are those that C3 did to add the `.dev.vars` entries.

create-cloudflare/next/wrangler.jsonc
petebacondarwin: Done

create-cloudflare/next/.gitignore
petebacondarwin: Done

create-cloudflare/next/.gitignore
petebacondarwin: this is what the create-next-app generates

create-cloudflare/next/package.json
```diff
@@ -0,0 +1,33 @@
...
+		"start": "next start",
+		"lint": "next lint",
+		"deploy": "opennextjs-cloudflare build && opennextjs-cloudflare deploy",
+		"upload": "opennextjs-cloudflare build && opennextjs-cloudflare upload",
+		"preview": "opennextjs-cloudflare build && opennextjs-cloudflare preview",
+		"cf-typegen": "wrangler types --env-interface CloudflareEnv ./cloudflare-env.d.ts"
+	},
+	"dependencies": {
+		"@opennextjs/cloudflare": "^1.11.0",
+		"next": "15.4.6",
```
vicb: The c3 template was not updated because of a CI error on Linux.
Turns out that was only because turbopack is used by default on Linux.
This should be fixed by https://github.com/cloudflare/workers-sdk/pull/11014 and we can update here
_Note that the e2e tests in this repo are already running [15.5.6](https://github.com/opennextjs/opennextjs-cloudflare/blob/73b711187b2dffb67db4dfeec05df2fbc6506bbb/pnpm-workspace.yaml#L48)_  

```suggestion
		"next": "15.5.6",
```

create-cloudflare/next/public/_headers
```diff
@@ -0,0 +1,3 @@
+# https://developers.cloudflare.com/workers/static-assets/headers
```
vicb, line 1:
```suggestion
# https://developers.cloudflare.com/workers/static-assets/headers
# https://opennext.js.org/cloudflare/caching#static-assets-caching
```

vicb (approved):
Thanks Pete!

PR #982, opened 2025-11-10
misc minor changes

vicb:
- lock file updates are a side effect of running `npx update-browserslist-db@latest` but great to have
- while `next-env.d.ts` was in the `.gitignore`, it was still tracked by git.


vicb:
Thanks for the review James!

PR #956, opened 2025-10-20
Add support for turbopack in Next 16

conico974:
Introduce support for turbopack for Next 16+.
Bump e2e/experimental to 16.

It's a draft for now, need to investigate this https://github.com/opennextjs/opennextjs-aws/issues/1022 (might not be necessary in every case)

PR #983, opened 2025-11-10
feat: turbopack support

vicb:
The code here is extracted from https://github.com/opennextjs/opennextjs-cloudflare/pull/956 by @conico974 🙏

It does not update the experimental example as composable cache fails on Next 16 for now.

### DIFF
--- a/packages/cloudflare/src/cli/build/bundle-server.ts
+++ b/packages/cloudflare/src/cli/build/bundle-server.ts
@@ -50,14 +50,11 @@ export async function bundleServer(buildOpts: BuildOptions, projectOpts: Project
 	copyPackageCliFiles(packageDistDir, buildOpts);
 
 	const { appPath, outputDir, monorepoRoot, debug } = buildOpts;
-	const baseManifestPath = path.join(
-		outputDir,
-		"server-functions/default",
-		getPackagePath(buildOpts),
-		".next"
-	);
-	const serverFiles = path.join(baseManifestPath, "required-server-files.json");
+	const dotNextPath = path.join(outputDir, "server-functions/default", getPackagePath(buildOpts), ".next");
+	const serverFiles = path.join(dotNextPath, "required-server-files.json");
 	const nextConfig = JSON.parse(fs.readFileSync(serverFiles, "utf-8")).config;
+
+	const useTurbopack = fs.existsSync(path.join(dotNextPath, "server/chunks/[turbopack]_runtime.js"));
 
 	console.log(`\x1b[35m⚙️ Bundling the OpenNext server...\n\x1b[0m`);
 
@@ -141,6 +138,9 @@ export async function bundleServer(buildOpts: BuildOptions, projectOpts: Project
 			// Note: we need the __non_webpack_require__ variable declared as it is used by next-server:
 			// https://github.com/vercel/next.js/blob/be0c3283/packages/next/src/server/next-server.ts#L116-L119
 			__non_webpack_require__: "require",
+			// The 2 following defines are used to reduce the bundle size by removing unnecessary code
+			// Next uses different precompiled renderers (i.e. `app-page.runtime.prod.js`) based on if you use `TURBOPACK` or some experimental React features
+			...(useTurbopack ? {} : { "process.env.TURBOPACK": "false" }),
 			// We make sure that environment variables that Next.js expects are properly defined
 			"process.env.NEXT_RUNTIME": '"nodejs"',
 			"process.env.NODE_ENV": '"production"',


vicb:
There seems to be a legit issue with loading `og`, will add a fixup commit

packages/cloudflare/src/cli/build/bundle-server.ts
```diff
@@ -144,10 +144,6 @@ export async function bundleServer(buildOpts: BuildOptions, projectOpts: Project
 			// We make sure that environment variables that Next.js expects are properly defined
 			"process.env.NEXT_RUNTIME": '"nodejs"',
 			"process.env.NODE_ENV": '"production"',
-			// The 2 following defines are used to reduce the bundle size by removing unnecessary code
-			// Next uses different precompiled renderers (i.e. `app-page.runtime.prod.js`) based on if you use `TURBOPACK` or some experimental React features
-			// Turbopack is not supported for build at the moment, so we disable it
-			"process.env.TURBOPACK": "false",
```
conico974, line 150: I think we should do that only if webpack is not used, otherwise it may increase the size of the bundle for these cases.

packages/cloudflare/src/cli/build/patches/ast/patch-vercel-og-library.ts
```diff
@@ -40,17 +42,23 @@ export function patchVercelOgLibrary(buildOpts: BuildOptions) {
...
-			const fontFileName = matches[0]!.getMatch("PATH")!.text();
-			renameSync(path.join(outputDir, fontFileName), path.join(outputDir, `${fontFileName}.bin`));
+			if (matches.length > 0) {
+				const fontFileName = matches[0]!.getMatch("PATH")!.text();
+				renameSync(path.join(outputDir, fontFileName), path.join(outputDir, `${fontFileName}.bin`));
+			}
 		}
 
 		// Change node imports for the library to edge imports.
+		// This is only useful when tubopack is not used to bundle the function.
```
conico974:
```suggestion
		// This is only useful when turbopack is not used to bundle the function.
```

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
```diff
@@ -0,0 +1,86 @@
...
+	)
+	.join("\n")}
+      default:
+        throw new Error(\`Not found \${chunkPath}\`);
+    }
+  }
+`;
+}
+
+// Turbopack imports `og` via `externalImport`.
```
conico974, line 67: This may not be the only lib where turbopack does that. Might be worth adding a comment with the error

pnpm-lock.yaml
conico974, line 1: Was it necessary to update the lockfile ?

packages/cloudflare/src/cli/build/bundle-server.ts
vicb: Do you mean `"process.env.TURBOPACK": "true"` when turbopack is used?
The problem with doing that is that the next build generates `process.env.TURBOPACK = "1";` and then ESBuild emits a build warning.

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
vicb: Great idea, we'll probably get more :(
Added the comment

pnpm-lock.yaml
vicb: Good catch, I probably messed when extracting the changes from your original PR

vicb:
Great feedback, thanks Nico

packages/cloudflare/src/cli/build/bundle-server.ts
conico974: No we keep it as false when webpack is used and remove the line when it's turbopack.

vicb: Added a fixup commit for that, thanks!

rvanlaak:
Can confirm that with this PR we're able to upgrade to Next 16 on Cloudflare Workers, where before this PR was released (v1.11.0) the preview URLs resulted in an "Internal server error" message 🎉

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
cvbuelow: i'm getting this error for other modules that are not packaged by next:
```
✘ [ERROR]  ⨯ Error: Failed to load external module swr: Error: No such module "swr".

      at Context.externalImport [as y]
  (file:///xxxxx/.open-next/server-functions/default/handler.mjs:1279:100622)
      at null.<anonymous> (async
  file:///xxxxx/.wrangler/tmp/dev-nsmCkq/worker.js:45821:20)
```
do i need to add them all here?

PR #991, opened 2025-11-17
use `wrangler r2 bulk put` for R2 cache population

vicb:
Use `wrangler r2 bulk put` instead of relying on `rclone`

The DX with `rclone` is not great:
- need to create S3 tokens and provide them as environment variables
- does not work well across all OSes (see pending issues)
- does not work in local dev - albeit R2 population is as slow as with remote objects

`wrangler r2 bulk put` is instant in local dev and fast for remote buckets (not as fast as rclone).

The e2e for e2e/app-router lasts 5min before this PR and <50s after.

TODO:
- [x] bump wrangler when https://github.com/cloudflare/workers-sdk/pull/11285 is released

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -85,7 +85,7 @@ catalogs:
       specifier: ^2.1.1
       version: 2.1.1
     wrangler:
-      specifier: 4.49.0
+      specifier: ^4.49.0
       version: 4.49.0
     yargs:
       specifier: ^18.0.0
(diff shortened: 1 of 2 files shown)


vicb:
> Wow! this looks like a great UX improvement! 🫶
> 
> I think that https://opennext.js.org/cloudflare/cli#populatecache-command also needs updating?

Thanks for the review and the reminder on the docs. I'll update the docs right after the release.

PR #997, opened 2025-11-19
Add ability to pass multiple configuration files to the dev server

mattvick:
Adds support for developing with multiple Workers. Specifically, running multiple Workers in a single dev command by passing multiple configuration files to the server.

This is the [documented and recommended approach](https://developers.cloudflare.com/workers/development-testing/multi-workers/#single-dev-command) for most development workflows as it ensures the best compatibility with bindings.

This is needed when using [cross-script calls](https://developers.cloudflare.com/workflows/build/workers-api/#cross-script-calls) for Workflows, Queues, and other features based on Durable Objects. As outlined in the [supported bindings](https://developers.cloudflare.com/workers/wrangler/api/#supported-bindings) section of the Workers docs.

### DIFF
--- a/.changeset/whole-coats-poke.md
+++ b/.changeset/whole-coats-poke.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": minor
+---
+
+Add ability to pass multiple configuration files to the dev server


vicb:
See the comments in https://github.com/opennextjs/opennextjs-cloudflare/pull/896

PR #999, opened 2025-11-20
Add image optimization with Cloudflare Images

pilcrowonpaper:


packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -13,120 +13,463 @@ export type LocalPattern = {
...
+				return new Response('"url" parameter is valid but upstream response timed out', {
+					status: 504,
+				});
+			}
+			if (fetchImageResult.error === "too_many_redirects") {
+				return new Response('"url" parameter is valid but upstream response is invalid', {
+					status: 508,
+				});
+			}
+			throw new Error("Failed to fetch image");
```
pilcrowonpaper, line 70: This is consistent with Next.js but not sure if a 500 response is the best if the URL is invalid

sommeeeer:
Hello @pilcrowonpaper, good to see you here :) looks good on first glance, I'll do another review later!

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -13,120 +13,463 @@ export type LocalPattern = {
...
 
-		if (url == null) {
-			return getUrlErrorResponse();
+	let immutable = false;
+	if (parseResult.static) {
+		immutable = true;
+	} else {
+		const cacheControlHeader = imageResponse.headers.get("Cache-Control");
+		if (cacheControlHeader !== null) {
+			immutable = cacheControlHeader.includes("immutable");
```
pilcrowonpaper, line 88: Next.js does something more complicated and bases the optimized response on the upstream `Cache-Control` header. I didn't want to do anything complex for now so it just checks if the upstream image is a static content

pilcrowonpaper:
It looks like the Playwright tests are failing because the test cases don't include the required `w` and `q` parameters.

packages/cloudflare/src/cli/build/open-next/compile-images.ts
```diff
@@ -19,14 +19,25 @@ export async function compileImages(options: BuildOptions) {
...
+		imagesManifest?.images?.minimumCacheTTL ?? defaultMinimumCacheTTL
+	);
 	const __IMAGES_ALLOW_SVG__ = JSON.stringify(Boolean(imagesManifest?.images?.dangerouslyAllowSVG));
 	const __IMAGES_CONTENT_SECURITY_POLICY__ = JSON.stringify(
 		imagesManifest?.images?.contentSecurityPolicy ?? "script-src 'none'; frame-src 'none'; sandbox;"
 	);
 	const __IMAGES_CONTENT_DISPOSITION__ = JSON.stringify(
 		imagesManifest?.images?.contentDispositionType ?? "attachment"
 	);
+	const __IMAGES_MAX_REDIRECTS__ = JSON.stringify(imagesManifest?.images?.maximumRedirects ?? 3);
```
pilcrowonpaper: This was changed to 3 from unlimited in Next.js 16

packages/cloudflare/src/cli/build/open-next/compile-images.ts
```diff
@@ -38,10 +49,23 @@ export async function compileImages(options: BuildOptions) {
...
 			__IMAGES_CONTENT_SECURITY_POLICY__,
 			__IMAGES_CONTENT_DISPOSITION__,
+			__IMAGES_MAX_REDIRECTS__,
 		},
 	});
 }
+
+const defaultDeviceSizes = [640, 750, 828, 1080, 1200, 1920, 2048, 3840];
+const defaultImageSizes = [32, 48, 64, 96, 128, 256, 384];
+const defaultQualities = [75];
```
pilcrowonpaper, line 73: Before Next.js 16, the default behavior was to allow values from 1-100

packages/cloudflare/src/cli/build/open-next/compile-images.ts
```diff
@@ -38,10 +49,23 @@ export async function compileImages(options: BuildOptions) {
...
 			__IMAGES_ALLOW_SVG__,
 			__IMAGES_CONTENT_SECURITY_POLICY__,
 			__IMAGES_CONTENT_DISPOSITION__,
+			__IMAGES_MAX_REDIRECTS__,
 		},
 	});
 }
+
+const defaultDeviceSizes = [640, 750, 828, 1080, 1200, 1920, 2048, 3840];
+const defaultImageSizes = [32, 48, 64, 96, 128, 256, 384];
```
pilcrowonpaper, line 70: Before Next.js 16, `16` was also included

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -13,120 +13,463 @@ export type LocalPattern = {
...
+
+	return new Response('"url" parameter is valid but image type is not allowed', {
+		status: 400,
+	});
+}
+
+async function fetchImage(url: string, count: number): Promise<FetchImageResult> {
+	let response: Response;
+	try {
+		response = await fetch(url, {
```
pilcrowonpaper, line 204: I haven't added support for `dangerouslyAllowLocalIP`, which checks if a domain resolves to a private IP.

packages/cloudflare/src/cli/build/open-next/compile-images.ts
vicb: We could use unlimited then (and maybe add your comment as a code comment on this file/line)

packages/cloudflare/src/cli/build/open-next/compile-images.ts
vicb: I would use Next 16 (and add comments on the lines).

nit: maybe move those constants up (i.e. before they are used)

packages/cloudflare/src/cli/templates/worker.ts
```diff
@@ -43,8 +42,7 @@ export default {
 				url.pathname ===
 				`${globalThis.__NEXT_BASE_PATH__}/_next/image${globalThis.__TRAILING_SLASH__ ? "/" : ""}`
 			) {
-				const imageUrl = url.searchParams.get("url") ?? "";
-				return await fetchImage(env.ASSETS, imageUrl, ctx);
+				return await handleImageRequest(url, request.headers, env);
```
vicb, line 45: nit: maybe pass the `request` instead of `url` and `headers` to `handleImageRequest` ?

pilcrowonpaper:
@vicb 
Should the behavior match Next.js 15 or 16 in general? Or should we check the version and change the behavior based on it? Not sure what the policy is on Next.js versions

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -13,120 +13,463 @@ export type LocalPattern = {
 	search?: string;
 };
 
-let NEXT_IMAGE_REGEXP: RegExp;
+export async function handleImageRequest(
```
vicb, line 30: Could you please add JSDoc to this function

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -13,120 +13,463 @@ export type LocalPattern = {
...
+	const localPatterns = __IMAGES_LOCAL_PATTERNS__;
+	const qualities: number[] = __IMAGES_QUALITIES__;
+
+	const urlQueryValues = requestURL.searchParams.getAll("url");
+	if (urlQueryValues.length < 1) {
+		const result: ErrorResult = {
+			ok: false,
+			message: '"url" parameter is required',
+		};
+		return result;
```
vicb, line 365: nit: That should be enough?

```suggestion
		return {
			ok: false,
			message: '"url" parameter is required',
		};
```

packages/cloudflare/src/cli/templates/worker.ts
pilcrowonpaper: We already create a `URL` instance in the worker so this allows us to reuse it and avoid parsing the URL string twice

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -13,120 +13,463 @@ export type LocalPattern = {
...
-		if (!contentType) {
-			// Fallback to upstream header when the type can not be detected
-			// https://github.com/vercel/next.js/blob/d76f0b1/packages/next/src/server/image-optimizer.ts#L748
-			contentType = imgResponse.headers.get("content-type") ?? "";
+	let url: string;
+	if (urlQueryValue.startsWith("/")) {
+		url = urlQueryValue;
+
+		const pathname = getPathnameFromRelativeURL(url);
+		if (/\/_next\/image($|\/)/.test(decodeURIComponent(pathname))) {
```
vicb, line 396: What about `/_next/image?url`

packages/cloudflare/src/cli/build/open-next/compile-images.ts
```diff
@@ -19,14 +19,25 @@ export async function compileImages(options: BuildOptions) {
 		: {};
 
 	const __IMAGES_REMOTE_PATTERNS__ = JSON.stringify(imagesManifest?.images?.remotePatterns ?? []);
+	const __IMAGES_LOCAL_PATTERNS_DEFINED__ = JSON.stringify(
```
vicb: Is this needed given that `__IMAGES_LOCAL_PATTERNS__` default to an empty list?

pilcrowonpaper: Yes, Next.js treats undefined and an empty list differently

packages/cloudflare/src/cli/templates/images.ts
pilcrowonpaper: Personally I don't like returning objects directly and prefer explicitly defining the type. I can change it if it bothers you

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -13,120 +13,463 @@ export type LocalPattern = {
...
+	}
+	if (widthQueryValues.length > 1) {
+		const result: ErrorResult = {
+			ok: false,
+			message: '"w" parameter (width) cannot be an array',
+		};
+		return result;
+	}
+	const widthQueryValue = widthQueryValues[0]!;
+	if (!/^[0-9]+$/.test(widthQueryValue)) {
```
vicb, line 463: nit
```suggestion
	if (!/^\d+$/.test(widthQueryValue)) {
```

packages/cloudflare/src/cli/templates/images.ts
pilcrowonpaper: Sorry, I don't understand?

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -13,120 +13,463 @@ export type LocalPattern = {
...
+			const result: ErrorResult = {
+				ok: false,
+				message: '"url" parameter is invalid',
+			};
+			return result;
+		}
+		if (!hasRemoteMatch(remotePatterns, parsedURL)) {
+			const result: ErrorResult = {
+				ok: false,
+				message: '"url" parameter is not allowed',
```
vicb: maybe you can drop ok and rename "message" to "error"?

packages/cloudflare/src/cli/templates/images.ts
pilcrowonpaper: For context, most of the parameter validation is based on Next.js' own code

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -13,120 +13,463 @@ export type LocalPattern = {
...
 
-		// Cancel the unused stream
-		ctx.waitUntil(respBody.cancel());
+	const result: ParseImageRequestURLSuccessResult = {
+		ok: true,
+		url: url,
+		width: width,
+		quality: quality,
+		format: format,
+		static: staticAsset,
```
vicb: nit:
```suggestion
		url,
		width,
		quality,
		format,
		static: staticAsset,
```

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -13,120 +13,463 @@ export type LocalPattern = {
...
+}
 
-		return new Response('"url" parameter is valid but image type is not allowed', {
-			status: 400,
-		});
-	} catch {
-		return new Response('"url" parameter is valid but upstream response is invalid', {
-			status: 400,
-		});
+type ParseImageRequestURLResult = ParseImageRequestURLSuccessResult | ErrorResult;
```
vicb: nit: move the type above the function where they are used

packages/cloudflare/src/cli/templates/images.ts
pilcrowonpaper: I personally don't like dealing with `undefined` and optional parameters so I prefer having with an `ok` property that is always defined, but let me know if you want me to change it

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -163,28 +506,22 @@ export function matchLocalPattern(pattern: LocalPattern, url: URL): boolean {
...
- */
-function getUrlErrorResponse() {
-	return new Response(`"url" parameter is not allowed`, { status: 400 });
-}
-
 const AVIF = "image/avif";
 const WEBP = "image/webp";
 const PNG = "image/png";
 const JPEG = "image/jpeg";
-const JXL = "image/jxl";
```
vicb, line 177: Question: why have those been removed? (IIRC it comes from Next)

pilcrowonpaper: I've removed formats that aren't supported by Cloudflare:
https://developers.cloudflare.com/images/transform-images/

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -13,120 +13,463 @@ export type LocalPattern = {
...
- * Remote images should match the configured remote patterns or a 404 response is returned.
- */
-export async function fetchImage(fetcher: Fetcher | undefined, imageUrl: string, ctx: ExecutionContext) {
-	// https://github.com/vercel/next.js/blob/d76f0b1/packages/next/src/server/image-optimizer.ts#L208
-	if (!imageUrl || imageUrl.length > 3072 || imageUrl.startsWith("//")) {
-		return getUrlErrorResponse();
+	let imageResponse: Response;
+	if (parseResult.url.startsWith("/")) {
+		if (env.ASSETS === undefined) {
+			console.error("env.ASSETS binding is not defined.");
```
vicb: use the logger rather than console.

Also we should fallback to the former behavior (serve unoptimized images) when the binding is not present)

vicb: Oops that's ASSET, nvm the comment about former behavior but still please switch to using the logger.

pilcrowonpaper: I've had some issues with wrangler not being able resolve `@opennextjs/aws` imports but I'll look into it again

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -13,120 +13,463 @@ export type LocalPattern = {
...
+		}
 
-	if (url.protocol !== "http:" && url.protocol !== "https:") {
-		return getUrlErrorResponse();
+		const outputFormat = parseResult.format ?? contentType;
+		const imageSource = env.IMAGES.input(imageStream);
+		const imageTransformationResult = await imageSource
+			.transform({
+				width: parseResult.width,
+				fit: "scale-down",
```
vicb, line 164: question: Does ~not~ Next always use scale-down?

pilcrowonpaper: Unsure. There's so no documentation on the default value
https://developers.cloudflare.com/images/transform-images/transform-via-url/#fit

packages/cloudflare/src/cli/templates/images.ts
vicb: It would not match as you have a `$`

packages/cloudflare/src/cli/templates/images.ts
vicb: you can use `satisfies` if you really want to add a type.
But all "nit" comments are mostly suggestion, pick what you prefer.

packages/cloudflare/src/cli/templates/images.ts
vicb: What happens if the image binding is not used and we try to access one of this removed type?

packages/cloudflare/src/cli/templates/images.ts
vicb: Sorry for the typo in my initial message, the question was about Next

packages/cloudflare/src/cli/templates/images.ts
pilcrowonpaper: It matches against the pathname so it should be fine, no? It matches against `/_next/image`, `/_next/image/`, and `/_next/image/*`

packages/cloudflare/src/cli/templates/images.ts
pilcrowonpaper: That part I was unsure. I personally think we should only serve images that are supported by the optimization process, even when the binding isn't defined. It'd be weird if your images stopped being served when you turn on image optimization by defining the binding. It will be a breaking change though

packages/cloudflare/src/cli/templates/images.ts
pilcrowonpaper: I'll need to double check, but I just assumed so because you mentioned that the images shouldn't be scaled more than the original size. (For context, `scale-down` will only scale down and never scale up images.)

pilcrowonpaper: Yes. Next.js doesn't scale the images more that the original

vicb: Thinking about [Object Fit](https://nextjs.org/docs/app/api-reference/components/image#:~:text=position%3A%20%22absolute%22.-,Object%20Fit,-%3A) here

pilcrowonpaper: I think that's a client-side/CSS thing? The server endpoint doesn't have any query parameters that controls that behavior

packages/cloudflare/src/cli/build/open-next/compile-images.ts
```diff
@@ -19,29 +19,53 @@ export async function compileImages(options: BuildOptions) {
...
 	const __IMAGES_CONTENT_DISPOSITION__ = JSON.stringify(
 		imagesManifest?.images?.contentDispositionType ?? "attachment"
 	);
+	const __IMAGES_MAX_REDIRECTS__ = JSON.stringify(imagesManifest?.images?.maximumRedirects ?? 3);
 
 	await build({
 		entryPoints: [imagesPath],
 		outdir: path.join(options.outputDir, "cloudflare"),
-		bundle: false,
+		bundle: true,
```
pilcrowonpaper, line 46: Will this cause any issues? Had to enable it for esbuild to resolve the import from `@opennextjs/aws/adapters/logger.js`

examples/playground14/e2e/cloudflare.spec.ts
vicb, line 1: We probably need to add a test making sure that the image has been resized (checking the w)?

packages/cloudflare/src/api/cloudflare-context.ts
```diff
@@ -13,6 +13,9 @@ declare global {
 		// Asset binding
 		ASSETS?: Fetcher;
 
+		// Images binding
```
vicb: Could you please add a comment saying that it is used to resize image when configured?

packages/cloudflare/src/cli/build/open-next/compile-images.ts
vicb: How so?
If the behavior is different then we could pass the value differently to avoid having to defines.

packages/cloudflare/src/cli/build/open-next/compile-images.ts
vicb: What's the status here?

packages/cloudflare/src/cli/build/open-next/compile-images.ts
vicb: It shouldn't.

packages/cloudflare/src/cli/build/open-next/compile-images.ts
vicb: What's the status here?

packages/cloudflare/src/cli/build/open-next/compile-images.ts
vicb: What's the status here?

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -13,120 +15,491 @@ export type LocalPattern = {
 	search?: string;
 };
 
-let NEXT_IMAGE_REGEXP: RegExp;
-
 /**
- * Fetches an images.
+ * Handles requests to /_next/image(/), including image optimizations.
+ * Image optimization is disabled if env.IMAGES is undefined.
```
vicb: nit:
```suggestion
 * Image optimization is disabled if env.IMAGES is undefined (the source image is returned unchanged)
```

packages/cloudflare/src/cli/templates/images.ts
vicb: Could you please add a note in the JSDoc?

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -13,120 +15,491 @@ export type LocalPattern = {
...
+			throw new Error("Failed to fetch image");
+		}
+		imageResponse = fetchImageResult.response;
 	}
 
-	// Local
-	if (imageUrl.startsWith("/")) {
-		// @ts-expect-error TS2339 Missing types for URL.parse
-		const url = URL.parse(imageUrl, "http://n");
+	if (imageResponse.status !== 200 || imageResponse.body === null) {
```
vicb: nit
```suggestion
	if (!imageResponse.ok || imageResponse.body === null) {
```

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -13,120 +15,491 @@ export type LocalPattern = {
...
+		return new Response('"url" parameter is valid but upstream response is invalid', {
+			status: imageResponse.status,
```
vicb, line 77: Should we use a fixed status here?

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -13,120 +15,491 @@ export type LocalPattern = {
...
+	});
+	const readImageHeaderBytesResult = await contentTypeImageReader.readAtLeast(32, imageHeaderBytes);
+	if (readImageHeaderBytesResult.value === undefined) {
+		await imageResponse.body.cancel();
 
-		NEXT_IMAGE_REGEXP ??= /\/_next\/image($|\/)/;
-		if (NEXT_IMAGE_REGEXP.test(pathname)) {
-			return getUrlErrorResponse();
+		return new Response('"url" parameter is valid but upstream response is invalid', {
+			status: imageResponse.status,
```
vicb: ditto for fixed status - IIUC this will probably be 200 at this point?

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -13,120 +15,491 @@ export type LocalPattern = {
...
-	// Remote
-	let url: URL;
-	try {
-		url = new URL(imageUrl);
-	} catch {
-		return getUrlErrorResponse();
-	}
+	if (contentType === AVIF || contentType === WEBP || contentType === JPEG || contentType === PNG) {
+		if (env.IMAGES === undefined) {
+			warn("env.IMAGES binding is not defined.");
```
vicb: nit: add actionable details to the warning (saying that the `${contentType}` can not be transformed)

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -13,120 +15,491 @@ export type LocalPattern = {
...
+		status: 400,
+	});
+}
+
+/**
+ * Fetch call with max redirects and timeouts.
+ *
+ * Re-throws the exception thrown by a fetch call.
+ * @param url
+ * @param timeout Timeout for a single fetch call.
```
vicb: nit: timeout -> timeoutMs

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -13,120 +15,491 @@ export type LocalPattern = {
...
+async function fetchWithRedirects(
+	url: string,
+	timeout: number,
+	maxRedirectCount: number
+): Promise<FetchWithRedirectsResult> {
+	let response: Response;
+	try {
+		response = await fetch(url, {
+			signal: AbortSignal.timeout(timeout),
+			redirect: "manual",
```
vicb, line 206: nit: only use "manual" when redirect is not max?

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -13,120 +15,491 @@ export type LocalPattern = {
...
+	contentType: string,
+	imageResponseFlags: ImageResponseFlags
+): Response {
+	const response = new Response(image);
+	response.headers.set("Vary", "Accept");
+	response.headers.set("Content-Type", contentType);
+	response.headers.set("Content-Disposition", __IMAGES_CONTENT_DISPOSITION__);
+	response.headers.set("Content-Security-Policy", __IMAGES_CONTENT_SECURITY_POLICY__);
+	if (imageResponseFlags.immutable) {
+		response.headers.set("Cache-Control", "public, max-age=315360000, immutable");
```
vicb: nit: passing this in the Response ctor (options.headers) might be less verbose

packages/cloudflare/src/cli/templates/images.ts
```diff
@@ -13,120 +15,491 @@ export type LocalPattern = {
...
-		// `ctx.waitUntil` keeps the runtime running until the promise settles without having to wait here.
-		ctx.waitUntil(reader.cancel());
-
-		if (value) {
-			contentType = detectContentType(value);
-		}
+function parseImageRequest(requestURL: URL, requestHeaders: Headers): ParseImageRequestURLResult {
+	const deviceSizes: number[] = __IMAGES_DEVICE_SIZES__;
+	const imageSizes: number[] = __IMAGES_IMAGE_SIZES__;
+	// const minimumCacheTTLSeconds = __IMAGES_MINIMUM_CACHE_TTL__;
```
vicb: What should we do with that?

packages/cloudflare/src/cli/templates/images.ts
pilcrowonpaper: Like passing `error`?

packages/cloudflare/src/cli/templates/images.ts
pilcrowonpaper: Yes this should be a 400

packages/cloudflare/src/cli/templates/images.ts
pilcrowonpaper: Removed

packages/cloudflare/src/cli/templates/images.ts
pilcrowonpaper: Updated to `ok` since that's why Next.js use

pilcrowonpaper:
It looks like we can't add tests for the `q` parameter and some behaviors of the `w` parameter because Cloudflare image transformation doesn't support `fit` and `quality` locally.

packages/cloudflare/src/cli/build/open-next/compile-images.ts
pilcrowonpaper: I updated the variable name so we don't have an awkward "is defined" variable

packages/cloudflare/src/cli/build/open-next/compile-images.ts
```diff
@@ -19,29 +19,53 @@ export async function compileImages(options: BuildOptions) {
 		: {};
 
 	const __IMAGES_REMOTE_PATTERNS__ = JSON.stringify(imagesManifest?.images?.remotePatterns ?? []);
+	const __IMAGES_ALLOW_ALL_LOCAL_PATHS__ = JSON.stringify(
```
vicb: Can't we completely remove that?

i.e. set `__IMAGES_LOCAL_PATTERNS__` to `[{pathname: '**'}]` when the local pattern is undefined?

packages/cloudflare/src/cli/templates/images.ts
vicb: This was marked as resolved without comment?

packages/cloudflare/src/cli/templates/images.ts
vicb: You commented that Next 16 do not limit the number of redirects.
In that case we should not do manual redirects but just follow redirects.
Do that make sense?

packages/cloudflare/src/cli/templates/images.ts
vicb: ??

packages/cloudflare/src/cli/templates/images.ts
vicb: Let's not do this in this PR to keep the current behavior.
We can decide to changer at a later point.

packages/cloudflare/src/cli/templates/images.ts
pilcrowonpaper: No, I meant the opposite. Next.js 16 does limit redirects, while it didn't in Next.js 15

packages/cloudflare/src/cli/build/open-next/compile-images.ts
pilcrowonpaper: Agree, my bad

packages/cloudflare/src/cli/templates/images.ts
pilcrowonpaper: I fixed it to use `.ok`

packages/cloudflare/src/cli/templates/images.ts
pilcrowonpaper: Don't have a strong opinion on this so updated it

packages/cloudflare/src/cli/build/open-next/compile-images.ts
pilcrowonpaper: Applied

packages/cloudflare/src/cli/templates/images.ts
pilcrowonpaper: IMO it's cleaner if the functions that we can about is at the top and it's "dependencies" (e.g. helper types and functions) are right below it. Don't have a strong opinion on this tho

packages/cloudflare/src/cli/templates/images.ts
pilcrowonpaper: Added a TODO

packages/cloudflare/src/cli/build/open-next/compile-images.ts
pilcrowonpaper: Added a comment

packages/cloudflare/src/cli/build/open-next/compile-images.ts
pilcrowonpaper: Added a comment on it

packages/cloudflare/src/cli/build/open-next/compile-images.ts
pilcrowonpaper: Set it to 3 to match with Next.js and added a comment on it

pilcrowonpaper:
@vicb I updated the endpoint to match the previous behavior and return the unoptimized image as is if the image format isn't supported by Cloudflare Images

vicb:
Rebased + fix tests

PR #1015, opened 2025-11-28
retrieve the preview ID from `process.env`

vicb:
The following was causing troubles in the memory queue:

```ts
			// TODO: Drop the import - https://github.com/opennextjs/opennextjs-cloudflare/issues/361
			// @ts-ignore
			const manifest = await import("./.next/prerender-manifest.json");
```

### DIFF
--- a/.changeset/honest-games-tell.md
+++ b/.changeset/honest-games-tell.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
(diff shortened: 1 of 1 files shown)


james-elicx (approved):

james-elicx:
probably want a changeset :)

vicb:
Thanks for the review and feedback James!

PR #1016, opened 2025-11-28
Update e2e tests to Next 16 (webpack)

vicb:
The tests would still fail for turbopack on Next 16 but this will be addressed in a follow up PR (we will test with both webpack and turbopack for Next 16)


vicb:
Thanks James!

PR #1017, opened 2025-12-01
fix: `response.validate` with the turbo pages runtime

vicb:
Fix SSG in the pages router e2e

I'll send a follow up PR to run e2e with turbopack

### DIFF
--- a/packages/cloudflare/src/cli/build/patches/plugins/res-revalidate.ts
+++ b/packages/cloudflare/src/cli/build/patches/plugins/res-revalidate.ts
@@ -14,12 +14,13 @@ rule:
     kind: if_statement
     stopBy: end
     has:
-      kind: parenthesized_expression
-      has: { kind: property_identifier , stopBy: end, regex: trustHostHeader}
+      field: condition
+      stopBy: end
+      pattern: $$$_.trustHostHeader
   has:
     kind: call_expression
     all:
-      - has: {kind: identifier, pattern: fetch}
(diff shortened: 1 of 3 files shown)


packages/cloudflare/src/cli/build/patches/plugins/res-revalidate.ts
```diff
@@ -14,12 +14,13 @@ rule:
     kind: if_statement
     stopBy: end
     has:
-      kind: parenthesized_expression
-      has: { kind: property_identifier , stopBy: end, regex: trustHostHeader}
+      field: condition
```
vicb, line 17: updates to the rule are for readability only

packages/cloudflare/src/cli/build/patches/plugins/res-revalidate.ts
```diff
@@ -69,7 +62,7 @@ export const patchResRevalidate: CodePatcher = {
 		{
 			versions: ">=14.2.0",
 			pathFilter: getCrossPlatformPathRegex(
-				String.raw`(pages-api\.runtime\.prod\.js|node/api-resolver\.js)$`,
+				String.raw`(pages-api(?:-turbo)?\.runtime\.prod\.js|node/api-resolver\.js)$`,
```
vicb, line 65: This is the actual change, applying the pacth to the turbo runtime

vicb:
Thanks Nico!

PR #1018, opened 2025-12-01
test: run e2e tests with both webpack and turbopack

vicb:
I am not super happy with the fact that we need a dedicated playwright config for turbopack because it builds the worker locally.

Any better idea?

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Build workers (webpack)
         run: pnpm -r build:worker
 
-      - name: Run playwright tests
+      - name: Run playwright tests (webpack)
         run: pnpm e2e
 
       - name: Build workers (turbopack)


james-elicx:
> I am not super happy with the fact that we need a dedicated playwright config for turbopack because it builds the worker locally.
> 
> Any better idea?

You could potentially use an environment variable to switch between the commands in the playwright config. The script to run the E2Es could set it before running the playwright command, something like `"e2e-turbopack": "USE_TURBOPACK=1 playwright test -c e2e/playwright.config.ts"`. And then the shared config switches based on that flag. I think that might work.

Not necessarily better, but would allow you to get rid of the second config file.

james-elicx (approved):

vicb:
> > I am not super happy with the fact that we need a dedicated playwright config for turbopack because it builds the worker locally.
> > Any better idea?
> 
> You could potentially use an environment variable to switch between the commands in the playwright config. The script to run the E2Es could set it before running the playwright command, something like `"e2e-turbopack": "USE_TURBOPACK=1 playwright test -c e2e/playwright.config.ts"`. And then the shared config switches based on that flag. I think that might work.
> 
> Not necessarily better, but would allow you to get rid of the second config file.

Thanks for the review and suggestion James!

Not really keen on the environment variable which is yet another way to pass parameters.

I'll merge as is for now and we can find better ways latter.

PR #1020, opened 2025-12-03
Add `WORKER_SELF_REFERENCE` service configuration to the `wrangler.jsonc` template

dario-piotrowicz:
Tested by running `pnpm opennextjs-cloudflare build` in a project without wrangler config file.

### DIFF
--- a/packages/cloudflare/src/cli/build/utils/create-config-files.ts
+++ b/packages/cloudflare/src/cli/build/utils/create-config-files.ts
@@ -40,7 +40,7 @@ export async function createWranglerConfigIfNotExistent(projectOpts: ProjectOpti
 
 	const appName = getAppNameFromPackageJson(projectOpts.sourceDir) ?? "app-name";
 	if (appName) {
-		wranglerConfig = wranglerConfig.replace('"app-name"', JSON.stringify(appName.replaceAll("_", "-")));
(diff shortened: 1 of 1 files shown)


packages/cloudflare/templates/wrangler.jsonc
```diff
@@ -8,6 +8,12 @@
 		"directory": ".open-next/assets",
 		"binding": "ASSETS"
 	},
+	"services": [
+		{
+			"binding": "WORKER_SELF_REFERENCE",
+			"service": "app-name"
```
vicb, line 14: q: isn't 'app-name' changed when the template is instantiated?

vicb: a: yes, [here](https://github.com/opennextjs/opennextjs-cloudflare/blob/b2feee09783b9239a91e538db326d764e62daf77/packages/cloudflare/src/cli/build/utils/create-config-files.ts#L43)

So I think the code will not work because it uses `replace` and not `replaceAll`

vicb: Fixed by adding a commit

dario-piotrowicz:
> (I added a fixup commit to replace the app name everywhere and tested that it works)

oh I see! I missed that, sorry 🙇  (and thanks 🙏)

PR #1023, opened 2025-12-03
When running `wrangler deploy` add a `OPEN_NEXT_DEPLOY` environment variable to let wrangler know that it is being run by open-next

dario-piotrowicz:
This is mostly beneficial just for `wrangler deploy` with autoconfig, since we're updating `wrangler deploy` to potentially call `opennextjs-cloudflare deploy`, the latter calls `wrangler deploy` so we need to have a mechanism in place for avoiding an infinite recursion in such cases

### DIFF
--- a/packages/cloudflare/src/cli/utils/run-wrangler.ts
+++ b/packages/cloudflare/src/cli/utils/run-wrangler.ts
@@ -80,11 +80,11 @@ export function runWrangler(options: BuildOptions, args: string[], wranglerOpts:
 			stdio: shouldPipeLogs ? ["ignore", "pipe", "pipe"] : "inherit",
 			env: {
 				...process.env,
+				...wranglerOpts.env,
 				// `.env` files are handled by the adapter.
 				// Wrangler would load `.env.<wrangler env>` while we should load `.env.<process.env.NEXTJS_ENV>`
 				// See https://opennext.js.org/cloudflare/howtos/env-vars
 				CLOUDFLARE_LOAD_DEV_VARS_FROM_DOT_ENV: "false",
-				...wranglerOpts.env,
 			},
 		}
 	);


packages/cloudflare/src/cli/utils/run-wrangler.ts
```diff
@@ -82,6 +82,8 @@ export function runWrangler(options: BuildOptions, args: string[], wranglerOpts:
 				// Wrangler would load `.env.<wrangler env>` while we should load `.env.<process.env.NEXTJS_ENV>`
 				// See https://opennext.js.org/cloudflare/howtos/env-vars
 				CLOUDFLARE_LOAD_DEV_VARS_FROM_DOT_ENV: "false",
+				// We pass this variable to wrangler so that it can be aware that it is being run by us
```
vicb: Let's discuss that more.
I am really not convinced that have code in wrangler that depends on Open Next (and vice versa) is the right thing to do as it introduce coupling and make harder to change things.

dario-piotrowicz: As we discussed let's proceed with this 🙂

packages/cloudflare/src/cli/utils/run-wrangler.ts
```diff
@@ -56,6 +56,8 @@ function injectPassthroughFlagForArgs(options: BuildOptions, args: string[]) {
 }
 
 export function runWrangler(options: BuildOptions, args: string[], wranglerOpts: WranglerOptions = {}) {
+	const isDeployCommand = args[0] === "deploy";
```
vicb: could you please reverse that and have the caller specify an (optional) extra env instead of trying to guess who called us here?

dario-piotrowicz: ~~Is this ok? https://github.com/opennextjs/opennextjs-cloudflare/pull/1023/commits/84e9cfad7e51fce91a3acf2878f37aaedb38ae8c~~

I misread your comment, let me update the code...

dario-piotrowicz: Done, please have another look 🙏

packages/cloudflare/src/cli/utils/run-wrangler.ts
```diff
@@ -13,6 +13,7 @@ type WranglerOptions = {
 	environment?: string;
 	configPath?: string;
 	logging?: "all" | "error";
+	env?: Record<string, unknown>;
```
vicb:
```suggestion
	env?: Record<string, string>;
```

packages/cloudflare/src/cli/utils/run-wrangler.ts
```diff
@@ -83,6 +84,7 @@ export function runWrangler(options: BuildOptions, args: string[], wranglerOpts:
 				// Wrangler would load `.env.<wrangler env>` while we should load `.env.<process.env.NEXTJS_ENV>`
 				// See https://opennext.js.org/cloudflare/howtos/env-vars
 				CLOUDFLARE_LOAD_DEV_VARS_FROM_DOT_ENV: "false",
+				...wranglerOpts.env,
 			},
```
vicb, line 88:
```suggestion
env: {
...process.env,
...wranglerOpts.env,
// `.env` files are handled by the adapter.
// Wrangler would load `.env.<wrangler env>` while we should load `.env.<process.env.NEXTJS_ENV>`
// See https://opennext.js.org/cloudflare/howtos/env-vars
CLOUDFLARE_LOAD_DEV_VARS_FROM_DOT_ENV: "false",				
},
```

vicb: 👆Sorry, not sure why but GH messes with this diff, the idea here to move `...wranglerOpts.env` one position up to disallow overriding `CLOUDFLARE_LOAD_DEV_VARS_FROM_DOT_ENV`

vicb (approved):
Thanks Dario!
LGTM with nits

packages/cloudflare/src/cli/utils/run-wrangler.ts
dario-piotrowicz: ah right, I see, good point! 👍

PR #1024, opened 2025-12-03
bump `@opennextjs/cloudflare` and react in the c3 template

vicb:
Tested that the versions are ok by updating the package.json of an app created with the template from before the PR


vicb:
Thanks James!

PR #1032, opened 2025-12-06
fix(images): allow any local path when no localPatterns are specified

vicb:
fixes #1029

`defaultLocalPatterns`:

- should be an array
- should contain regexp (not raw object as in the next config).

The regexp was generated using the image manifest generated by setting the local patterns to `[{ pathname: "/**" }]` in the next config.

/cc @pilcrowonpaper

PR #1033, opened 2025-12-06
add missing changeset for #1032

vicb:
Note: I'll fix the PR/commit ref by hand in the Version packages PR


vicb:
Thanks again Magnus :)

Issue #1030, opened 2025-12-05
[BUG] KV Population fails when warming the cache


roybarber:
### Describe the bug

Hey guys, is there a way i can debug whats going on here: Locally on my macbook i can deploy, on my mac pro i cant (get the same issue) and every standard deployment build fails in the cloudflare dashboard with the following:

13:50:15.573 | Populating KV incremental cache...
-- | --
13:50:15.584 | Inserting 200 assets to KV in chunks of 25
13:50:15.972 | ERROR Wrangler command failed
13:50:16.032 | 12%\|█████                                        \| 1/8 [00:00:<00:00:, 0.00it/s] ELIFECYCLE  Command failed with exit code 1.
13:50:16.061 | Failed: error occurred while running deploy command

Thats all, nothing is written to KV. Can we skip this? 

I've downgraded openNext to a older version as we have another replicated project thats running fine.

### Steps to reproduce

Deploy a ISR project using the recommended setup

My wrangler file:
````
{
    "$schema": "node_modules/wrangler/config-schema.json",
    "main": ".open-next/worker.js",
    "name": "uber-uat",
    "compatibility_date": "2025-11-17",
    "compatibility_flags": ["nodejs_compat", "global_fetch_strictly_public"],
    "account_id": "XXX",
    "minify": true,
    "assets": {
        "directory": ".open-next/assets",
        "binding": "ASSETS"
    },
    "observability": {
        "enabled": true
    },
    "placement": {
        "mode": "smart"
    },
    "preview_urls": false,
    "keep_vars": true,
    "durable_objects": {
        "bindings": [
            {
                "name": "NEXT_CACHE_DO_QUEUE",
                "class_name": "DOQueueHandler"
            },
            {
                "name": "NEXT_TAG_CACHE_DO_SHARDED",
                "class_name": "DOShardedTagCache"
            },
            {
                "name": "NEXT_CACHE_DO_PURGE",
                "class_name": "BucketCachePurge"
            }
        ]
    },
    "migrations": [
        {
            "tag": "v1",
            "new_sqlite_classes": ["DOQueueHandler", "DOShardedTagCache", "BucketCachePurge"]
        }
    ],
    "kv_namespaces": [
        {
            "binding": "NEXT_INC_CACHE_KV",
            "id": "XXX"
        }
    ],
    "services": [
        {
            "binding": "WORKER_SELF_REFERENCE",
            "service": "uber-uat"
        }
    ]
}
````

[...]

PR #1034, opened 2025-12-06
output more wrangler command logs on errors

james-elicx:
Intended to let people get the full logs from Wrangler instead of only stderr when an error happens

related to https://github.com/opennextjs/opennextjs-cloudflare/issues/1030#issuecomment-3618546024

image

### DIFF
--- a/.changeset/fresh-taxis-pay.md
+++ b/.changeset/fresh-taxis-pay.md
@@ -2,4 +2,4 @@
 "@opennextjs/cloudflare": patch
 ---
 
-output more wrangler command logs in debug mode
+output more wrangler command logs when the command fails


vicb:
Thanks James for looking into that 🙏

I haven't looked at the details of your changes yet but you say "relates to 1030". Do you have any idea why the user does not get error logs there?

One more thought: having more info with `OPEN_NEXT_DEBUG=1` is already an improvement. Do you think it would be possible/relevant to record all of the outputs of wrangler and print them when it exists with a non 0 status only (i.e. no need to use debug and no need to print extra info if wrangler succeeds)?

james-elicx:
> Thanks James for looking into that 🙏
> 
> I haven't looked at the details of your changes yet but you say "relates to 1030". Do you have any idea why the user does not get error logs there?

Without seeing the full log, it's hard to say. Presumably whatever error they're getting isn't being sent to stderr by Wrangler otherwise I'd expect to see it...

> One more thought: having more info with `OPEN_NEXT_DEBUG=1` is already an improvement. Do you think it would be possible/relevant to record all of the outputs of wrangler and print them when it exists with a non 0 status only (i.e. no need to use debug and no need to print extra info if wrangler succeeds)?

Makes sense, good idea

.changeset/fresh-taxis-pay.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+output more wrangler command logs in debug mode
```
vicb:
```suggestion
output more wrangler command logs when the command fails
```

james-elicx:
Thanks, sorry I forgot to update the changeset

PR #1038, opened 2025-12-09
Update the c3 template to use placeholders

vicb:
This PR update the c3 template to use the placeholders recently introduced in c3 (but not released yet).

The placeholders enabled to correctly setup the self binding name - it should match the worker name.

Notes:

- The current c3 template overrides files from this template with [those files](https://github.com/cloudflare/workers-sdk/tree/main/packages/create-cloudflare/templates/next/templates)
- The changes have been tested with the current released version of c3
- The changes have been tested with the future version of c3 (see [this PR](https://github.com/cloudflare/workers-sdk/pull/11575))
- There is also a [wrangler.jsonc template](https://github.com/opennextjs/opennextjs-cloudflare/blob/main/packages/cloudflare/templates/wrangler.jsonc) used when the project does not have one. It has been updated to use the same placeholder format and the corresponding code has been updated.

### DIFF
--- a/create-cloudflare/next/wrangler.jsonc
+++ b/create-cloudflare/next/wrangler.jsonc
@@ -12,9 +12,6 @@
 		"binding": "ASSETS",
 		"directory": ".open-next/assets"
 	},
-	"observability": {
-		"enabled": true
-	},
 	"images": {
 		// Enable image optimization
 		// see https://opennext.js.org/cloudflare/howtos/image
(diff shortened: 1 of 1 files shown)


vicb:
> LGTM although I am not sure how this can work with the current C3? Won't you end up with a project that has an invalid wrangler.jsonc where the name is `<WORKER_NAME>`?

No because the `wrangler.jsonc` in this PR is not used by the current c3. c3 has its own version of `wrangler.jsonc` - that's the first bullet point in the description (arguably [the link was wrong](https://github.com/cloudflare/workers-sdk/tree/main/packages/create-cloudflare/templates/next/templates) and I just fixed that).

And thanks for the review

PR #1039, opened 2025-12-09
switch back to `<TBD>` in the `package.json` of the c3 template

vicb:
`<PACKAGE_NAME>` is not supported with the current c3 and then not replaced


petebacondarwin (approved):

vicb:
Thanks Pete!

PR #1040, opened 2025-12-10
remove `.dev.vars.example` from the c3 template

vicb:
`.dev.vars` was added to the template (ignored in `.gitignore`) so we don't need the `.dev.vars.example` as well.


petebacondarwin (approved):

vicb:
Thanks Pete!

PR #1055, opened 2025-12-20
fix: add Next.js 16.1.0 compatibility for fast-set-immediate module

### LINKED ISSUE #1049: [BUG] Next.js 16.1.0 crashes on Cloudflare Workers (Error 1101): Cannot assign to read only property 'setImmediate' of [object Module]
### Describe the bug

After upgrading to Next.js 16.1.0, the Worker crashes on startup on Cloudflare Workers with Error 1101.

The exception is:
TypeError: Cannot assign to read only property 'setImmediate' of object '[object Module]'

The stack trace points to:
next/dist/server/node-environment-extensions/fast-set-immediate.external.js
and the app never serves the request.

### Steps to reproduce

1. Deploy a Next.js app to Cloudflare Workers using @opennextjs/cloudflare (OpenNext).
2. Ensure wrangler has nodejs_compat enabled.
3. Upgrade Next.js to 16.1.0.
4. Build & deploy.
5. Request `/` -> Worker throws TypeError and Cloudflare reports Error 1101.

Repro notes:
- Works with Next.js 16.0.x (downgrading avoids the crash). 

### Expected behavior

The Worker should start and serve requests normally (e.g. GET / returns the page) without throwing a startup exception.

1.14.6

### Wrangler version

4.56.0

### next info output

```shell
Operating System:
  Platform: linux
  Arch: x64
  Version: #37~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Nov 20 10:25:38 UTC 2
  Available memory (MB): 30716
  Available CPU cores: 16
Binaries:
  Node: 24.12.0
  npm: 11.7.0
  Yarn: 4.12.0
  pnpm: 10.24.0
Relevant Packages:
  next: 16.1.0 // Latest available version is detected (16.1.0).
  eslint-config-next: N/A
  react: 19.2.3
  react-dom: 19.2.3
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

### Additional context

_No response_

dulapahv:
Next.js 16.1.0 introduced a new internal module (`fast-set-immediate.external.js`) that attempts to assign to read-only module exports during the esbuild bundling process. This causes a startup crash on Cloudflare Workers with the error:

```
TypeError: Cannot assign to read only property 'setImmediate' of object '[object Module]'
```

Fixes #1049 

### Changes

- Created an esbuild plugin that intercepts the `fast-set-immediate.external` module and replaces it with a shim.

Verified by building the `playground16` example project with Next.js 16.1.0. Shimmed exports exist in the bundle at `.open-next/server-functions/default/examples/playground16/handler.mjs` and this error no longer occur.

### DIFF
--- a/.changeset/deep-breads-camp.md
+++ b/.changeset/deep-breads-camp.md
@@ -4,5 +4,4 @@
 
 fix: Add compatibility for Next.js 16.1.0 fast-set-immediate module
 
-Next.js 16.1.0 introduced a new internal module that causes build errors on Cloudflare Workers. This fix adds an
-esbuild plugin to shim the problematic module, allowing Next.js 16.1.0 apps to build and deploy successfully.
+Next.js 16.1.0 introduced an internal `fast-set-immediate.external` module that assigns to read-only module exports, causing a startup crash on Cloudflare Workers. This patch adds an esbuild shim to replace the module and restore compatibility.


vicb:
Great finding @dulapahv !
I think we need to understand why [Next does patch setImmediate](https://github.com/vercel/next.js/pull/86018) before merging this and if this PR will have side effects.

_Note that https://github.com/cloudflare/workerd/issues/5662 should help here too_

dulapahv:
Thanks for the feedback. From what I understand is that the `setImmediate` patch ([in this PR](https://github.com/vercel/next.js/pull/86018)) changes the usual behavior of the event loop by making `setImmediate` run immediately after the current task instead of waiting for the next event loop iteration.


Next.js's `fast-set-immediate.external.ts` runs this at module initialization:

```ts
function install() {
  if (process.env.NEXT_RUNTIME === 'edge') {
    // Nothing to patch. The exported functions all error if used in the edge runtime,
    // so we're not going to violate any assumptions by not patching.
    return
  } else {
    debug?.('installing fast setImmediate patch')

    const nodeTimers = require('node:timers') as typeof import('node:timers')

    // ⚠️ THIS LINE FAILS WHEN BUNDLED TO ESM
    globalThis.setImmediate = nodeTimers.setImmediate =
      // Workaround for missing __promisify__ which is not a real property
      patchedSetImmediate as unknown as typeof setImmediate
    globalThis.clearImmediate = nodeTimers.clearImmediate =
      patchedClearImmediate

    const nodeTimersPromises =
      require('node:timers/promises') as typeof import('node:timers/promises')
    nodeTimersPromises.setImmediate =
      patchedSetImmediatePromise as typeof import('node:timers/promises').setImmediate

    process.nextTick = patchedNextTick
  }
}

// Called at module load time
install()
```

The problem is this part:

```ts
globalThis.setImmediate = nodeTimers.setImmediate = patchedSetImmediate
```

This works in Node.js with CommonJS because `require('node:timers')` returns a mutable object. However, when esbuild bundles this to ESM, `node:timers` becomes an ES module namespace object with immutable exports. The assignment to `nodeTimers.setImmediate` throws:

```
TypeError: Cannot assign to read only property 'setImmediate' of object '[object Module]'
```

This happens at module initialization time when `install()` is called, so the Worker crashed before any request is handled.

I don't fully understand what Next.js relies on this patch for beyond the timing optimization described [in this PR](https://github.com/vercel/next.js/pull/86018). I can't confidently say there are no side effects from shimming this out.

inayayousfi:
Maybe until this PR is merged you could say it in the documentation. I'm sorry if I'm bothering but I just spent half my day trying to debug this just to find this PR.

vicb:
> Maybe until this PR is merged you could say it in the documentation. I'm sorry if I'm bothering but I just spent half my day trying to debug this just to find this PR.

The problem is that a lot of users don't read the docs https://opennext.js.org/cloudflare#supported-nextjs-versions ;)

inayayousfi:
> > Maybe until this PR is merged you could say it in the documentation. I'm sorry if I'm bothering but I just spent half my day trying to debug this just to find this PR.
> 
> The problem is that a lot of users don't read the docs https://opennext.js.org/cloudflare#supported-nextjs-versions ;)

Ok my bad but maybe you can put a link to the issue page as I think it could be very relevant for this specific project because Next.js is very easy to pinpoint what the known problem are?

dulapahv:
Given that the root cause needs to be addressed at the runtime layer and that #1057 implements a proper fix, this shim approach should probably not be merged. Happy to close this in favor of #1057.

vicb:
> Given that the root cause needs to be addressed at the runtime layer and that #1057 implements a proper fix, this shim approach should probably not be merged. Happy to close this in favor of #1057.

Yep I think this can be close.
There are still issues with the other PRs and Next 16.1 but I'll take a look tomorrow.
Thanks!

PR #1063, opened 2026-01-05
Fix r2 bulk put - respect bucket jurisdiction

kaziu687:
There is issue when you try to deploy to R2 bucket NEXT_INC_CACHE_R2_BUCKET with specified jurisdiction. This PR fixes this.


vicb (commented):
LGTM, thanks for the fix.

Would you mind adding a test to `packages/cloudflare/src/cli/commands/populate-cache.spec.ts`, should be easy enough.

kaziu687:
Already done: #1064

PR #1067, opened 2026-01-08
Add an example for importing `cloudflare:*` modules

vicb:


sommeeeer (approved):
LGTM, should this be in the docs somewhere as well?

vicb:
> LGTM, should this be in the docs somewhere as well?

I'm not sure.
It's better if nobody uses this to avoid code not working in `next dev`.
Maybe we can add that if we see a need later.

Thanks for the review!

vicb:
Thanks for the reviews!

PR #1069, opened 2026-01-16
check wrangler >= 4.59.2 to build Next 16.1+

vicb:
Fixes https://github.com/opennextjs/opennextjs-cloudflare/issues/1049

wrangler >= 4.59.2 uses a version of `workerd` that fixes `setImmediate`.

Tested locally

> [!warning]
> Next 16 is still not fully supported (as the warning warns you)
> That is the next thing we'll work on, should hopefully land in the coming days.
> _(The e2e tests do not pass with 16.1+ yet, that's the reason why Next is not updated in this PR)_

### DIFF
--- a/.changeset/young-regions-strive.md
+++ b/.changeset/young-regions-strive.md
@@ -1,0 +1,7 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+Check that wrangler is >= 4.59.2 when building Next 16.1+.
+
+To ensure `workerd` has a required fix to `setImmediate`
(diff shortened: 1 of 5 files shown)


sommeeeer (approved):
LGTM, however should this perhaps be an error and exit? If its not working with `workerd < 4.59.2`? 

Edit: I guess a warning is good enough for this case. If someone has already a workaround in place and can still use an affected workerd version.

vicb:
Thanks for the reviews!

PR #1071, opened 2026-01-16
Improve support for Next 16.1

vicb:
There was a bug in Next preventing 16.1+ to run well on Cloudflare.
The Next team fixed that in https://github.com/vercel/next.js/pull/88733 (released in 16.1.4)

We have a patch for 16.1.0-16.1.3.
The patch has been improved in aws and we should merge the next release (3.9.10) before merging this PR.


vicb:
Thanks for the review Magnus.
I'll merge and release if the CI gets green.
Then we only need to look at the build size before we can say 16.x is supported 🚀

PR #1072, opened 2026-01-16
fix(cloudflare): copy complete Next.js dist directory for Next.js 16+ support

buley:
When building Next.js 16+ apps for Cloudflare Workers, the build fails with 50+ module resolution errors like:
-  Could not resolve "./lib/incremental-cache"
-  Could not resolve "./lib/trace/tracer"
-  Could not resolve "./lib/router-utils/is-postpone"

# Root Cause

Next.js 16 changed the internal file structure. NFT (Node File Tracing) used by copyTracedFiles doesn't follow all internal require chains, resulting in an incomplete next/dist/ directory being copied to the output. Only ~5 files were being copied instead of the 176+ required.

# Solution

After copyTracedFiles runs, copy the complete next/dist/ directory from the actual workspace node_modules (not the standalone output) to ensure all internal modules are available. This is only done for Next.js 16+.

  Changes:
  - Added copyMissingNextDistDirs() function to createServerBundle.ts
  - Function runs only for Next.js 16+
  - Copies the complete next/dist/ directory from the source installation
  - Handles both standard and monorepo setups (bun, pnpm, npm)
  - Resolves symlinks correctly for bun/pnpm hoisted node_modules

  Testing:
  - Tested with Next.js 16.1.1 in a bun monorepo
  - Build completes successfully (previous 49+ module errors resolved)
  - Only unrelated project-specific errors remain

### DIFF
--- a/packages/cloudflare/src/cli/build/bundle-server.ts
+++ b/packages/cloudflare/src/cli/build/bundle-server.ts
@@ -1,5 +1,6 @@
+import { execSync } from "node:child_process";
 import fs from "node:fs";
-import { readFile, writeFile } from "node:fs/promises";
+import { readFile, stat, writeFile } from "node:fs/promises";
 import path from "node:path";
 import { fileURLToPath } from "node:url";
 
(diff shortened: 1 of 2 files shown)


vicb:
May I ask if the PR was AI generated?
To be clear I have no problem with using AI to generate PRs - I use it myself and it can help.

However AI sometimes is wrong about the root cause analysis and while what it generates "work", it is definitely not the right thing to do. It looks to me it's what happened here.

Anyway, could you please create an issue with a repro so that we can take a look at what the actual issue is?

PR #1076, opened 2026-01-20
fix: do not bundle og when not used

vicb:
Tested with playground16 in turbopack mode

Before:

image

After:

image


conico974 (approved):

vicb:
Thanks Nico!

PR #1079, opened 2026-01-21
fix: use the correct runtime (i.e. webpack or turbopack)

vicb:
See https://github.com/opennextjs/opennextjs-aws/pull/1087


examples/playground16/package.json
```diff
@@ -4,7 +4,7 @@
 	"private": true,
 	"scripts": {
 		"dev": "next dev --webpack",
```
dario-piotrowicz, line 6: just for consistency? (although it might not make much of a difference):
```suggestion
		"dev": "next dev",
```

vicb: Great suggestion.
I can actually do both as we do for other e2e tests

sommeeeer (approved):

vicb:
Thanks for the reviews ✅✅✅✅ 😮

PR #1083, opened 2026-01-21
Add `migrate` command for initializing open-next for existing Next.js projects

dario-piotrowicz:
Fixes https://jira.cfdata.org/browse/DEVX-2383

___

Docs PR: https://github.com/opennextjs/docs/pull/205

___

This PR addes a new `migrate` CLI command that adds the OpenNext adapter to standard Next.js projects (this basically automates the steps in https://opennext.js.org/cloudflare/get-started).

___

You can manually test this command by running:
```sh
npm create next-app  my-next-app
```
to create a new Next.js project, cd into the project:
```sh
cd my-next-app
```
install a compatible version of Next.js:
```sh
npm i next@16.0.x
```
and then run the `init` command:
```sh
npx https://pkg.pr.new/@opennextjs/cloudflare@1083 migrate
```
You should see something like the following:
Screenshot 2026-01-22 at 14 36 24

___

Thanks to @2u841r to spearheading this effort 🫶

### DIFF
--- a/packages/cloudflare/src/cli/commands/migrate.ts
+++ b/packages/cloudflare/src/cli/commands/migrate.ts
@@ -235,7 +235,7 @@ export function addMigrateCommand<T extends yargs.Argv>(y: T) {
 		"migrate",
 		"Set up the OpenNext Cloudflare adapter in an existing Next.js project",
 		(args) =>
-			args.option("force-install", {
+			args.option("forceInstall", {
 				type: "boolean",
 				alias: "f",
 				desc: "Install the dependencies using the `--force` flag.",


.changeset/add-init-command.md
```diff
@@ -0,0 +1,9 @@
+---
+"@opennextjs/cloudflare": minor
+---
+
+feature: add init command to set up OpenNext.js for Cloudflare
+
+This command helps users migrate existing Next.js applications to OpenNext.js for Cloudflare by automatically setting up all necessary configuration files, dependencies, and scripts. It provides an interactive package manager selection (`npm`, `pnpm`, `yarn`, `bun`, `deno`) with keyboard navigation and performs comprehensive setup including `wrangler.jsonc`, `open-next.config.ts`, `.dev.vars`, `package.json` scripts, Next.js config updates, and edge runtime detection.
```
vicb: I don't think we need more than that in the changeset. Details can go in the docs.

There was some discussion about the name init vs migrate too

```suggestion
This command helps users migrate existing Next.js applications to OpenNext.js for Cloudflare by automatically setting up all necessary configuration files, dependencies, and scripts.
```

pnpm-lock.yaml
vicb, line 1: I don't think this should be here given than no package.json is updated?

dario-piotrowicz: sorry, you're right, I committed this by mistake! 🙇

.changeset/add-init-command.md
dario-piotrowicz: yes, sorry this came from the first iteration of this and I forgot to check it 🙇

.changeset/add-init-command.md
```diff
@@ -0,0 +1,9 @@
+---
+"@opennextjs/cloudflare": minor
+---
+
+feature: add init command to set up OpenNext.js for Cloudflare
+
+This command helps users migrate existing Next.js applications to OpenNext.js for Cloudflare by automatically setting up all necessary configuration files, dependencies, and scripts.
```
vicb:
```suggestion
This command helps users migrate existing Next.js applications to the OpenNext Cloudflare adapter by automatically setting up all necessary configuration files, dependencies, and scripts.
```

.changeset/add-init-command.md
```diff
@@ -0,0 +1,9 @@
+---
+"@opennextjs/cloudflare": minor
+---
+
+feature: add init command to set up OpenNext.js for Cloudflare
```
vicb: Again, I think we need to think about the best name.

FWIW "init" makes me think create from scratch. @james-elicx suggested "migrate" yesterday, I like it better.

packages/cloudflare/src/cli/build/utils/create-config-files.ts
vicb, line 28: nit: should we support `"true"`, and `1` in top of `"yes "` and doc `"true"` in this message because I think it's more common

vicb: nit: move the detection to the same file used to create the file? (and drop `create-` prefix from the file name)

vicb: Same comment about moving the detection than for the wrangler config

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,238 @@
...
+	installDev: string;
+	run: string;
+}
+
+const packageManagers = {
+	pnpm: { name: "pnpm", install: "pnpm add", installDev: "pnpm add -D", run: "pnpm" },
+	npm: { name: "npm", install: "npm install", installDev: "npm install --save-dev", run: "npm run" },
+	bun: { name: "bun", install: "bun add", installDev: "bun add -D", run: "bun" },
+	yarn: { name: "yarn", install: "yarn add", installDev: "yarn add -D", run: "yarn" },
+} satisfies Record<string, PackageManager>;
```
vicb: nit: 
- type
- do we need the name prop (vs the key) ? I know some prefer normalized data, we can keep the prop if you prefer.

```suggestion
const packageManagers: Record<string, PackageManager> = {
	pnpm: { install: "pnpm add", installDev: "pnpm add -D", run: "pnpm" },
	npm: { install: "npm install", installDev: "npm install --save-dev", run: "npm run" },
	bun: { install: "bun add", installDev: "bun add -D", run: "bun" },
	yarn: { install: "yarn add", installDev: "yarn add -D", run: "yarn" },
};
```

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,238 @@
...
+		try {
+			let packageJson: { scripts?: Record<string, string> } = {};
+			if (fs.existsSync("package.json")) {
+				packageJson = JSON.parse(fs.readFileSync("package.json", "utf8")) as {
+					scripts?: Record<string, string>;
+				};
+			}
+
+			if (!packageJson.scripts) {
+				packageJson.scripts = {};
```
james-elicx: this feels a bit nuclear for pre-existing apps, can we merge instead of losing all the existing ones? And in the case of next-on-pages apps, remove the scripts that we had in the next-on-pages template/docs?

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,238 @@
...
+}
+
+const packageManagers = {
+	pnpm: { name: "pnpm", install: "pnpm add", installDev: "pnpm add -D", run: "pnpm" },
+	npm: { name: "npm", install: "npm install", installDev: "npm install --save-dev", run: "npm run" },
+	bun: { name: "bun", install: "bun add", installDev: "bun add -D", run: "bun" },
+	yarn: { name: "yarn", install: "yarn add", installDev: "yarn add -D", run: "yarn" },
+} satisfies Record<string, PackageManager>;
+
+function findFilesRecursive(dir: string, extensions: string[], fileList: string[] = []): string[] {
```
vicb: This does more than `findFilesRecursive`, could you please add JSDocs.

Those [utils in aws](https://github.com/opennextjs/opennextjs-aws/blob/4255fb71a2e513d1c69c16d4e8da396375dd4f6e/packages/open-next/src/build/helper.ts#L214) might help too

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,238 @@
...
+
+	return fileList;
+}
+
+/**
+ * Implementation of the `opennextjs-cloudflare init` command.
+ *
+ * @param args
+ */
+async function initCommand(): Promise<void> {
```
vicb: nit: move this to the top as the main entry point

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,238 @@
...
+	console.log("🚀 Setting up OpenNext.js for Cloudflare...\n");
```
vicb:
```suggestion
	console.log("🚀 Setting up the OpenNext Cloudflare adapter...\n");
```

packages/cloudflare/src/cli/commands/migrate.ts
```diff
@@ -0,0 +1,238 @@
...
+	runStep("Creating open-next.config.ts", async () => {
+		await createOpenNextConfig("./");
+	});
+
+	if (!fs.existsSync(".dev.vars")) {
+		runStep("Creating .dev.vars", () => {
+			const devVarsContent = `NEXTJS_ENV=development\n`;
+			fs.writeFileSync(".dev.vars", devVarsContent);
+		});
+	}
```
james-elicx: for ones that exist, should we update with the var in the event it's not already there?

dario-piotrowicz:
@vicb right now if an incorrect version of Next.js is used we get this:
Screenshot 2026-01-22 at 14 41 43

probably we should check the version and handle it more gracefully, right?

or shall we even try to bump the version? what do you think?

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,238 @@
...
+			fs.writeFileSync(".dev.vars", devVarsContent);
+		});
+	}
+
+	runStep("Creating _headers in public folder", () => {
+		if (!fs.existsSync("public")) {
+			fs.mkdirSync("public");
+		}
+		if (fs.existsSync("public/_headers")) {
+			console.log("⚠️  No public/_headers file already exists\n");
```
james-elicx:
```suggestion
			console.log("⚠️ public/_headers file already exists\n");
```

packages/cloudflare/src/cli/commands/init.ts
vicb: console -> import logger from "@opennextjs/aws/logger.js";

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,238 @@
...
+
+	if (fs.existsSync("open-next.config.ts")) {
+		console.log(
+			`Exiting since the project is already configured for OpenNext (an \`open-next.config.ts\` file already exists)\n`
+		);
+		return;
+	}
+
+	// Check if running on Windows
+	if (process.platform === "win32") {
```
vicb: re-use aws/packages/open-next/dist/build/utils.js

packages/cloudflare/src/cli/utils/wrangler-config.ts
```diff
@@ -0,0 +1,60 @@
...
+import { join } from "node:path";
+
+import { getPackageTemplatesDirPath } from "../../utils/get-package-templates-dir-path.js";
+
+/**
+ * Creates a wrangler.jsonc config file in the target directory for the project.
+ *
+ * @param projectDir The target directory for the project
+ */
+export async function createWranglerConfigFile(projectDir: string) {
```
james-elicx, line 32: should we merge with pre-existing if one exists already?

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	// Package manager selection
+	const { packager } = findPackagerAndRoot(".");
+	const packageManager = packageManagers[packager];
+
+	runStep("Installing dependencies", () => {
+		try {
+			execSync(`${packageManager.install} @opennextjs/cloudflare@latest`, { stdio: "inherit" });
+			execSync(`${packageManager.installDev} wrangler@latest`, { stdio: "inherit" });
```
vicb: I think this should not be dev (i.e. https://github.com/opennextjs/opennextjs-cloudflare/blob/e7c92eef068a0d9863c2687568bfb70fb0302813/packages/cloudflare/src/api/cloudflare-context.ts#L3)

But I also that's what we document/use everywhere :(

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,238 @@
...
+		} catch (error) {
+			console.error("❌ Failed to install dependencies:", (error as Error).message);
+			process.exit(1);
+		}
+	});
+
+	runStep("Creating wrangler.jsonc", async () => {
+		await createWranglerConfigFile("./");
```
vicb: I think we should check if files exist before creating new ones (here and below)

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,238 @@
...
+	});
+
+	runStep("Creating open-next.config.ts", async () => {
+		await createOpenNextConfig("./");
+	});
+
+	if (!fs.existsSync(".dev.vars")) {
+		runStep("Creating .dev.vars", () => {
+			const devVarsContent = `NEXTJS_ENV=development\n`;
+			fs.writeFileSync(".dev.vars", devVarsContent);
```
vicb:
```suggestion
			fs.writeFileSync(".dev.vars", `NEXTJS_ENV=development\n`);
```

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,238 @@
...
+		});
+	}
+
+	runStep("Creating _headers in public folder", () => {
+		if (!fs.existsSync("public")) {
```
vicb: Do we need to check if we use force?

packages/cloudflare/src/cli/commands/init.ts
vicb: No or Yes?
Also why logging for this one and not the others?

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,238 @@
...
+				packageJson = JSON.parse(fs.readFileSync("package.json", "utf8")) as {
+					scripts?: Record<string, string>;
+				};
+			}
+
+			if (!packageJson.scripts) {
+				packageJson.scripts = {};
+			}
+
+			packageJson.scripts.build = "next build";
```
vicb: nit:
`p.script = { ...p.script,
  preview: ...
  ...
}
`

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,238 @@
...
+			}
+		}
+
+		if (configFile) {
+			let configContent = fs.readFileSync(configFile, "utf8");
+			const importLine = 'import { initOpenNextCloudflareForDev } from "@opennextjs/cloudflare";';
+			const initLine = "initOpenNextCloudflareForDev();";
+
+			if (!configContent.includes(importLine)) {
+				// Add import at the top
```
vicb: not: Do we really need a comment here and on l174

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,238 @@
...
+			fs.writeFileSync(configFile, configContent);
+		} else {
+			console.log("⚠️  No Next.js config file found, you may need to create one\n");
+		}
+	});
+
+	runStep("Checking for edge runtime usage", () => {
+		try {
+			const extensions = [".ts", ".tsx", ".js", ".jsx", ".mjs"];
+			const files = findFilesRecursive(".", extensions).slice(0, 100); // Limit to first 100 files
```
vicb: nits:
- `.slice(0, 100)` would be better on l 190 (where it changes something)
- or use the util I mention above and stop finding files when you reach the max

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,238 @@
...
+			let foundEdgeRuntime = false;
+
+			for (const file of files) {
+				try {
+					const content = fs.readFileSync(file, "utf8");
```
vicb: nit: async who probably help a lot but I guess we do not care too much about this command perf.

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,238 @@
...
+						console.log(`⚠️  Found edge runtime in: ${file}`);
+						foundEdgeRuntime = true;
+					}
+				} catch {
+					// Skip files that can't be read
+				}
+			}
+
+			if (foundEdgeRuntime) {
+				console.log("\n🚨 WARNING:");
```
vicb: nit: use multine string?

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,238 @@
...
+
+	console.log("🎉 OpenNext.js for Cloudflare setup complete!");
+	console.log("\nNext steps:");
+	console.log(
+		`- Run: "${packageManager.run} preview" to build and preview your Cloudflare application locally`
+	);
+	console.log(`- Run: "${packageManager.run} deploy" to deploy your application to Cloudflare Workers`);
+}
+
+async function runStep(stepText: string, stepLogic: () => void | Promise<void>): Promise<void> {
```
vicb: nit: this function looks a bit overkill?
IIUC it only helps formatting the step title in a consistent way.
I would create a `printStepTitle(title)` instead (drop the callback)
Up to you

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,238 @@
...
+	await stepLogic();
+}
+
+/**
+ * Add the `init` command to yargs configuration.
+ */
+export function addInitCommand<T extends yargs.Argv>(y: T) {
+	return y.command(
+		"init",
+		"Set up OpenNext.js for Cloudflare in an existing Next.js project",
```
vicb: ...adapter

packages/cloudflare/src/cli/commands/init.ts
dario-piotrowicz: yes definitely! nice catch! 🙏  (PS: it's probably better to just ignore next-on-pages for now? here? 🤔 (since that migration also require users to manually check their app and rip out the edge runtime))

packages/cloudflare/src/cli/utils/create-open-next-config.ts
```diff
@@ -0,0 +1,18 @@
+import { cpSync } from "node:fs";
+import { join } from "node:path";
+
+import { getPackageTemplatesDirPath } from "../../utils/get-package-templates-dir-path.js";
+
+/**
+ * Creates a `open-next.config.ts` file in the target directory for the project.
+ *
+ * @param projectDir The target directory for the project
```
vicb:
```suggestion
 * @param appDir The Next application root
```

packages/cloudflare/src/cli/commands/migrate.ts
dario-piotrowicz: I think we can assume that there's not `.dev.vars` since this is not a Cloudflare / OpenNext project, no? 🤔 

(I guess it could be a next-on-pages one....?)

packages/cloudflare/src/cli/commands/init.ts
dario-piotrowicz: 🤦 👍

packages/cloudflare/src/cli/commands/init.ts
james-elicx: If they were using it for static pages without edge, not necessary ;p
Perhaps worthy of logging a warning or error if we spot next-on-pages in the package.json?

packages/cloudflare/src/cli/commands/migrate.ts
james-elicx: yeah my thinking was people migrating from next-on-pages

packages/cloudflare/src/cli/utils/wrangler-config.ts
dario-piotrowicz: as I mentioned above, it feels to me valid to assume that there's no wrangler config file, because if there was one then the application would already by an OpenNext one. Unless it was a next-on-pages one... in which case we'd need to convert the config from pages to workers?... mh.... we can do that, but as I asked above, maybe we should (at least for now) just ignore the next-on-pages case?

mybe one good thing to do would be to check, if the current project is a next-on-pages one we can bail and point the user to the docs? (saying that they need to manually migrate) 🤔

.changeset/add-init-command.md
dario-piotrowicz: I thought we agreed to go with init... but if `migrate` is the preference I'm ok with it 👍

packages/cloudflare/src/cli/build/utils/create-config-files.ts
dario-piotrowicz: shall we create a separate issue for this? (as it feels unrelated to this PR to me)

packages/cloudflare/src/cli/commands/init.ts
dario-piotrowicz: mh.... `satisfies` makes working with the object much nicer since the keys will be known, if we had `Record<string< PackageManager>` then things like `packageManagers['npm']` would be recognized as `PackageManager | undefined` forcing us to add type assertions and similar code... so I do personally feel that `satisfies` here help us having more robust and readable code

packages/cloudflare/src/cli/utils/wrangler-config.ts
james-elicx: I wonder if there are people who use wrnagler configs with a pure SSG Next.js app that might try run this to convert from SSG+static assets to OpenNext

packages/cloudflare/src/cli/commands/init.ts
dario-piotrowicz: can you clarify your comment?

packages/cloudflare/src/cli/commands/init.ts
dario-piotrowicz: The idea is that this function can be in the future improved for example by adding a loader that runs while the callback is executing

I'll drop the callback anyways and that can be re-added back when/if needed 👍

packages/cloudflare/src/cli/commands/init.ts
vicb: oops multiline, sorry

packages/cloudflare/src/cli/commands/init.ts
dario-piotrowicz: > No or Yes? 

Yes, I already fixed this but the GH diff still show the No... 🤦 

> Also why logging for this one and not the others?

Good point... let me check the various file creations and see what seems more appropriate...

packages/cloudflare/src/cli/commands/init.ts
dario-piotrowicz: no 👍

packages/cloudflare/src/cli/commands/init.ts
dario-piotrowicz: I've replaced this with `printHeaders` which seems to be the more appropriate way to go about this (it both prints the command and shows the windows warning), is this ok?

https://github.com/opennextjs/opennextjs-cloudflare/pull/1083/commits/a1f08a9a62dd52e68b2147c30a18e472e6dead97

packages/cloudflare/src/cli/commands/init.ts
dario-piotrowicz: Sorry for not noticing it right away, but `packageJson.scripts = {};` is behind `if (!packageJson.scripts) {` so there is no scripts deletion here, is there?

packages/cloudflare/src/cli/commands/init.ts
dario-piotrowicz: https://github.com/opennextjs/opennextjs-cloudflare/pull/1083/commits/55436a997fd0d4912866a6a887e4336f039bd6fa

packages/cloudflare/src/cli/commands/init.ts
dario-piotrowicz: mh... do you have a utility function to ignore indentation such as `dedent` in the repo? 🤔

packages/cloudflare/src/cli/build/utils/create-config-files.ts
dario-piotrowicz: Is this what you had in mind?
https://github.com/opennextjs/opennextjs-cloudflare/pull/1083/commits/9071bbf93876342355d9d5c600aae8eb5561ac60

dario-piotrowicz: does this work?

https://github.com/opennextjs/opennextjs-cloudflare/pull/1083/commits/4886b79e7bfd58413f661df960d138bb1911a688

packages/cloudflare/src/cli/commands/init.ts
vicb: No, just don't indent, i.e. https://github.com/opennextjs/opennextjs-cloudflare/blob/11bf0cc778c60f60700e1bfc0d4c0310661cb550/packages/cloudflare/src/cli/build/patches/plugins/load-manifest.ts#L58-L59

packages/cloudflare/src/cli/build/utils/create-config-files.ts
vicb: yes, thanks!

vicb: It works.

I usually add a "maybe" in the mix if it can return undefined or throw when there is no "maybe"

vicb: Also would be nice if the API is consistent for the different files

packages/cloudflare/src/cli/commands/init.ts
vicb: If it prints what you are looking for it's ok

packages/cloudflare/src/cli/commands/init.ts
dario-piotrowicz: what about this? https://github.com/opennextjs/opennextjs-cloudflare/pull/1083/commits/9987ba742ea5d0e8aa4fea9c28e12cb108c6b602

packages/cloudflare/src/cli/commands/init.ts
dario-piotrowicz: I don't think `force` helps here, `recursive` should do the trick 👍 

https://github.com/opennextjs/opennextjs-cloudflare/pull/1083/commits/8bc4f636f6794b6d36632ef1e7f92c6bd5d3c490

packages/cloudflare/src/cli/utils/wrangler-config.ts
dario-piotrowicz: mh....

> should we merge with pre-existing if one exists already?

By the way, this might be very tricky especially if the config is toml... I think we should just override it? 🤔 (maybe with a confirmation prompt)

packages/cloudflare/src/cli/commands/init.ts
james-elicx: oh man, now i feel dumb

packages/cloudflare/src/cli/utils/wrangler-config.ts
james-elicx: people still use toml? 😅

packages/cloudflare/src/cli/build/utils/create-config-files.ts
```diff
@@ -14,12 +12,7 @@ import { askConfirmation } from "../../utils/ask-confirmation.js";
  * @param projectOpts The options for the project
  */
 export async function createWranglerConfigIfNotExistent(projectOpts: ProjectOptions): Promise<void> {
```
vicb: nit: not existent -> non existent ? (gemini prefers non)

packages/cloudflare/src/cli/build/utils/create-config-files.ts
```diff
@@ -1,9 +1,7 @@
-import { cpSync, existsSync, readFileSync, writeFileSync } from "node:fs";
-import { join } from "node:path";
-
-import { getPackageTemplatesDirPath } from "../../../utils/get-package-templates-dir-path.js";
 import type { ProjectOptions } from "../../project-options.js";
 import { askConfirmation } from "../../utils/ask-confirmation.js";
+import { createOpenNextConfig, getOpenNextConfigPath } from "../../utils/open-next-config.js";
+import { createWranglerConfigFile, wranglerConfigFileExists } from "../../utils/wrangler-config.js";
```
vicb: would be nice to have a similar API for both?

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,228 @@
...
+ * Implementation of the `opennextjs-cloudflare init` command.
+ *
+ * @param args
+ */
+async function initCommand(): Promise<void> {
+	printHeaders("init");
+
+	logger.info("🚀 Setting up the OpenNext Cloudflare adapter...\n");
+
+	if (fs.existsSync("open-next.config.ts")) {
```
vicb: unify with https://github.com/opennextjs/opennextjs-cloudflare/pull/1083/changes#diff-973582590561094d3a783ec7627ff7d932998f7b4c2dcb701c4651b5ccba1bc2R44?

packages/cloudflare/src/cli/commands/migrate.ts
```diff
@@ -0,0 +1,228 @@
...
+
+	// Package manager selection
+	const { packager } = findPackagerAndRoot(".");
+	const packageManager = packageManagers[packager];
+
+	printStepTitle("Installing dependencies");
+	try {
+		execSync(`${packageManager.install} @opennextjs/cloudflare@latest`, { stdio: "inherit" });
+		execSync(`${packageManager.installDev} wrangler@latest`, { stdio: "inherit" });
```
vicb: Could you create an issue to move that to non dev (please mention open next docs, cloudflare doc, template, examples, ... in the issue)
Thx

packages/cloudflare/src/cli/commands/migrate.ts
```diff
@@ -0,0 +1,228 @@
...
+	} catch (error) {
+		logger.error("❌ Failed to install dependencies:", (error as Error).message);
+		process.exit(1);
+	}
+
+	printStepTitle("Creating wrangler.jsonc");
+	await createWranglerConfigFile("./");
```
vicb, line 66: What if the file exists?
And I think the JSDoc should answer this question.

unify createWranglerConfigFile vs createOpenNextConfig

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,228 @@
...
+	await createOpenNextConfig("./");
+
+	if (!fs.existsSync(".dev.vars")) {
+		printStepTitle("Creating .dev.vars");
+		fs.writeFileSync(".dev.vars", `NEXTJS_ENV=development\n`);
+	}
+
+	printStepTitle("Creating _headers in public folder");
+	if (fs.existsSync("public/_headers")) {
+		logger.warn("⚠️ public/_headers file already exists\n");
```
vicb: we do not warn for other, maybe add a comment about why we should here?

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,228 @@
...
+			build: "next build",
+			preview: "opennextjs-cloudflare build && opennextjs-cloudflare preview",
+			deploy: "opennextjs-cloudflare build && opennextjs-cloudflare deploy",
+			upload: "opennextjs-cloudflare build && opennextjs-cloudflare upload",
+			["cf-typegen"]: "wrangler types --env-interface CloudflareEnv cloudflare-env.d.ts",
+		};
+
+		fs.writeFileSync("package.json", JSON.stringify(packageJson, null, 2));
+	} catch (error) {
+		logger.error("❌ Failed to update package.json:", (error as Error).message);
```
vicb: Same comment as for warn before.
Why here and not i.e. for .dev.vars?

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,228 @@
...
+
+	printStepTitle(`${gitIgnoreExists ? "Updating" : "Creating"} .gitignore file`);
+	const gitIgnoreOpenNextText = "# OpenNext\n.open-next\n";
+
+	if (!gitIgnoreExists) {
+		fs.writeFileSync(".gitignore", gitIgnoreOpenNextText);
+	} else {
+		const gitignoreContent = fs.readFileSync(".gitignore", "utf8");
+		if (!gitignoreContent.includes(".open-next")) {
+			fs.writeFileSync(".gitignore", `${gitignoreContent}\n${gitIgnoreOpenNextText}`);
```
vicb: https://nodejs.org/api/fs.html#fsappendfilesyncpath-data-options

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,228 @@
...
+		}
+	}
+
+	printStepTitle("Updating Next.js config");
+	const configFiles = ["next.config.ts", "next.config.js", "next.config.mjs"];
```
vicb: See https://github.com/opennextjs/opennextjs-aws/blob/4255fb71a2e513d1c69c16d4e8da396375dd4f6e/packages/open-next/src/build/patch/patches/patchOriginalNextConfig.ts#L80 for a list of possible extensions.

We should handle CJS differently

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,228 @@
...
+		}
+
+		const initLine = "initOpenNextCloudflareForDev();";
+		if (!configContent.includes(initLine)) {
+			configContent += "\n" + initLine + "\n";
+		}
+
+		fs.writeFileSync(configFile, configContent);
+	} else {
+		logger.warn("⚠️  No Next.js config file found, you may need to create one\n");
```
vicb: Should we gate this whole command on this instead?

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,228 @@
...
+	}
+
+	printStepTitle("Checking for edge runtime usage");
+	try {
+		const extensions = [".ts", ".tsx", ".js", ".jsx", ".mjs"];
```
vicb: +mts
Do we need to handle CJS variants too?

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,228 @@
...
+		const files = findFilesRecursive(".", extensions).slice(0, 100); // Limit to first 100 files
```
vicb: I really don't like that `slice` here.

Go find me 10k files but only keep 100 is much better as go find me up to 100 files. Add a `maxNumFiles` param to `findFilesRecursive`?

packages/cloudflare/src/cli/commands/migrate.ts
```diff
@@ -0,0 +1,228 @@
...
+					logger.warn(`⚠️  Found edge runtime in: ${file}`);
+					foundEdgeRuntime = true;
+				}
+			} catch {
+				// Skip files that can't be read
+			}
+		}
+
+		if (foundEdgeRuntime) {
+			logger.warn(
```
vicb, line 157: nit: doesn't `logger.warn` has a `WARN` prefix yet

packages/cloudflare/src/cli/commands/init.ts
```diff
@@ -0,0 +1,228 @@
...
+		if (foundEdgeRuntime) {
+			logger.warn(
+				"\n🚨 WARNING:\n" +
+					'Remove any export const runtime = "edge"; if present\n' +
+					'Before deploying your app, remove the export const runtime = "edge"; line from any of your source files.\n' +
+					"The edge runtime is not supported yet with @opennextjs/cloudflare.\n"
+			);
+		}
+	} catch {
+		logger.warn("⚠️  Could not check for edge runtime usage\n");
```
vicb: If it's not actionable, don't print it?

packages/cloudflare/src/cli/commands/init.ts
vicb: Looks like JS from 10y ago but I won't oppose

vicb:
Could you please merge master in this PR instead of rebasing?
I realize that I have been commenting the same thing multiple time because it doesn't show up in the diff view.
GH is a pain here :(

packages/cloudflare/src/cli/utils/wrangler-config.ts
dario-piotrowicz: yes, I know of people that do! 🙈 

mh... we could check the type of file and handle json(c) cases and throw for tomls... 🤔

packages/cloudflare/src/cli/build/utils/create-config-files.ts
dario-piotrowicz: > I usually add a "maybe" in the mix if it can return undefined or throw when there is no "maybe"

The return type make it clear to me so adding a `maybe` feels unnecessary to me, but I'm happy to add that since it's your preference 👍 

> Also would be nice if the API is consistent for the different files

I thought the same when implementing the two solutions... but having a return from a function that is never actually used feels ugly/potentially confusing to me, but it's not a huge deal, I'm totally happy to use a consistent API 👍

packages/cloudflare/src/cli/commands/init.ts
dario-piotrowicz: I could see the header being nicely printed 👍 , I can't manually check the Windows warning since I don't have a windows machine, but that should definitely be there on windows systems as well

packages/cloudflare/src/cli/commands/init.ts
dario-piotrowicz: > Looks like JS from 10y ago but I won't oppose

Thanks! 😄 

___

> No, just don't indent

I really really don't like breaking indentation like that! I think that it makes code so much uglier and hard to read 😬

packages/cloudflare/src/cli/commands/init.ts
dario-piotrowicz: why `instead`? we can gate the command behind more than one thing, what about gating it to the open-next config (like now) and also the next config file?

packages/cloudflare/src/cli/commands/init.ts
dario-piotrowicz: mh... I am actually not sure why we're limiting this to 100... (that comes from the first iteration of this command) what about just leave this unbounded? is it too risky?

packages/cloudflare/src/cli/commands/migrate.ts
dario-piotrowicz: yes, good call, thanks 👍

packages/cloudflare/src/cli/commands/init.ts
dario-piotrowicz: I think that it is actionable, let me improve the message 👍

dario-piotrowicz: https://github.com/opennextjs/opennextjs-cloudflare/pull/1083/commits/b1ebb730520e44045b0b3509a2ac51970f5f7b61

packages/cloudflare/src/cli/commands/init.ts
dario-piotrowicz: https://github.com/opennextjs/opennextjs-cloudflare/pull/1083/commits/0d8a694e4e27c19afad0756483370c89bfb0c499

packages/cloudflare/src/cli/commands/init.ts
dario-piotrowicz: > +mts

done 👍  (6b47b3b3c6509e)

> Do we need to handle CJS variants too?

Route segment configs seem to only be ESM: https://nextjs.org/docs/13/app/api-reference/file-conventions/route-segment-config so I don't think there's a CJS variant that we need to take care of

packages/cloudflare/src/cli/commands/init.ts
dario-piotrowicz: > We should handle CJS differently

I think we should just do this:
https://github.com/cloudflare/workers-sdk/blob/bbd8a5e98cbe3048d80652ecf74368b9c26bd2ff/packages/wrangler/src/autoconfig/frameworks/next.ts#L128

as this works both for cjs and esm (so there wouldn't be a need to check and handle the two cases differently)

packages/cloudflare/src/cli/commands/migrate.ts
dario-piotrowicz: mh... I was about to create one, but can you explain why you think `wrangler` should be a standard dependency?

Is it just because of https://github.com/opennextjs/opennextjs-cloudflare/pull/1083#discussion_r2717242928?

But if that's the case, wrangler being a dev dependency of the application does seem correct to me

packages/cloudflare/src/cli/build/utils/create-config-files.ts
dario-piotrowicz: https://github.com/opennextjs/opennextjs-cloudflare/pull/1083/commits/197d561e7771ac2a33df3526d97119978bfca950

packages/cloudflare/src/cli/utils/wrangler-config.ts
vicb: I would keep things simple for the first iteration and actually always keep it simple unless there are numerous demands (we could info a link to the doc)

packages/cloudflare/src/cli/commands/init.ts
vicb: [Those people](https://google.github.io/styleguide/jsguide.html#features-strings-template-strings) know something about CS

vicb: ...or at least they managed to wash my brain

packages/cloudflare/src/cli/commands/init.ts
dario-piotrowicz: https://github.com/opennextjs/opennextjs-cloudflare/pull/1083/commits/802110c1d276fb633f132ce16fdf24783bba4b27

packages/cloudflare/src/cli/commands/init.ts
dario-piotrowicz: yeah, I've also removed the 100 limit, so this being async (and parallel) could consume lots of memory, so maybe it is better to keep it as is (and change things later if this ends up being slow?)

packages/cloudflare/src/cli/commands/init.ts
dario-piotrowicz: [2aba7fc](https://github.com/opennextjs/opennextjs-cloudflare/pull/1083/commits/2aba7fc3d8686dc0ef4be5de5ee006de7e2ae8ee)

packages/cloudflare/src/cli/commands/migrate.ts
dario-piotrowicz: > unify createWranglerConfigFile vs createOpenNextConfig

Can you clarify what you mean? 🙏

packages/cloudflare/src/cli/commands/init.ts
dario-piotrowicz: https://github.com/opennextjs/opennextjs-cloudflare/pull/1083/commits/dce268140010ff6a1868c21d5578d7be0fb887f3

packages/cloudflare/src/cli/utils/wrangler-config.ts
dario-piotrowicz: I've implemented this: https://github.com/opennextjs/opennextjs-cloudflare/pull/1083/commits/908c4b8b610feee0ff3f35ba21b7d539c5bbb48c to keep things simple right now, we can always improve on this later, what do you think? 🙂

packages/cloudflare/src/cli/build/utils/create-config-files.ts
dario-piotrowicz: done 👍

packages/cloudflare/src/cli/commands/migrate.ts
vicb: The names Config vs ConfigFile

packages/cloudflare/src/cli/commands/migrate.ts
dario-piotrowicz: updated 🫡

.changeset/add-init-command.md
dario-piotrowicz: updated 🫡

packages/cloudflare/src/cli/commands/migrate.ts
dario-piotrowicz: got it! thanks 🙂

dario-piotrowicz: ah, right, updated 😄

packages/cloudflare/src/cli/commands/init.ts
dario-piotrowicz: I've removed it and updated the logic here to be consistent with the creation/update of other files 👍

packages/cloudflare/src/cli/commands/migrate.ts
vicb: and https://github.com/opennextjs/opennextjs-cloudflare/blob/11bf0cc778c60f60700e1bfc0d4c0310661cb550/packages/cloudflare/src/cli/commands/utils.ts#L10

... anytime we import from wrangler in the code

packages/cloudflare/src/cli/build/utils/create-config-files.ts
vicb: > The return type make it clear to me so adding a maybe feels unnecessary to me, but I'm happy to add that since it's your preference 👍

When you review code on GH, you don't get to see the return type easily

> I thought the same when implementing the two solutions... but having a return from a function that is never actually used feels ugly/potentially confusing to me, but it's not a huge deal, I'm totally happy to use a consistent API 👍

I need to double check the code here, I thought it was used

.changeset/add-init-command.md
```diff
@@ -0,0 +1,9 @@
+---
+"@opennextjs/cloudflare": minor
+---
+
+feature: add `migrate` command to set up OpenNext.js for Cloudflare
```
vicb: ... aadpter

packages/cloudflare/src/cli/build/utils/create-config-files.ts
```diff
@@ -13,12 +11,8 @@ import { askConfirmation } from "../../utils/ask-confirmation.js";
...
  * @param projectOpts The options for the project
  */
-export async function createWranglerConfigIfNotExistent(projectOpts: ProjectOptions): Promise<void> {
-	const possibleExts = ["toml", "json", "jsonc"];
-
-	const wranglerConfigFileExists = possibleExts.some((ext) =>
-		existsSync(join(projectOpts.sourceDir, `wrangler.${ext}`))
-	);
+export async function createWranglerConfigIfNonExistent(projectOpts: ProjectOptions): Promise<void> {
+	const wranglerConfigFileExists = !!getWranglerConfigPath(projectOpts.sourceDir);
```
vicb:
```suggestion
	const wranglerConfigFileExists = Boolean(getWranglerConfigPath(projectOpts.sourceDir));
```

packages/cloudflare/src/cli/utils/wrangler-config.ts
```diff
@@ -0,0 +1,79 @@
...
+		if (existsSync(path)) {
+			return path;
+		}
+	}
+
+	return undefined;
+}
+
+/**
+ * Creates a wrangler.jsonc config file in the target directory for the project.
```
vicb, line 26: Add a note about what happens if the file already exists?

packages/cloudflare/src/cli/build/utils/files.ts
```diff
@@ -0,0 +1,26 @@
+import fs from "node:fs";
+
+/**
+ * Creates a file with the given text, or appends to it if it already exists and the condition is met.
+ *
+ * @param filepath The path to the file to create or append to.
+ * @param text The text to write or append to the file.
+ * @param condition A function that receives the current file content and returns `true` if the text should be appended.
+ */
+export function createOrAppendToFile(
```
vicb: which I think is appendToFile

> The fs.appendFileSync() method is used to synchronously append the given data to a file. A new file is created if it does not exist

packages/cloudflare/src/cli/commands/migrate.ts
```diff
@@ -0,0 +1,249 @@
...
+		logger.error(
+			`The project already contains a Wrangler config file (at ${wranglerConfigFilePath}).\n` +
+				"This means that your application is either a static site or a next-on-pages project.\n" +
+				"If your project is a static site and you want to migrate to OpenNext, delete the Wrangler configuration file, convert the project to a full stack one and try again." +
+				" if your project is a next-on-pages one remove any next-on-pages configuration, any edge runtime usage and try again."
+		);
+		process.exit(1);
+	}
+
+	if (getOpenNextConfigPath(projectDir)) {
```
vicb: IMO "getPath" should not check that the file exists

packages/cloudflare/src/cli/commands/migrate.ts
```diff
@@ -0,0 +1,249 @@
...
+
+	printStepTitle("Creating open-next.config.ts");
+	await createOpenNextConfigFile("./");
+
+	const devVarsExists = fs.existsSync(".dev.vars");
+	printStepTitle(`${devVarsExists ? "Updating" : "Creating"} .dev.vars file`);
+	createOrAppendToFile(
+		".dev.vars",
+		"NEXTJS_ENV=development",
+		(gitIgnoreContent) => !gitIgnoreContent.includes("NEXTJS_ENV=")
```
vicb: `gitIgnoreContent` -> `content`

It could be `NEXTJS_ENV =...` so maybe only check for `NEXTJS_ENV`

packages/cloudflare/src/cli/commands/migrate.ts
```diff
@@ -0,0 +1,249 @@
...
+	);
+
+	if (!fs.existsSync(".dev.vars")) {
+		printStepTitle("Creating .dev.vars");
+		fs.writeFileSync(".dev.vars", "NEXTJS_ENV=development\n");
+	}
```
vicb: Duplicate from right above
```suggestion
```

packages/cloudflare/src/cli/commands/migrate.ts
```diff
@@ -0,0 +1,249 @@
...
+	createOrAppendToFile(
+		"public/_headers",
+		[
+			"",
+			"# https://developers.cloudflare.com/workers/static-assets/headers",
+			"# https://opennext.js.org/cloudflare/caching#static-assets-caching",
+			"/_next/static/*",
+			"  Cache-Control: public,max-age=31536000,immutable",
+			"",
+		].join("\n"),
```
vicb: why `join` instead of a string literal ?

packages/cloudflare/src/cli/commands/migrate.ts
```diff
@@ -0,0 +1,249 @@
...
+		let packageJson: { scripts?: Record<string, string> } = {};
+		if (fs.existsSync("package.json")) {
+			packageJson = JSON.parse(fs.readFileSync("package.json", "utf8")) as {
+				scripts?: Record<string, string>;
+			};
+		}
+
+		packageJson.scripts = {
+			...packageJson.scripts,
+			build: "next build",
```
vicb: build should be there already... at least we should not override it
```suggestion
			build: "next build",
			...packageJson.scripts,
```

packages/cloudflare/src/cli/commands/migrate.ts
```diff
@@ -0,0 +1,249 @@
...
+		logger.warn(
+			"Failed to check for edge runtime usage.\n" +
+				"The edge runtime is not supported yet with @opennextjs/cloudflare.\n" +
+				'If present, remove all the `export const runtime = "edge";` lines from your source files'
+		);
+	}
+
+	logger.info("🎉 OpenNext.js for Cloudflare setup complete!");
+	logger.info("\nNext steps:");
+	logger.info(
```
vicb: why not logger.info(multi line string) instead of calling logger.info 4 times ?

packages/cloudflare/src/cli/commands/migrate.ts
dario-piotrowicz: mh... that's in the adapter's source code, I am not sure how that relates to applications

Applications won't be importing from `wrangler` directly, would they? (and especially not for builds)

packages/cloudflare/src/cli/build/utils/files.ts
dario-piotrowicz: oops, good callout! 👍

packages/cloudflare/src/cli/commands/migrate.ts
dario-piotrowicz: removed: https://github.com/opennextjs/opennextjs-cloudflare/pull/1083/commits/5d522f598f743420b05ce0a717d3ff468e84c6ad

whoever the return type/behavior is now inconsistent with `getWranglerConfigPath`:
https://github.com/opennextjs/opennextjs-cloudflare/blob/5d522f598f743420b05ce0a717d3ff468e84c6ad/packages/cloudflare/src/cli/utils/wrangler-config.ts#L6-L23

let me know if you're ok with the inconsistency

vicb: It might be that `getWranglerConfigPath` has this name because I asked names to be consistent.

BUT this function does not "get wrangler config path" so either the name should be changed or the behavior should be changed.

Compare to:

https://github.com/opennextjs/opennextjs-cloudflare/blob/5d522f598f743420b05ce0a717d3ff468e84c6ad/packages/cloudflare/src/cli/utils/open-next-config.ts#L6-L14

There `getOpenNextConfigPath` does what the name says.
_(BUT The JSDoc is wrong and needs fixing)_

packages/cloudflare/src/cli/build/utils/create-config-files.ts
```diff
@@ -95,8 +44,7 @@ export async function getLatestCompatDate(): Promise<string | undefined> {
  * @return The path to the created source file
  */
 export async function createOpenNextConfigIfNotExistent(sourceDir: string): Promise<string> {
-	const openNextConfigPath = join(sourceDir, "open-next.config.ts");
-
+	const openNextConfigPath = getOpenNextConfigPath(sourceDir);
 	if (!existsSync(openNextConfigPath)) {
```
vicb: Related to the discussion, why do we `existsSync` here while 
https://github.com/opennextjs/opennextjs-cloudflare/pull/1083/changes#diff-973582590561094d3a783ec7627ff7d932998f7b4c2dcb701c4651b5ccba1bc2R17

That's what I mean by consistency

packages/cloudflare/src/cli/build/utils/files.ts
```diff
@@ -0,0 +1,22 @@
+import fs from "node:fs";
+
+/**
+ * Runs fs' `appendFileSync` on a target file, but only a specified condition is met in regards to the file's content.
```
vicb:
```suggestion
 * Appends text to a file
 * 
 * When the file does not exists, it is always created with the text content.
 * When the file exists, the text is appended only when the predicate return `true`.  
```

packages/cloudflare/src/cli/build/utils/files.ts
```diff
@@ -0,0 +1,22 @@
...
+	text: string,
+	condition: (fileContent: string) => boolean
+): void {
+	const fileExists = fs.existsSync(filepath);
+
+	const fileContent = fileExists ? fs.readFileSync(filepath, "utf8") : undefined;
+
+	if (fileContent !== undefined && condition(fileContent)) {
+		fs.appendFileSync(filepath, `\n${text}\n`);
+	}
```
vicb: nit: simplify ?
```suggestion
	if (!fileExists || condition(fs.readFileSync(filepath, "utf8"))) {
		fs.appendFileSync(filepath, `\n${text}\n`);
	}
```

packages/cloudflare/src/cli/build/utils/files.ts
```diff
@@ -0,0 +1,22 @@
...
+	filepath: string,
+	text: string,
+	condition: (fileContent: string) => boolean
+): void {
+	const fileExists = fs.existsSync(filepath);
+
+	const fileContent = fileExists ? fs.readFileSync(filepath, "utf8") : undefined;
+
+	if (fileContent !== undefined && condition(fileContent)) {
+		fs.appendFileSync(filepath, `\n${text}\n`);
```
vicb: If wer really need to add leading and trailing newlines, it should be called out in the JSDoc IMO.
I would just use the unmodified text and move them to the call site when needed (principle of least surprise)

packages/cloudflare/src/cli/commands/migrate.ts
```diff
@@ -0,0 +1,240 @@
+import { execSync } from "node:child_process";
```
vicb: consistency
```suggestion
import childProcess from "node:child_process";
```

packages/cloudflare/src/cli/commands/migrate.ts
vicb: It's a peer dep of the adapter

https://github.com/opennextjs/opennextjs-cloudflare/blob/397f035b5d937abb6d58a2c00fc4aff6cb8b83af/packages/cloudflare/package.json#L87-L89

Are those always installed by all packages manager?
Probably a question for some AI

packages/cloudflare/src/cli/commands/migrate.ts
```diff
@@ -0,0 +1,240 @@
...
+	yarn: { name: "yarn", install: "yarn add", installDev: "yarn add -D", run: "yarn" },
+} satisfies Record<string, PackageManager>;
+
+/**
+ * Recursively searches a directory for files with specified extensions.
+ *
+ * Skips common build/cache directories: node_modules, .next, .open-next, .git, dist, build.
+ *
+ * @param dir - The directory path to start searching from
+ * @param extensions - Array of file extensions to match (e.g., ['.ts', '.js'])
```
```suggestion
 * @param extensions - Array of file extensions to match
```

packages/cloudflare/src/cli/commands/migrate.ts
vicb: OMG I seem to remember that Imade you rename `getOpenNextConfigPath` from `getOpenNextConfig`.

Sorry that was a bad idea.

Maybe `findNextConfig` would be a good name and keep the current API (undefined/path)?

packages/cloudflare/src/cli/commands/migrate.ts
```diff
@@ -0,0 +1,240 @@
...
+ * @param args
+ */
+async function migrateCommand(): Promise<void> {
+	printHeaders("migrate");
+
+	logger.info("🚀 Setting up the OpenNext Cloudflare adapter...\n");
+
+	const projectDir = process.cwd();
+
+	const nextConfigFilePath = getNextConfigPath(projectDir);
```
vicb: Reuse [`checkRunningInsideNextjsApp`](https://github.com/opennextjs/opennextjs-aws/blob/4255fb71a2e513d1c69c16d4e8da396375dd4f6e/packages/open-next/src/build/helper.ts#L398) ?

packages/cloudflare/src/cli/commands/migrate.ts
dario-piotrowicz: sorry, I am not sure what it being a peer dependency of the adapter has to do with what we're discussing

the question here is: should `wrangler` be a dev dependency or a standard dependency of the target **application**? (not the adapter itself)

and the answer for me is a clear yes, since the application doesn't get shipped to production with wrangler (wrangler can be used by the application only during development)

vicb: That's not clear to me.

If CI/CD do not install the dev deps, they will not install wrangler and will not be able to build the app

vicb: But it's most probably ok, we would have had complaints already.

packages/cloudflare/src/cli/commands/migrate.ts
dario-piotrowicz: We still need the config path in the rest of the function so either way I'd have to include logic that finds the path 😕 

We can reuse logic here but that would require changes in the aws package, can that be done as a followup?

dario-piotrowicz: I think we can do this if you want? https://github.com/opennextjs/opennextjs-aws/pull/1095

packages/cloudflare/src/cli/build/utils/files.ts
dario-piotrowicz: done 👍

packages/cloudflare/src/cli/build/utils/create-config-files.ts
dario-piotrowicz: I think that this is an outdated comment

packages/cloudflare/src/cli/commands/migrate.ts
dario-piotrowicz: Updated to `findNextConfig`

packages/cloudflare/src/cli/commands/migrate.ts
```diff
@@ -0,0 +1,240 @@
...
+
+	const nextConfigFilePath = findNextConfig(projectDir);
+
+	if (!nextConfigFilePath) {
+		logger.error(
+			`No Next.js config file detected, are you sure that this current directory contains a Next.js project? aborting\n`
+		);
+		process.exit(1);
+	}
+
```
vicb, line 31: Maybe replace that with `nextConfigFilePath` for consistency and inline `findNextConfig(projectDir)` where it is used below?

packages/cloudflare/src/cli/commands/migrate.ts
```diff
@@ -0,0 +1,240 @@
...
+		const extensions = [".ts", ".tsx", ".js", ".jsx", ".mjs", ".mts"];
+		const files = findFilesRecursive(projectDir, extensions);
+		let foundEdgeRuntime = false;
+
+		for (const file of files) {
+			try {
+				const content = fs.readFileSync(file, "utf8");
+				if (content.includes('export const runtime = "edge"')) {
+					logger.warn(`Found edge runtime in: ${file}`);
+					foundEdgeRuntime = true;
```
vicb, line 148:
```suggestion
					foundEdgeRuntime = true;
					break;
```

packages/cloudflare/src/cli/commands/migrate.ts
```diff
@@ -0,0 +1,240 @@
...
+	} catch {
+		logger.warn(
+			"Failed to check for edge runtime usage.\n" +
+				"The edge runtime is not supported yet with @opennextjs/cloudflare.\n" +
+				'If present, remove all the `export const runtime = "edge";` lines from your source files'
+		);
+	}
+
+	logger.info(
+		"🎉 OpenNext.js for Cloudflare setup complete!\n" +
```
vicb: ...adapter

.changeset/add-init-command.md
```diff
@@ -0,0 +1,9 @@
+---
+"@opennextjs/cloudflare": minor
+---
+
+feature: add `migrate` command to set up OpenNext.js for Cloudflare adapter
```
vicb:
```suggestion
feature: add `migrate` command to set up OpenNext for Cloudflare adapter
```

packages/cloudflare/src/cli/commands/migrate.ts
dario-piotrowicz: done 👍

.changeset/add-init-command.md
dario-piotrowicz: 🤦 👍

PR #1084, opened 2026-01-22
chore: update c3 template to Next 16

vicb:
Validated, helped catch a bug in 1.15.0 (1.15.1 has been released to fix that)


vicb:
Thanks Matt for the review!

PR #1085, opened 2026-01-22
fix the detection of the runtime (webpack vs turbopack)

vicb:
See https://github.com/opennextjs/opennextjs-aws/pull/1091


vicb:
Thanks Magnus!

PR #1087, opened 2026-01-22
refactor: provision for when unstable_readConfig becomes async

vicb:


vicb:
Thanks for the review Samuel!

packages/cloudflare/src/cli/commands/utils.ts
```diff
@@ -101,8 +101,11 @@ export function getNormalizedOptions(config: OpenNextConfig, buildDir = nextAppD
  * @param args Wrangler environment and config path.
  * @returns Wrangler config.
  */
-export function readWranglerConfig(args: WithWranglerArgs) {
-	return unstable_readConfig({ env: args.env, config: args.wranglerConfigPath });
+export async function readWranglerConfig(args: WithWranglerArgs) {
+	// Note: `unstable_readConfig` is sync as of wrangler 4.60.0
+	//       But it will ultimately become async.
```
vicb:
```suggestion
	//       But it will eventually become async.
```

PR #1092, opened 2026-01-26
Check for supported Next version

vicb:
As discussed offline, we will work on a support policy for Open Next.

A first step in this direction is to error by default when building foran  unmaintained Next version.

For now, you can bypass the check with `--dangerouslyUseUnmaintainedNextVersion`

@conico974 should I create a mirror PR on AWS ? (or I can create the check there and only use it here)

### DIFF
--- a/.changeset/six-walls-feel.md
+++ b/.changeset/six-walls-feel.md
@@ -5,4 +5,4 @@
 Check for supported Next version
 
 The build will now error for unsupported Next version which may contain unpatched security vulnerabilities.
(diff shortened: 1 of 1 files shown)


vicb:
> I guess it's better to have the check done in a single place, probably as a function, so that it could be called here.

Makes sense. Do we want this to be in AWS (see PR description) ?

conico974:
Yeah I think it should be in AWS

vicb:
Converted this PR to draft - I'll move the implementation to AWS

vicb:
Needs https://github.com/opennextjs/opennextjs-aws/pull/1094 to be merged and released

PR #1097, opened 2026-01-26
Apply quality of life improvements to the `opennextjs-cloudflare` CLI

dario-piotrowicz:
This PR improves the `opennextjs-cloudflare` CLI by:
- ensuring that unknown commands (e.g., `opennextjs-cloudflare foo`) display a clear and helpful error message
- adding a `-h`|`--help` flag to display the CLI's help message
- adding a `-v`|`--version` flag to display the package's version

### DIFF
--- a/packages/cloudflare/src/cli/index.ts
+++ b/packages/cloudflare/src/cli/index.ts
@@ -12,7 +12,7 @@ import { addPreviewCommand } from "./commands/preview.js";
 import { addUploadCommand } from "./commands/upload.js";
 
 export function runCommand() {
-	const y = yargs(process.argv.slice(2))
+	const y = yargs(process.argv.slice(2).filter((arg) => arg !== "--"))
 		.scriptName("opennextjs-cloudflare")
 		.parserConfiguration({ "unknown-options-as-args": true })
 		.strictCommands()
--- a/packages/cloudflare/src/cli/commands/utils.ts
+++ b/packages/cloudflare/src/cli/commands/utils.ts
@@ -164,11 +164,8 @@ function getWranglerArgs(
 		...(args.config ? ["--config", args.config] : []),
 		...(args.env ? ["--env", args.env] : []),
 		...(args.remote ? ["--remote"] : []),
-		// Note: the `args` array contains unrecognised flags before the `--` separator.
+		// Note: the `args` array contains unrecognised flags.
 		...(args.args?.map((a) => `${a}`) ?? []),
-		// Note: the first args in `_` will be the commands.
-		//       other values will be unrecognised flags after the `--` separator.
-		...args._.slice(args._[0] === "populateCache" ? 2 : 1).map((a) => `${a}`),
 	];
 }
 


packages/cloudflare/src/cli/index.ts
```diff
@@ -12,7 +13,22 @@ import { addUploadCommand } from "./commands/upload.js";
 export function runCommand() {
 	const y = yargs(process.argv.slice(2).filter((arg) => arg !== "--"))
 		.scriptName("opennextjs-cloudflare")
-		.parserConfiguration({ "unknown-options-as-args": true });
+		.parserConfiguration({ "unknown-options-as-args": true })
+		.strictCommands()
+		.help()
+		.alias("h", "help")
+		.version()
```
dario-piotrowicz: oof... the version doesn't work properly for some reason! 😓

dario-piotrowicz: I've removed the version flag since it seems to be quite problematic with how the project is currently built (without bundling) 😓 

I think we can look into it separately if we want to include the version flag 😓  (but I think that it's going to be pretty tricky)

vicb: Have you tried to use https://github.com/opennextjs/opennextjs-cloudflare/blob/1f88c88e456266022aa79477ec18ad318f399f40/packages/cloudflare/src/cli/build/utils/version.ts#L8 ?

dario-piotrowicz: no I didn't, I've tried now, and that seems to be working nicely, thanks! 😄🥳

packages/cloudflare/src/cli/index.ts
```diff
@@ -12,7 +14,22 @@ import { addUploadCommand } from "./commands/upload.js";
...
-		.parserConfiguration({ "unknown-options-as-args": true });
+		.parserConfiguration({ "unknown-options-as-args": true })
+		.strictCommands()
+		.help()
+		.alias("h", "help")
+		.version(getVersion().cloudflare)
+		.alias("v", "version")
+		.fail((msg, err, yargs) => {
+			if (msg) {
+				logger.error(`Error: ${msg}\n`);
```
james-elicx: Double Error in the logs :p
image

```suggestion
				logger.error(`${msg}\n`);
```

.changeset/yellow-pears-relax.md
```diff
@@ -0,0 +1,11 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+Apply quality of life improvements to the `opennextjs-cloudflare` CLI
```
vicb: Suggest: maybe less ambitous, but clearer to me :) If you use this, please also update the PR title.
```suggestion
Add `--help` and `--version` to the `opennextjs-cloudflare` CLI
```

dario-piotrowicz: Well, one of the (in my opinion) most important fixes here is that this PR makes the CLI print a clear error on unknown commands, I think that's worth keeping in the changeset?

dario-piotrowicz:
oof... I guess I need to also instruct yargs to ignore the flags we forward to wrangler 😓 
https://github.com/opennextjs/opennextjs-cloudflare/actions/runs/21395938691/job/61594141196?pr=1097#step:6:31

james-elicx:
> oof... I guess I need to also instruct yargs to ignore the flags we forward to wrangler 😓 [opennextjs/opennextjs-cloudflare/actions/runs/21395938691/job/61594141196?pr=1097#step:6:31](https://github.com/opennextjs/opennextjs-cloudflare/actions/runs/21395938691/job/61594141196?pr=1097#step:6:31)

I guess that's a result of where .strictCommands() is used perhaps?

.changeset/yellow-pears-relax.md
vicb: Your call.
Will users read that and then go execute invalid commands to see how that works?
No strong opinion. I commented here because I don't think "quality of life improvements" brings me added value.

dario-piotrowicz:
> I guess that's a result of where .strictCommands() is used perhaps?

~No, I think that only checks commands and that's not the issue here 🤔~

maybe...? but without it passing wrong stuff to the opennext CLI doesn't trigger the error message 😓

james-elicx:
Just pushed a change that _should_ get the args working properly.

There's a couple things at play now:
- we're asking yargs to treat unknown options as args instead of failing, and so yargs was treating them like positionals.
- there's a syntax for "[variadic positional arguments](https://github.com/yargs/yargs/blob/main/docs/advanced.md#variadic-positional-arguments)" that we weren't using that would turns the unrecognised flags into args.
- since we're still stripping out the `--` separator, all unrecognised flags that are treated as positionals should now get parsed into an args array.

You can see the args ending up in the expected properties in the below screenshot:

image

james-elicx:
~~Hmm I guess I broke something else with the e2es 😅~~ Should be working now.

dario-piotrowicz:
Thanks for the fix @james-elicx 🙂

With your fix I think that things are a bit un-ideal since any flag passed to a command won't trigger any error/warning, however I don't think we can reasonably do much better given how the CLI accepts arguments

Looks good to me 🙂👍

vicb:
@dario-piotrowicz 

image

I won't at time to look at this before at least tomorrow.
Feel free to merge unless you think I really need to look at this.
Thanks!

dario-piotrowicz:
The current changes look ok... one small thing I noticed is that providing a wrong flag like `--test-123` errors saying `Unknown command` instead of `Unknown argument`:
Screenshot 2026-02-05 at 10 30 02

I'm not sure why... it's probably not a huge deal, this is definitely more helpful that the current behavior, so I'd say that this is ok and we can possibly improve it as a followup

what do you think @james-elicx?

james-elicx:
> The current changes look ok... one small thing I noticed is that providing a wrong flag like `--test-123` errors saying `Unknown command` instead of `Unknown argument`: Screenshot 2026-02-05 at 10 30 02
> 
> I'm not sure why... it's probably not a huge deal, this is definitely more helpful that the current behavior, so I'd say that this is ok and we can possibly improve it as a followup
> 
> what do you think @james-elicx?

It's because `unknown-options-as-args` treats them as positionals instead of erroring from what i understand.

PR #1099, opened 2026-01-28
fix: Use worker binding for R2 cache uploads to avoid API rate limits

### LINKED ISSUE #1088: [BUG] R2 cache upload fails with 500 Internal Server Error during deploy (2400 prerendered pages)
### Describe the bug

When deploying a large Next.js application with ~2,400 prerendered pages, the R2 incremental cache upload fails with a 500 Internal Server
  Error. This occurs even after reducing `--cacheChunkSize` from the default 50 to 10.

  The deploy process:
  1. Next.js build completes successfully (2,423 static pages generated)
  2. OpenNext bundling completes
  3. R2 cache population begins uploading 2,400 objects
  4. Upload fails at ~16% (400/2,400 objects) with HTTP 500

  Error:
  ✘ [ERROR] Error uploading ".open-next/cache/.../activities.cache"
    Error: Failed to fetch /accounts/.../r2/buckets/opennext-cache/objects/... - 500: Internal Server Error

  Additionally, an esbuild deadlock error appears earlier in the build (though the build continues):
  fatal error: all goroutines are asleep - deadlock!
  goroutine 1 [chan receive]:
  github.com/evanw/esbuild/internal/helpers.(*ThreadSafeWaitGroup).Wait(...)

### Steps to reproduce

1. Have a Next.js 16 app with Partial Prerendering that generates ~2,400+ static pages
2. Configure R2 incremental cache in wrangler.jsonc
3. Run `opennextjs-cloudflare build && opennextjs-cloudflare deploy --cacheChunkSize 10`
4. Observe R2 upload failing partway through with 500 error

### Expected behavior

  The R2 cache should upload all 2,400 objects successfully, with proper retry logic for transient API errors.

  1.15.0

### Wrangler version

4.59.3

### next info output

```shell
Operating System:
    Platform: darwin
    Arch: arm64
    Version: Darwin Kernel Version 25.2.0
    Available memory (MB): 36864
    Available CPU cores: 12
  Binaries:
    Node: 24.11.1
    npm: 11.7.0
  Relevant Packages:
    next: 16.1.4
    react: 19.2.3
    react-dom: 19.2.3
    typescript: 5.9.3
  Next.js Config:
    output: standalone
```

### Root Cause Analysis

  The R2 upload uses wrangler's `r2 bulk put` command which calls the **Cloudflare API** (not the S3-compatible API). This has significant
  limitations:

  1. **Cloudflare API Rate Limit**: 1,200 requests per 5 minutes (wrangler uses 1,100 with a buffer)
     - Source: wrangler code at `cli.js:270141-270147`
     ```javascript
     const API_RATE_LIMIT_WINDOWS_MS = 5 * 60 * 1e3;  // 5 minutes
     const API_RATE_LIMIT_REQUESTS = 1200 - 100;      // 1100 requests

  2. No Retry Logic: The fetchR2Objects function in wrangler throws immediately on any non-200/404 response:
  if (response.ok && response.body) {
    return response;
  } else if (response.status === 404) {
    return null;
  } else {
    throw new Error(`Failed to fetch ${resource} - ${response.status}: ${response.statusText});`);
  }
  3. No Retry in opennextjs: The runWrangler function exits immediately on any non-zero status code with no retry mechanism.
  4. Single Batch Upload: Unlike KV cache (which chunks uploads), R2 sends all 2,400 objects in a single r2 bulk put command. A single file
  failure aborts the entire upload.

  Scale Issue

  For 2,400 files at the rate limit:
  - Best case: 2,400 / 1,100 = 2.18 rate limit cycles = ~11 minutes minimum
  - With concurrency 10, the queue still counts toward the same rate limit window
  - 500 errors may be Cloudflare API overload rather than actual rate limiting

  Cloudflare's Recommendation

  Per Cloudflare community forums and docs, wrangler is designed for single-object uploads. For bulk uploads, Cloudflare recommends:
  - rclone with S3-compatible API
  - AWS SDK with R2 S3-compatible endpoint

  Feature Request

  Consider adding S3-compatible API support for R2 bulk uploads in opennextjs-cloudflare:
  - Uses endpoint: <account-id>.r2.cloudflarestorage.com
  - R2 API tokens (Access Key ID + Secret Access Key)
  - Bypasses Cloudflare API rate limits
  - Can use parallel uploads with AWS SDK's built-in retry logic

OR (even better), remote bindings would probably fix this.

  Workarounds Attempted

  - Reduced --cacheChunkSize from 50 to 10 - still failed
  - The error occurs at ~400/2,400 objects (~16%)

  Environment

  - Build environment: Cloudflare Pages (Node.js v24.13.0)
  - @opennextjs/aws version: 3.9.11
  - Cache Components and Partial Prerender (PPR) enabled
  - ~2,400 prerendered pages generating ~2,400 cache files

  ---

  **Sources:**
  - [Cloudflare API Rate Limits](https://developers.cloudflare.com/fundamentals/api/reference/limits/) - 1,200 requests/5 min
  - [R2 Platform Limits](https://developers.cloudflare.com/r2/platform/limits/)
  - [Community: R2 Wrangler Rate
  Limits](https://community.cloudflare.com/t/cloudflare-r2-wrangler-what-is-the-rate-limit-for-uploading-objects-to-buckets/772242) - HTTP 429
   issues

isaacrowntree:
## Summary

Fixes the R2 cache upload failure for large Next.js applications with thousands of prerendered pages. The issue is caused by the Cloudflare API rate limit of 1,200 requests per 5 minutes when using wrangler's `r2 bulk put` command.

**This approach has been successfully running in production** on [campermate.com](https://campermate.com) since the initial fix, populating 15+ cache entries per deploy with zero failures.

## Approach

Uses `unstable_startWorker` (wrangler) to spin up an isolated local worker with a remote R2 binding, bypassing Cloudflare API rate limits entirely.

### How it works

1. **Create an empty `wrangler.toml`** in a temp directory to isolate the worker from the project's config
2. **Start a local worker** via `unstable_startWorker` with only the R2 cache binding
3. **Send batched cache entries** to the worker via `worker.fetch` directly
4. **Retry on failures** — handles both partial failures (207) and network errors with exponential backoff
5. **Clean up** — dispose worker and remove temp directory

### Key design decisions

- **Config isolation via empty `wrangler.toml`** — `unstable_startWorker` uses `findUpSync` from `path.dirname(entrypoint)` to discover config. Without isolation, it picks up the project's full `wrangler.toml` and binds ALL project resources (DOs, KV, AI, etc.) to the worker, causing hangs and errors. Passing `config: emptyConfigPath` short-circuits the search.

- **`worker.fetch` instead of `global.fetch(worker.url)`** — Using the worker's fetch method directly avoids an unnecessary HTTP hop through localhost. The `DispatchFetch` type (miniflare) is compatible at runtime, requiring only a type cast: `worker.fetch as unknown as typeof globalThis.fetch`.

- **Retry logic with exponential backoff** — R2 writes can fail transiently (especially remote). The worker returns 207 with error details on partial failure, and the caller retries only the failed entries.

- **Single code path** — always uses `unstable_startWorker` for both local and remote targets. The `remote` flag on the R2 binding controls which bucket is used.

### Why `unstable_startWorker` over Miniflare directly?

We evaluated using `new Miniflare()` directly, which would avoid the config discovery issue entirely (fully programmatic, no `findUpSync`). However:

- **Remote R2 is the dealbreaker** — `unstable_startWorker` with `remote: true` handles all auth/proxy plumbing to write to actual Cloudflare R2 buckets automatically. With raw Miniflare, you'd need to manually construct a `RemoteProxyConnectionString` and handle the auth flow.
- **For local-only R2** (e.g. `target: "local"`), Miniflare would actually be cleaner — no config discovery, no temp files, and direct access via `mf.getR2Bucket()`.
- The empty `wrangler.toml` workaround is a one-liner fix for a well-understood problem, making it an acceptable trade-off vs reimplementing wrangler's remote proxy logic.

## Files changed

| File | Change |
|------|--------|
| `populate-cache.ts` | Replace `wrangler r2 bulk put` with `unstable_startWorker` approach, config isolation, retry logic |
| `r2-cache-populate-handler.ts` | Standalone worker for R2 writes (batch POST endpoint, concurrent writes) |
| `populate-cache.spec.ts` | Updated tests for `worker.fetch` pattern |

## Test plan

- [x] TypeScript compiles without errors
- [x] All tests pass
- [x] **Production validated** — deployed to campermate.com, 15 cache entries populated in ~9 seconds with isolated worker bindings
- [x] Worker shows only `NEXT_INC_CACHE_R2_BUCKET` binding (no leaked project bindings)
- [x] Retry logic handles 207 partial failures and network errors

## Related

- Fixes #1088
- See also #1121 (@vicb's follow-up WIP exploring the same approach)

### DIFF
--- a/.changeset/fix-r2-cache-upload.md
+++ b/.changeset/fix-r2-cache-upload.md
@@ -1,0 +1,18 @@
+---
+"@opennextjs/cloudflare": minor
+---
+
+fix: Use remote R2 binding for cache population to avoid API rate limits
+
+For deployments with prerendered pages, the R2 incremental cache is now populated
+using `unstable_startWorker` with a remote R2 binding instead of `wrangler r2 bulk put`.
+This bypasses the Cloudflare API rate limit of 1,200 requests per 5 minutes that caused
+failures for large applications with thousands of prerendered pages.
+
+The new approach:
+1. Starts a local worker via `unstable_startWorker` with the R2 binding configured programmatically
+2. Sends cache entries to the local worker a few at a time for low memory usage
+3. The worker writes entries to R2 via the binding (no API rate limits)
+4. No deployment, temp config files, or authentication tokens required
+
+Closes #1088
--- a/packages/cloudflare/src/cli/commands/populate-cache.ts
+++ b/packages/cloudflare/src/cli/commands/populate-cache.ts
@@ -257,33 +257,46 @@ async function populateR2IncrementalCache(
 	const useRemote = populateCacheOptions.target === "remote";
 	const handlerPath = getCachePopulateHandlerPath();
 
-	const worker = await unstable_startWorker({
-		name: "open-next-cache-populate",
-		entrypoint: handlerPath,
-		compatibilityDate: "2026-01-01",
-		bindings: {
-			[R2_CACHE_BINDING_NAME]: {
-				type: "r2_bucket",
-				bucket_name: binding.bucket_name,
-				...(binding.jurisdiction && { jurisdiction: binding.jurisdiction }),
-				remote: useRemote,
-			},
-		},
-		dev: {
-			server: { port: 0 },
-			inspector: false,
-			watch: false,
-			liveReload: false,
-		},
-	});
+	// Start the worker from a temp directory to prevent unstable_startWorker
+	// from picking up the project's wrangler.jsonc (which would merge in all
+	// bindings like DOs, KV, services, AI, etc. and cause hangs/errors).
+	const tempDir = await fsp.mkdtemp(path.join(os.tmpdir(), "open-next-r2-populate-"));
+	const originalCwd = process.cwd();
 
 	try {
-		await worker.ready;
-		const url = await worker.url;
-		const populateUrl = new URL("/populate", url).href;
-		await sendCacheEntries(populateUrl, assets, prefix, populateCacheOptions.cacheChunkSize);
+		process.chdir(tempDir);
+
+		const worker = await unstable_startWorker({
+			name: "open-next-cache-populate",
+			entrypoint: handlerPath,
+			compatibilityDate: "2026-01-01",
+			bindings: {
+				[R2_CACHE_BINDING_NAME]: {
+					type: "r2_bucket",
+					bucket_name: binding.bucket_name,
+					...(binding.jurisdiction && { jurisdiction: binding.jurisdiction }),
+					remote: useRemote,
+				},
+			},
+			dev: {
+				server: { port: 0 },
+				inspector: false,
+				watch: false,
+				liveReload: false,
+			},
+		});
+
+		try {
+			await worker.ready;
+			const url = await worker.url;
+			const populateUrl = new URL("/populate", url).href;
+			await sendCacheEntries(populateUrl, assets, prefix, populateCacheOptions.cacheChunkSize);
+		} finally {
+			await worker.dispose();
+		}
 	} finally {
-		await worker.dispose();
+		process.chdir(originalCwd);
+		fs.rmSync(tempDir, { recursive: true, force: true });
 	}
 
 	logger.info(`Successfully populated cache with ${assets.length} assets`);
--- a/packages/cloudflare/src/cli/commands/populate-cache.ts
+++ b/packages/cloudflare/src/cli/commands/populate-cache.ts
@@ -355,6 +355,11 @@ async function sendCacheEntries(
 
 /**
  * Sends a batch of cache entries to the local worker with retry logic.
+ *
+ * Retries on:
+ * - HTTP errors (non-200/207 responses)
+ * - Network failures
+ * - Partial failures (207 responses) — only the failed entries are retried
  */
 async function sendBatchWithRetry(
 	workerUrl: string,
@@ -363,10 +368,12 @@ async function sendBatchWithRetry(
 	retryDelayMs = 1000
 ): Promise<{ written: number; failed: number; errors?: string[] }> {
 	let lastError: Error | undefined;
+	let totalWritten = 0;
+	let remainingEntries = entries;
 
 	for (let attempt = 0; attempt < maxRetries; attempt++) {
 		if (attempt > 0) {
-			logger.info(`Retrying batch (attempt ${attempt + 1}/${maxRetries})...`);
+			logger.info(`Retrying ${remainingEntries.length} failed entries (attempt ${attempt + 1}/${maxRetries})...`);
 			await sleep(retryDelayMs * Math.pow(2, attempt - 1));
 		}
 
@@ -374,7 +381,7 @@ async function sendBatchWithRetry(
 			const response = await fetch(workerUrl, {
 				method: "POST",
 				headers: { "Content-Type": "application/json" },
-				body: JSON.stringify({ entries }),
+				body: JSON.stringify({ entries: remainingEntries }),
 			});
 
 			if (!response.ok && response.status !== 207) {
@@ -382,14 +389,31 @@ async function sendBatchWithRetry(
 				throw new Error(`HTTP ${response.status}: ${text || response.statusText}`);
 			}
 
-			return (await response.json()) as { written: number; failed: number; errors?: string[] };
+			const result = (await response.json()) as { written: number; failed: number; errors?: string[] };
+			totalWritten += result.written;
+
+			if (result.failed === 0) {
+				return { written: totalWritten, failed: 0 };
+			}
+
+			// Extract the keys that failed so we can retry just those
+			const failedKeys = new Set(result.errors?.map((e) => e.split(":")[0]?.trim()).filter(Boolean));
+			if (failedKeys.size > 0) {
+				remainingEntries = remainingEntries.filter((e) => failedKeys.has(e.key));
+			}
+
+			lastError = new Error(`${result.failed} entries failed: ${result.errors?.slice(0, 3).join(", ")}`);
 		} catch (error) {
 			lastError = error instanceof Error ? error : new Error(String(error));
 			logger.warn(`Batch failed: ${lastError.message}`);
 		}
 	}
 
-	throw new Error(`Failed to populate batch after ${maxRetries} attempts: ${lastError?.message}`);
+	return {
+		written: totalWritten,
+		failed: remainingEntries.length,
+		errors: lastError ? [lastError.message] : undefined,
+	};
 }
 
 function sleep(ms: number): Promise<void> {
--- a/packages/cloudflare/src/cli/commands/populate-cache.ts
+++ b/packages/cloudflare/src/cli/commands/populate-cache.ts
@@ -2,6 +2,7 @@ import fs from "node:fs";
 import fsp from "node:fs/promises";
 import os from "node:os";
 import path from "node:path";
+import { fileURLToPath } from "node:url";
 
 import type { BuildOptions } from "@opennextjs/aws/build/helper.js";
 import logger from "@opennextjs/aws/logger.js";
@@ -14,6 +15,7 @@ import type {
 import type { IncrementalCache, TagCache } from "@opennextjs/aws/types/overrides.js";
 import { globSync } from "glob";
 import { tqdm } from "ts-tqdm";
+import { unstable_startWorker } from "wrangler";
 import type { Unstable_Config as WranglerConfig } from "wrangler";
 import type yargs from "yargs";
 
@@ -196,9 +198,9 @@ type PopulateCacheOptions = {
 	 */
 	wranglerConfigPath?: string;
 	/**
-	 * Chunk sizes to use when populating KV cache. Ignored for R2.
+	 * Chunk sizes to use when populating the cache.
 	 *
-	 * @default 25 for KV, 50 for R2
+	 * @default 25 for KV, 5 for R2
 	 */
 	cacheChunkSize?: number;
 	/**
@@ -207,6 +209,28 @@ type PopulateCacheOptions = {
 	shouldUsePreviewId: boolean;
 };
 
+/**
+ * Resolves the path to the R2 cache populate handler file.
+ *
+ * The handler is a standalone worker used with `unstable_startWorker`.
+ * It's located in the workers directory relative to this file.
+ */
+function getCachePopulateHandlerPath(): string {
+	const currentDir = path.dirname(fileURLToPath(import.meta.url));
+	return path.join(currentDir, "../workers/r2-cache-populate-handler.js");
+}
+
+/**
+ * Populates the R2 incremental cache using a local worker with a remote R2 binding.
+ *
+ * This approach:
+ * 1. Starts a local worker via `unstable_startWorker` with the R2 cache binding.
+ * 2. Sends cache entries to the local worker a few at a time.
+ * 3. The worker writes entries to R2 using the binding (no API rate limits).
+ *
+ * This bypasses the Cloudflare API rate limit of 1,200 requests per 5 minutes
+ * that affects `wrangler r2 bulk put`.
+ */
 async function populateR2IncrementalCache(
 	buildOpts: BuildOptions,
 	config: WranglerConfig,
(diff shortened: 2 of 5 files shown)


vicb:
Thanks for the PR @isaacrowntree 
I can't promise I'll get to it today but by tomorrow EOD for sure

vicb:
Sorry @isaacrowntree 
I wasted my day today... it happens...
I'll try to review tomorrow

.changeset/fix-r2-cache-upload.md
```diff
@@ -0,0 +1,36 @@
...
+```bash
+# Auto-select based on cache size (default) - uses binding for 500+ entries
+opennextjs-cloudflare deploy --cacheMethod=auto
+
+# Force wrangler CLI (original behavior)
+opennextjs-cloudflare deploy --cacheMethod=wrangler
+
+# Force worker binding (bypasses API rate limits)
+opennextjs-cloudflare deploy --cacheMethod=binding
+```
```
vicb: I would always use remote bindings to simplify things.

It can do both > and <= 500.

Then we do not need to maintain 2 different code paths each with a different behavior (and bugs)

packages/cloudflare/src/cli/build/open-next/compile-cache-populate-handler.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import type { BuildOptions } from "@opennextjs/aws/build/helper.js";
+import { build } from "esbuild";
+
+/**
+ * Compiles the cache populate handler for the workerd runtime.
+ * This handler allows the CLI to populate the R2 cache directly via the worker,
+ * bypassing wrangler's rate-limited r2 bulk put command.
+ */
+export async function compileCachePopulateHandler(options: BuildOptions) {
```
vicb: Do we need to compile this?

`wrangler` accepts ts file

packages/cloudflare/src/cli/commands/deploy.ts
```diff
@@ -57,16 +155,175 @@ export async function deployCommand(args: WithWranglerArgs<{ cacheChunkSize?: nu
...
+): Promise<void> {
+	// Generate a temporary token for cache population
+	const cachePopulateToken = generateCachePopulateToken();
+
+	const deploymentMapping = await getDeploymentMapping(buildOpts, config, envVars);
+
+	// Step 1: Deploy worker with the cache populate token
+	logger.info("\n📦 Step 1/3: Deploying worker with cache populate endpoint...");
+
+	const deployArgs = [
```
vicb: I don't think we need to deploy anything.

Remote bindings can interact with remote R2 in dev

vicb (changes requested):
I had something simpler in mind.

1) Derive a wrangler config from the project's wrangler config. What needs to be propagated is the R2 cache binding. It should have `remote` set to `true` if we want to populate the remote R2 (`false` otherwise).
2) Start a local worker using this config with a POSt endpoint
3) Send rquests to this post endpoint.

Disclaimer: quick review, not much time tonight but I hope it makes at least some sense.

isaacrowntree:
Hey @vicb, thanks for the review! I've reworked the approach based on your feedback.

**What changed:**

Replaced the deploy→populate→redeploy cycle with your suggested approach: a local `wrangler dev` worker with a remote R2 binding. The flow is now:

1. Derive a temp wrangler config from the project's config (just the R2 binding with `remote: true`)
2. Start `wrangler dev` locally with a standalone worker that has a POST endpoint
3. Send batched cache entries to the local worker, which writes to R2 via the binding
4. Stop the worker, clean up, then deploy normally

**What was removed:**

- `--cacheMethod` flag and dual code paths — single path for all cache sizes
- Token auth system — local worker doesn't need it
- The esbuild compilation step — wrangler handles TS natively
- Changes to `deploy.ts`, `worker.ts`, `cloudflare-context.ts`, `build.ts`, `run-wrangler.ts` — all reverted to main
- `compile-cache-populate-handler.ts` — deleted

Net result is ~470 fewer lines. The `remote` flag on the R2 binding controls local vs remote target, so the same code path handles both `populateCache local` and `populateCache remote`.

All 206 tests pass. Let me know if this is closer to what you had in mind.

.changeset/fix-r2-cache-upload.md
isaacrowntree: Done — removed the dual code paths and `--cacheMethod` flag entirely. Now always uses remote bindings via a local `wrangler dev` worker for R2 cache population, regardless of cache size. Single code path.

packages/cloudflare/src/cli/build/open-next/compile-cache-populate-handler.ts
isaacrowntree: Removed — deleted `compile-cache-populate-handler.ts` and the esbuild step. The standalone worker is now a plain `.ts` file that wrangler dev runs directly.

packages/cloudflare/src/cli/commands/deploy.ts
isaacrowntree: Done — reverted `deploy.ts` back to main (no changes). The R2 cache is now populated by starting a local `wrangler dev` worker with the R2 binding set to `remote: true`. No deploy/redeploy cycle needed.

packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -222,53 +249,212 @@ async function populateR2IncrementalCache(
...
+
+	// Create a temporary wrangler config derived from the project's config.
+	// Only the R2 cache binding is propagated, with `remote` set appropriately.
+	const tempDir = await fsp.mkdtemp(path.join(os.tmpdir(), "open-next-r2-populate-"));
+
+	try {
+		const tempWranglerConfig = {
+			name: "open-next-cache-populate",
+			main: handlerPath,
+			compatibility_date: "2024-12-01",
```
vicb:
```suggestion
			compatibility_date: "2026-01-01",
```

packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -222,53 +249,212 @@ async function populateR2IncrementalCache(
...
+			r2_buckets: [
+				{
+					binding: R2_CACHE_BINDING_NAME,
+					bucket_name: binding.bucket_name,
+					...(binding.jurisdiction && { jurisdiction: binding.jurisdiction }),
+					...(useRemote && { remote: true }),
```
vicb:
```suggestion
					remote: useRemote,
```

packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -222,53 +249,212 @@ async function populateR2IncrementalCache(
...
+				"--port",
+				"0",
+			],
+			{
+				shell: true,
+				stdio: ["ignore", "pipe", "pipe"],
+				env: {
+					...process.env,
+					CLOUDFLARE_LOAD_DEV_VARS_FROM_DOT_ENV: "false",
+				},
```
vicb: I don't think we need it for this simple worker
```suggestion
```

packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -222,53 +249,212 @@ async function populateR2IncrementalCache(
...
-			// R2 does not support the environment flag and results in the following error:
-			// Incorrect type for the 'cacheExpiry' field on 'HttpMetadata': the provided value is not of type 'date'.
-			environment: undefined,
-			logging: "error",
+/**
+ * Starts `wrangler dev` with the given config and waits for it to be ready.
+ *
+ * @returns The local URL and a function to stop the worker.
+ */
+function startWranglerDev(
```
vicb: Could you please use the [`unstable_startWorker`](https://developers.cloudflare.com/workers/wrangler/api/#unstable_startworker)?

This should make the code quite simpler.

packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -222,53 +249,212 @@ async function populateR2IncrementalCache(
...
+		const entries = batchAssets.map(({ fullPath, key, buildId, isFetch }) => ({
+			key: computeCacheKey(key, {
+				prefix,
+				buildId,
+				cacheType: isFetch ? "fetch" : "cache",
+			}),
+			value: fs.readFileSync(fullPath, "utf8"),
+		}));
+
+		const result = await sendBatchWithRetry(workerUrl, entries);
```
vicb, line 337: Given that the server is local, there is probably no need to batch (which might consumes a lot of memory if you have a lot of assets).

Sending a few assets at a time should be good enough and less memory intensive.

packages/cloudflare/src/cli/workers/r2-cache-populate-handler.ts
vicb, line 1: The `templates` stores part of the worker that will be deployed as the app.

What about creating a `workers` folder at the same level and add r2 to the file name?

packages/cloudflare/src/cli/templates/cache-populate-handler.ts
```diff
@@ -0,0 +1,129 @@
...
+	};
+
+	return new Response(JSON.stringify(response), {
+		status: failed === 0 ? 200 : 207,
+		headers: { "Content-Type": "application/json" },
+	});
+}
+
+export default {
+	fetch: (request: Request, env: CachePopulateEnv) => handleCachePopulate(request, env),
```
vicb: Maybe the fetch handler should check the method (POSTà, the pathname and if ok call `popuateCache`.

I think it would be easier to test `populateCache` that way.

vicb (commented):
Nice to see this PR taking a great shape!

Thanks for your work on this

packages/cloudflare/src/cli/commands/populate-cache.ts
isaacrowntree: Done — updated to `"2026-01-01"`. Also removed the temp config file entirely by switching to `unstable_startWorker` with programmatic bindings (see comment on #305).

packages/cloudflare/src/cli/commands/populate-cache.ts
isaacrowntree: Done — simplified to `remote: useRemote`. Now passed directly in the `unstable_startWorker` bindings config.

packages/cloudflare/src/cli/commands/populate-cache.ts
isaacrowntree: Removed — the entire `spawn` approach is gone now that we use `unstable_startWorker`, so this env var is no longer relevant.

packages/cloudflare/src/cli/commands/populate-cache.ts
isaacrowntree: Done — replaced the manual `spawn` + stdout parsing with `unstable_startWorker`. This also eliminated the temp wrangler config file since bindings are now passed programmatically. The code is significantly simpler:

```ts
const worker = await unstable_startWorker({
  name: "open-next-cache-populate",
  entrypoint: handlerPath,
  compatibilityDate: "2026-01-01",
  bindings: {
    [R2_CACHE_BINDING_NAME]: {
      type: "r2_bucket",
      bucket_name: binding.bucket_name,
      remote: useRemote,
    },
  },
  dev: { server: { port: 0 }, inspector: false, watch: false, liveReload: false },
});
```

Uses `worker.ready`, `worker.url`, and `worker.dispose()` for lifecycle management.

packages/cloudflare/src/cli/commands/populate-cache.ts
isaacrowntree: Good call — reduced the default batch size from 100 to 5. Since the server is local, the overhead per request is negligible and this keeps memory usage low even for projects with thousands of prerendered pages.

packages/cloudflare/src/cli/workers/r2-cache-populate-handler.ts
isaacrowntree: Done — moved to `packages/cloudflare/src/cli/workers/r2-cache-populate-handler.ts`. Created a new `workers/` directory at the same level as `templates/`, with "r2" in the filename.

packages/cloudflare/src/cli/templates/cache-populate-handler.ts
isaacrowntree: Done — the fetch handler now checks method and pathname, then delegates to `populateCache`:

```ts
export default {
  async fetch(request: Request, env: CachePopulateEnv): Promise<Response> {
    const url = new URL(request.url);
    if (request.method === "POST" && url.pathname === "/populate") {
      return populateCache(request, env);
    }
    return new Response("Not found", { status: 404 });
  },
};
```

`populateCache` is exported separately and can be tested directly without going through routing. Added routing tests as well.

isaacrowntree:
Hey @vicb — saw your follow-up in #1121, great to see this being explored further!

We've been running this version in production on [campermate.com](https://campermate.com) and wanted to share some findings from getting it working end-to-end:

### Config isolation is critical

The biggest issue we hit was `unstable_startWorker` picking up the project's full `wrangler.toml` via `findUpSync` from `path.dirname(entrypoint)`. This meant the worker got ALL project bindings (Durable Objects, KV, AI, etc.), causing it to hang indefinitely during cache population.

**Fix:** Pass `config: emptyConfigPath` pointing to an empty `wrangler.toml` in a temp directory. This short-circuits the config file discovery so the worker only gets the explicitly-passed R2 binding.

### `worker.fetch` > `global.fetch(worker.url)`

Using `worker.fetch` directly instead of fetching via the URL avoids an unnecessary localhost HTTP hop. The `DispatchFetch` type from miniflare is compatible at runtime — just needs a type cast.

### Retry logic for transient R2 failures

Remote R2 writes can fail transiently. The worker returns 207 with error details on partial failure, and the caller retries only the failed entries with exponential backoff.

### Why not Miniflare directly?

We also evaluated `new Miniflare()` directly, which would sidestep the config discovery issue entirely. The blocker is remote R2 — `unstable_startWorker` with `remote: true` handles all the auth/proxy plumbing automatically. For a **local-only** target, Miniflare would be cleaner (fully programmatic, no temp files, direct `mf.getR2Bucket()` access).

Happy to coordinate if you'd like to incorporate any of these fixes into #1121!

vicb:
This was implemented in #1159

PR #1104, opened 2026-02-02
chore: enable cache for eslint and prettier

vicb:
Speed up eslint and prettier after the first run 🎉


vicb:
Thanks for the reviews!

Issue #1103, opened 2026-02-02
Massive Number of Requests and Billable Duration on Durable Objects (DOShardedTagCache)


rickyfang:
We are running a Next.js application on Cloudflare Workers using ISR. The site is relatively large, with approximately 10k news posts and around 20k tag pages.

On average, the site receives about 150k–200k requests per day. Overall performance has been stable; however, after a few days of operation, we’ve observed a significant spike in Durable Objects usage.

As shown in the attached screenshot, Durable Objects are handling roughly 12 million requests in a single day, which would result in costs exceeding USD 50 per day. We’d like to understand what might be causing this behavior and whether there are any recommended ways to reduce Durable Objects request volume.

Additionally, we’ve noticed that KV storage usage continues to increase over time. Our assumption is that this will eventually stabilize once all pages have been generated via ISR, but we would appreciate confirmation on whether this behavior is expected.

Thank you so much for your help on this!!!

<img width="1628" height="996" alt="Image" src="https://github.com/user-attachments/assets/e7cfa908-581a-4737-ab3e-ef700134739b" />

vicb:
Could you please add your cache config here (open next config)

rickyfang:
@vicb thanks for your quick response, here you go 

```ts
import { defineCloudflareConfig } from "@opennextjs/cloudflare";
import { purgeCache } from "@opennextjs/cloudflare/overrides/cache-purge/index";
import kvIncrementalCache from "@opennextjs/cloudflare/overrides/incremental-cache/kv-incremental-cache";
import doQueue from "@opennextjs/cloudflare/overrides/queue/do-queue";
import doShardedTagCache from "@opennextjs/cloudflare/overrides/tag-cache/do-sharded-tag-cache";

export default defineCloudflareConfig({
  incrementalCache: kvIncrementalCache,
  queue: doQueue,
  // This is only required if you use On-demand revalidation
  tagCache: doShardedTagCache({ baseShardSize: 12 }),
  // Disable this if you want to use PPR
  enableCacheInterception: true,
  // you can also use the `durableObject` option to use a durable object as a cache purge
  cachePurge: purgeCache({ type: "direct" }),
});
```

vicb:
Any particular reason why you are using KV over R2?

See https://opennext.js.org/cloudflare/caching#large-site-using-revalidation

```ts
  incrementalCache: withRegionalCache(r2IncrementalCache, {
    mode: "long-lived",
[...]

PR #1105, opened 2026-02-02
allow using a regional cache with KV

vicb:
I think the name "regional cache" is misleading (for historical reason).

When it was first introduced, the idea was to add a regional cache to R2.
We would not allow it for KV at it already has a built-in regional cache.

But:

- R2 also has a builtin regional cache
- The regional cache handles lazy updates, purge, ... that could help even if the store already has a cache.

A longer term fix would be to rename the concept.

ref: https://github.com/opennextjs/opennextjs-cloudflare/issues/1103
/cc @rickyfang

### DIFF
--- a/.changeset/tidy-symbols-wonder.md
+++ b/.changeset/tidy-symbols-wonder.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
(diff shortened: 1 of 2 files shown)


vicb:
Thanks Nico!

PR #1107, opened 2026-02-02
refactor: vitetest config

vicb:
`pnpm test` was failing for me earlier this morning because it was looking at a generated file:

```
Scope: 24 of 25 workspace projects
packages/cloudflare test$ vitest --run
│ failed to load config from /Users/vberchet/code/work/vite.config.js
│ ⎯⎯⎯⎯⎯⎯⎯ Startup Error ⎯⎯⎯⎯⎯⎯⎯⎯
│ Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'vite' imported from /Users/vberchet/code/work/vite.config.js.timestamp-1770023965105-d1c08c2780d9b.mjs
│ Did you mean to import "vite/index.cjs"?
...
```

Claude thinks it's a reasonable fix, let's try that unless someone has a better idea?

### DIFF
--- a/packages/cloudflare/vitest.config.ts
+++ b/packages/cloudflare/vitest.config.ts
@@ -1,0 +1,20 @@
+import { defineConfig } from "vitest/config";
+
+export default defineConfig({
+	test: {
+		/**
+		 * Explicitly set root to current directory for this package.
+		 *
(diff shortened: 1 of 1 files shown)


vicb:
Thanks James!

Issue #834, opened 2025-08-01
Add documentation and example for using DO tag cache with skew protection


alex-all3dp:
Follow-up to https://github.com/opennextjs/opennextjs-cloudflare/issues/783

Stated in the [docs](https://opennext.js.org/cloudflare/howtos/skew):

> Preview URLs are disabled for [Workers that implement a Durable Object](https://developers.cloudflare.com/workers/configuration/previews/#limitations). If your app uses Durable Objects, they will need to be implemented in a separate Worker.

It would be great if the documentation would be extended with an example on how to get this multi-worker setup with a DO tag cache and queue working.

Thank you!

alex-all3dp:
@vicb Any update on when you might have time to provide an example for tihs? Would be great to be able to combine skew protection with DO tag cache and queue! :)

alex-all3dp:
@vicb Do you think you will have time for this anytime soon? Not being able to use DOs for tag cache as well as cache purge is causing issues on our end, due to frequest memory exceeded errors. Would be great to have a working production setup that combines skew protection with the benefits of offloading work to DOs

alex-all3dp:
@vicb 

I attempted to implement the approach you recommended (if I understood it correctly):  Moving all DOs (DOQueueHandler, DOShardedTagCache, BucketCachePurge) to a separate worker and referencing them from the main worker using `script_name`:

```
// main worker wrangler config
{
  "preview_urls": true,
  "durable_objects": {
    "bindings": [
      { "name": "NEXT_CACHE_DO_QUEUE", "class_name": "DOQueueHandler", "script_name": "my-app-do" },
      { "name": "NEXT_TAG_CACHE_DO_SHARDED", "class_name": "DOShardedTagCache", "script_name": "my-app-do" },
      { "name": "NEXT_CACHE_DO_PURGE", "class_name": "BucketCachePurge", "script_name": "my-app-do" }
    ]
  }
}

// separate DO worker wrangler config
{
  "name": "my-app-do",
  "main": "do-worker.ts", // re-exports DO classes from .open-next/worker.js
  "durable_objects": {
    "bindings": [
      { "name": "NEXT_CACHE_DO_QUEUE", "class_name": "DOQueueHandler" },
      { "name": "NEXT_TAG_CACHE_DO_SHARDED", "class_name": "DOShardedTagCache" },
      { "name": "NEXT_CACHE_DO_PURGE", "class_name": "BucketCachePurge" }
    ]
  },
[...]

Issue #783, opened 2025-07-14
Question: Durable Objects for revalidation working with skew protection enabled after all?


alex-all3dp:
Hi,

first of all thanks for shipping skew protection, really amazing to have this feature being available! :)

I deployed a test version of our app now with skew protection enabled now and was expecting some errors regarding caching in the logs, because the docs stated that Durable Objects are not supported for preview URLs.

Our open-next config is

```typescript
import {defineCloudflareConfig} from '@opennextjs/cloudflare'
import r2IncrementalCache from '@opennextjs/cloudflare/overrides/incremental-cache/r2-incremental-cache'
import {withRegionalCache} from '@opennextjs/cloudflare/overrides/incremental-cache/regional-cache'
import doQueue from '@opennextjs/cloudflare/overrides/queue/do-queue'
import doShardedTagCache from '@opennextjs/cloudflare/overrides/tag-cache/do-sharded-tag-cache'

const openNextConfig = {
  ...defineCloudflareConfig({
    incrementalCache: withRegionalCache(r2IncrementalCache, {
      mode: 'long-lived',
      shouldLazilyUpdateOnCacheHit: true,
    }),
    queue: doQueue,
    tagCache: doShardedTagCache,
  }),
  cloudflare: {
    skewProtection: {
      enabled: true,
      maxNumberOfVersions: 20, // Maximum number of previous versions to use.
      maxVersionAgeDays: 60, // Age of the oldest version to use (from the last deployment date)
    },
  },
}

export default openNextConfig
```

Using the DO sharded tag cache is the recommended setup for larger sites using revalidation, which aligns with our use case. 

When I checked the access metrics for the durable object of our test deployment I saw successful requests at the time when I tested the app.

<img width="1082" height="475" alt="Image" src="https://github.com/user-attachments/assets/91937448-b2f5-46eb-b1ac-2631b68b968a" />

Can you provide some insights regarding the expectations for combining skew protection and Durable Objects for revalidation? In the dashboard everything looks fine but should we work under the assumption that revalidation and the incremental cache in general will not be working properly with skew protection enabled?

Thank you for your support!

Best
Alex

vicb:
Do you see routable preview versions in your dashboard?

<img width="227" height="525" alt="Image" src="https://github.com/user-attachments/assets/7be5cfce-30b0-4dc0-a78e-b62f80c1d138" />

"Routable" meaning you can click on a preview version and see the corresponding deployment

[...]

PR #1120, opened 2026-02-09
feat: add DO proxy layer for supporting DO operations through service binding

alex-all3dp:
Follow-up to https://github.com/opennextjs/opennextjs-cloudflare/issues/783 and https://github.com/opennextjs/opennextjs-cloudflare/issues/834

@vicb It seemed that your suggested approach of using a separate worker does not work out of the box, see my comment in #834. I had Claude attempt a possible solution by adding a DO proxy layer for using service binding between the "main" and the "do worker". I tested it in our setup and the DO RPC traffic flowed correctly and skew protection also still worked for the main worker.

Looking forward to your feedback.

## Summary

Adds a DO proxy layer under `src/api/do-proxy/` that routes Durable Object
operations through a service binding to a separate DO worker via HTTP.

This is one of way solving the current limitation where workers with `durable_objects` config
cannot use preview URLs, which is requiref for skew protection. By splitting DOs into a
separate worker and proxying via service binding, the main worker retains preview
URL support.

## Usage

**Main worker** (`worker.ts`):
```ts
import { withDOProxy } from "@opennextjs/cloudflare/do-proxy/index";
import handler from "./.open-next/worker.js";
export default withDOProxy(handler);
DO worker (do-worker.ts):


import { createDOProxyHandler } from "@opennextjs/cloudflare/do-proxy/index";
export { DOQueueHandler, DOShardedTagCache, BucketCachePurge } from "./.open-next/worker.js";
export default createDOProxyHandler();
```

**Main worker wrangler config** (`wrangler.jsonc`):
```jsonc
{
  "name": "my-app",
  // No durable_objects block — this enables preview URLs / skew protection
  "services": [
    {
      "binding": "OPENNEXT_DO_WORKER",
      "service": "my-app-do"
    }
  ]
}
```

DO worker wrangler config (wrangler-do.jsonc):
```jsonc
{
  "name": "my-app-do",
  "main": "do-worker.ts",
  "durable_objects": {
    "bindings": [
      { "name": "NEXT_TAG_CACHE_DO_SHARDED", "class_name": "DOShardedTagCache" },
      { "name": "NEXT_CACHE_DO_QUEUE", "class_name": "DOQueueHandler" },
      { "name": "NEXT_CACHE_DO_PURGE", "class_name": "BucketCachePurge" }
    ]
  },
  "migrations": [
    { "tag": "v1", "new_sqlite_classes": ["DOShardedTagCache", "DOQueueHandler", "BucketCachePurge"] }
  ]
}
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/spotty-chefs-float.md
+++ b/.changeset/spotty-chefs-float.md
@@ -1,0 +1,10 @@
+---
+"@opennextjs/cloudflare": minor
+---
+
+feature: add Durable Object proxy for service binding approach
+
+Adds DO proxy utilities under `do-proxy/` that route Durable Object namespace
+operations through a service binding to a separate DO worker via HTTP. This
+enables using Durable Objects while keeping preview URLs (skew protection)
+enabled on the main worker.
--- a/packages/cloudflare/src/api/cloudflare-context.ts
+++ b/packages/cloudflare/src/api/cloudflare-context.ts
@@ -73,6 +73,13 @@ declare global {
 		// The API token to use for the cache purge. It should have the `Cache Purge` permission
 		CACHE_PURGE_API_TOKEN?: string;
 
+		// Service binding to an external DO worker.
+		// When set, proxy DO namespaces are automatically injected for
+		// NEXT_TAG_CACHE_DO_SHARDED, NEXT_CACHE_DO_QUEUE, and NEXT_CACHE_DO_PURGE
+		// if those bindings are not directly available.
+		// This enables using DOs with preview URLs / skew protection.
+		OPENNEXT_DO_WORKER?: Service;
+
 		// The following variables must be provided when skew protection is enabled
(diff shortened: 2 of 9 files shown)


alex-all3dp:
@vicb Any chance you could take a look at this proposed solution soon? I can't judge if the concept makes sense in the grand scheme of things or if it would be possible to add DO support for preview URLs or if the fix is somewhere else but regardless it would be important to being able to combine skew protection with your recommended production setup of using DOs for tag cache and cache purge.

alex-all3dp:
@vicb Could you take a look when you have the time? I'd be happy to help with implementing alternative solutions if you point me in the right direction as well. We would greatly appreciate if we could make the setup with skew protection enabled  more robust and performant by utilizing DOs

alex-all3dp:
Got it to work without the proxy layer after all :)

PR #1126, opened 2026-02-12
fix: prevent Worker hang on HEAD requests to static assets

alex-all3dp:
### Summary
 
When `run_worker_first` is enabled, HEAD requests to static assets hang the Worker because `response.body` is `null` (per HTTP spec) and the fallback `new ReadableStream()` creates a stream that never closes which results in the following error `The Workers runtime canceled this request because it detected that your Worker's code had hung and would never generate a response` for HEAD requests for css and other asset chunks.

I tested this version and all the "worker hung" errors for HEAD requests to css chunks were gone :) 
 
### Changes
  - Return `null` body for HEAD requests instead of falling through to the hanging ReadableStream fallback
  - Add tests for `maybeGetAssetResult` covering GET, HEAD, 404, POST, and `run_worker_first=false` cases

### DIFF
--- a/packages/cloudflare/src/api/overrides/asset-resolver/index.ts
+++ b/packages/cloudflare/src/api/overrides/asset-resolver/index.ts
@@ -53,10 +53,10 @@ const resolver: AssetResolver = {
 
 /**
  * Returns the response body for an asset result.
- *  
+ *
  * HEAD responses must return `null` because `response.body` is `null` per the HTTP spec
  * and the `new ReadableStream()` fallback would create a stream that never closes, hanging the Worker.
- * 
+ *
  * @param method - The HTTP method of the request.
  * @param response - The response from the ASSETS binding.
  * @returns The body to use in the internal result.


packages/cloudflare/src/api/overrides/asset-resolver/index.ts
```diff
@@ -46,7 +46,7 @@ const resolver: AssetResolver = {
 			headers: Object.fromEntries(response.headers.entries()),
 			// Workers and Node types differ.
 			// eslint-disable-next-line @typescript-eslint/no-explicit-any
-			body: response.body || (new ReadableStream() as any),
+			body: method === "HEAD" ? null : response.body || (new ReadableStream() as any), // // HEAD responses have no body; the `new ReadableStream()` fallback would hang the Worker.
```
vicb: Can you please extract that to a function with proper JSDoc?

alex-all3dp: Good idea. Done ✅

alex-all3dp:
> Thanks for the PR.
> 
> I suggested one little refactoring and this should be good to go.

Change pushed :)

packages/cloudflare/src/api/overrides/asset-resolver/index.ts
```diff
@@ -44,14 +44,31 @@ const resolver: AssetResolver = {
...
 };
 
+/**
+ * Returns the response body for an asset result.
+ * HEAD responses must return `null` because `response.body` is `null` per the HTTP spec
+ * and the `new ReadableStream()` fallback would create a stream that never closes, hanging the Worker.
+ * @param method - The HTTP method of the request.
+ * @param response - The response from the ASSETS binding.
+ * @returns The body to use in the internal result.
+ */
```
vicb, line 63:
```suggestion
/**
 * Returns the response body for an asset result.
 *  
 * HEAD responses must return `null` because `response.body` is `null` per the HTTP spec
 * and the `new ReadableStream()` fallback would create a stream that never closes, hanging the Worker.
 * 
 * @param method - The HTTP method of the request.
 * @param response - The response from the ASSETS binding.
 * @returns The body to use in the internal result.
 */
```

vicb (approved):
Awesome, thanks!

alex-all3dp:
@vicb Seems like your commit broke prettier unfortunately?

vicb:
> @vicb Seems like your commit broke prettier unfortunately?

Oops, sorry - fixed
I'll merge when the Ci is green.
Thanks!

PR #1127, opened 2026-02-12
In the `migrate` command, add an initial step to create a Next.js config file (required by open-next) if it doesn't exist

dario-piotrowicz:
Fixes https://jira.cfdata.org/browse/DEVX-2461

___

The next config file is generally present in Next.js apps but it's not actually mandatory (for example see: https://github.com/vercel/examples/tree/main/solutions/blog). 

In any case open-next does require the file (see: https://github.com/opennextjs/opennextjs-aws/issues/1111), so the changes here improve the `migrate` command to create this file if not already present (enabling a migration for projects that don't include the file).

### DIFF
--- a/packages/cloudflare/src/cli/commands/migrate.ts
+++ b/packages/cloudflare/src/cli/commands/migrate.ts
@@ -295,8 +295,8 @@ async function maybeCreateNextConfigFileIfMissing(
 		);
 	}
 
-	if (!skipNextVersionCheck && !ensureNextjsVersionSupported({ nextVersion })) {
-		throw new Error(`Next.js version ${nextVersion} is not compatible with OpenNext.`);
+	if (!skipNextVersionCheck) {
+		await ensureNextjsVersionSupported({ nextVersion });
 	}
 
 	const answer = await askConfirmation("Missing required next.config file. Do you want to create one?");


packages/cloudflare/src/cli/build/build.ts
```diff
@@ -105,7 +105,9 @@ export async function build(
 	logger.info("OpenNext build complete.");
 }
 
-async function ensureNextjsVersionSupported({ nextVersion }: buildHelper.BuildOptions) {
+export async function ensureNextjsVersionSupported({
+	nextVersion,
+}: Pick<buildHelper.BuildOptions, "nextVersion">) {
```
vicb: wrong place?

packages/cloudflare/src/cli/commands/migrate.ts
```diff
@@ -235,6 +244,69 @@ function printStepTitle(title: string): void {
 	logger.info(`⚙️  ${title}...\n`);
 }
 
+/**
+ * Creates a default next.config.ts file with the OpenNext Cloudflare dev initialization.
```
```suggestion
 * Creates a default next.config.ts file
```

packages/cloudflare/src/cli/commands/migrate.ts
```diff
@@ -235,6 +244,69 @@ function printStepTitle(title: string): void {
...
+		nextVersion = getNextVersion(projectDir);
+	} catch {
+		logger.error(
+			"Error: This does not appear to be a Next.js application.\n" +
+				"The 'next' package is not installed and no next.config file was found."
+		);
+		process.exit(1);
+	}
+
+	if (!skipNextVersionCheck && !ensureNextjsVersionSupported({ nextVersion: getNextVersion(projectDir) })) {
```
vicb: l 282
```suggestion
	if (!skipNextVersionCheck && !ensureNextjsVersionSupported({ nextVersion })) {
```

packages/cloudflare/src/cli/build/build.ts
vicb: nvm

packages/cloudflare/src/cli/commands/migrate.ts
```diff
@@ -235,6 +244,69 @@ function printStepTitle(title: string): void {
...
+	// No config file found - check if Next.js is installed
+	let nextVersion: string;
+	try {
+		nextVersion = getNextVersion(projectDir);
+	} catch {
+		logger.error(
+			"Error: This does not appear to be a Next.js application.\n" +
+				"The 'next' package is not installed and no next.config file was found."
+		);
+		process.exit(1);
```
vicb: `ensureNextConfigExists` is a pretty generic name. I would throw here and let the caller decide if they should log the error and/or process.exit. WDYT.

packages/cloudflare/src/cli/commands/migrate.ts
```diff
@@ -235,6 +244,69 @@ function printStepTitle(title: string): void {
...
+	}
+
+	if (!skipNextVersionCheck && !ensureNextjsVersionSupported({ nextVersion: getNextVersion(projectDir) })) {
+		logger.error(`Error: Next.js version ${nextVersion} is not compatible with OpenNext.\n`);
+		process.exit(1);
+	}
+
+	printStepTitle("Creating next.config.ts");
```
vicb: Again I don't think think should be the (single) responsibility of `ensureNextConfigExists` to do that ?

packages/cloudflare/src/cli/commands/migrate.ts
```diff
@@ -27,6 +30,8 @@ async function migrateCommand(args: { forceInstall: boolean }): Promise<void> {
 
 	const projectDir = process.cwd();
 
+	const nextConfigCreated = await ensureNextConfigExists(projectDir, args.forceInstall);
```
vicb: nit: what about `maybeCreateNextConfig` IMO it would be a better fit for the current meaning of the return value. Ensure returning false make me think it's not there?

packages/cloudflare/src/cli/commands/migrate.ts
```diff
@@ -128,18 +133,22 @@ async function migrateCommand(args: { forceInstall: boolean }): Promise<void> {
...
-			nextConfig.path,
-			"import('@opennextjs/cloudflare').then(m => m.initOpenNextCloudflareForDev());\n",
-			{
-				appendIf: (content) => !content.includes("initOpenNextCloudflareForDev"),
-				appendPrefix: "\n",
-			}
-		);
+	// Only update the Next.js config if it wasn't just created
+	// (newly created configs already include initOpenNextCloudflareForDev)
+	if (!nextConfigCreated) {
```
vicb: I don't think that's the best thing to do.

With that logic you have duplication and if you update one location and forget to update the other a mismatch.

Why not removing OpenNext specific stuff from the created next config and always run that logic?

packages/cloudflare/src/cli/build/build.ts
dario-piotrowicz: oops, yeah sorry I wanted to move it somewhere else but forgot! 🙇

packages/cloudflare/src/cli/commands/migrate.ts
vicb: I would also remove "default" - it does not add a lot of value.
And also it's kind of wrong with `@opennextjs/cloudflare` stuff (but I think this should be removed)

packages/cloudflare/src/cli/commands/migrate.ts
dario-piotrowicz: 🤦 👍

packages/cloudflare/src/cli/build/build.ts
dario-piotrowicz: > nvm

why? it was a valid comment 😮

packages/cloudflare/src/cli/commands/migrate.ts
vicb: `ensureNextjsVersionSupported` relies on wrangler being installed, is that the case?

Not sure would added that function (me) but they missed a JSDoc stating that, would you mind taking care of that?

packages/cloudflare/src/cli/build/build.ts
vicb: I thought you added a return value but I was wrong - happy with that change.

dario-piotrowicz: Ah ok, but what I was saying is that with this new `export` we have the migrate.ts file that imports from import `"../build/build.js"` which is not the best, probably this should be moved to a shared location instead.

vicb: Do what you think is best, np 🙏

packages/cloudflare/src/cli/commands/migrate.ts
dario-piotrowicz: Rgarding`process.exit`, I put it here as that seemed to me the general pattern in the open-next codebase, I'm happy to throw instead and have the caller exit 👍

dario-piotrowicz: > ensureNextConfigExists is a pretty generic name

I've updated the name to `createNextConfigFileIfMissing` I hope this is better?

dario-piotrowicz: In regards to `process.exit`, what do you think about
https://github.com/opennextjs/opennextjs-cloudflare/pull/1127/commits/75229f9abe086f1333d15291f7da914fb101aedd
?

packages/cloudflare/src/cli/commands/migrate.ts
dario-piotrowicz: I've renamed the function `createNextConfigFileIfMissing` and removed the `printStepTitle`, does this work for you?

packages/cloudflare/src/cli/commands/migrate.ts
dario-piotrowicz: renamed 👍

packages/cloudflare/src/cli/commands/migrate.ts
dario-piotrowicz: done 👍

packages/cloudflare/src/cli/build/build.ts
dario-piotrowicz: I've moved the function to `src/cli/commands/utils.ts` (https://github.com/opennextjs/opennextjs-cloudflare/pull/1127/commits/3dd0493a1a47c864096378539863a7d9dd41b802)

packages/cloudflare/src/cli/commands/utils.ts
```diff
@@ -33,6 +34,35 @@ export function printHeaders(command: string) {
 	showWarningOnWindows();
 }
 
+/**
+ * Validates that the Next.js version is supported and checks wrangler compatibility.
+ *
```
vicb, line 39: Could you please add a note that wrangler must be installed.

Otherwise l52 will throw

packages/cloudflare/src/cli/commands/migrate.ts
```diff
@@ -130,17 +145,18 @@ async function migrateCommand(args: { forceInstall: boolean }): Promise<void> {
...
-		conditionalAppendFileSync(
-			nextConfig.path,
-			"import('@opennextjs/cloudflare').then(m => m.initOpenNextCloudflareForDev());\n",
-			{
-				appendIf: (content) => !content.includes("initOpenNextCloudflareForDev"),
-				appendPrefix: "\n",
-			}
-		);
-	}
+	// At this point we're sure the next config exists
```
vicb: super nit: we are sure that it has been created... not that it (still) exists. Not asking any change here.

packages/cloudflare/src/cli/commands/migrate.ts
```diff
@@ -235,6 +251,64 @@ function printStepTitle(title: string): void {
...
+		nextVersion = getNextVersion(projectDir);
+	} catch {
+		throw new Error(
+			"This does not appear to be a Next.js application. The 'next' package is not installed and no next.config file was found."
+		);
+	}
+
+	if (!skipNextVersionCheck && !ensureNextjsVersionSupported({ nextVersion })) {
+		throw new Error(`Next.js version ${nextVersion} is not compatible with OpenNext.`);
+	}
```
vicb: Shouldn't this be just:

```suggestion
	if (!skipNextVersionCheck) {
	  ensureNextjsVersionSupported({ nextVersion });
	}
```

packages/cloudflare/src/cli/commands/utils.ts
dario-piotrowicz: added: a4b9f1b 🙂

packages/cloudflare/src/cli/commands/migrate.ts
dario-piotrowicz: ok 😄 

Anyways I've improved this a bit: 23ea438, is this better? 🙂

packages/cloudflare/src/cli/commands/migrate.ts
dario-piotrowicz: of course! 🤦

dario-piotrowicz: Updated: https://github.com/opennextjs/opennextjs-cloudflare/pull/1127/commits/6d0343a5c4b9caa2f63f052102e820583c5d7898 👍

PR #1129, opened 2026-02-15
Add R2 bucket creation as part of the creation of wrangler.jsonc files in the `build` and `migrate` commands

dario-piotrowicz:
The changes here create an R2 bucket for the user when open-next creates a `wrangler.jsonc` file for the project (which can happen in the `build` and `migrate` commands), this will enable users to then proceed with their application's deployment right away.

___

<sub>
A good portion of the changes here have been generated by opus 4.5, I've reviewed and tweaked them accordingly
</sub>

### DIFF
--- a/packages/cloudflare/src/cli/commands/populate-cache.ts
+++ b/packages/cloudflare/src/cli/commands/populate-cache.ts
@@ -269,8 +269,6 @@ async function populateR2IncrementalCache(
 	fs.rmSync(listFile, { force: true });
 
 	if (!result.success) {
-		process.stdout.write(result.stdout);
-		process.stderr.write(result.stderr);
 		logger.error("Wrangler command failed");
 		process.exit(1);
 	}
@@ -339,8 +337,6 @@ async function populateKVIncrementalCache(
 		fs.rmSync(chunkPath, { force: true });
 
 		if (!result.success) {
-			process.stdout.write(result.stdout);
-			process.stderr.write(result.stderr);
 			logger.error("Wrangler command failed");
 			process.exit(1);
 		}
@@ -380,8 +376,6 @@ function populateD1TagCache(
 	);
 
 	if (!result.success) {
-		process.stdout.write(result.stdout);
-		process.stderr.write(result.stderr);
 		logger.error("Wrangler command failed");
 		process.exit(1);
 	}
--- a/packages/cloudflare/src/cli/utils/run-wrangler.ts
+++ b/packages/cloudflare/src/cli/utils/run-wrangler.ts
@@ -104,6 +104,7 @@ export function runWrangler(
 		}
 	);
 
+	const success = result.status === 0;
 	const stdout = result.stdout?.toString() ?? "";
 	const stderr = result.stderr?.toString() ?? "";
 
@@ -112,8 +113,15 @@ export function runWrangler(
 		process.stderr.write(stderr);
 	}
 
+	if (!success) {
+		if (shouldPipeLogs) {
+			process.stdout.write(stdout);
+			process.stderr.write(stderr);
+		}
+	}
+
 	return {
-		success: result.status === 0,
+		success,
 		stdout,
 		stderr,
 	};


packages/cloudflare/src/cli/commands/migrate.ts
```diff
@@ -318,12 +318,19 @@ export function addMigrateCommand<T extends yargs.Argv>(y: T) {
...
+			args
+				.option("forceInstall", {
+					type: "boolean",
+					alias: "f",
+					desc: "Install the dependencies using the `--force` flag.",
+					default: false,
+				})
+				.option("skipConfirmations", {
+					type: "boolean",
+					alias: "s",
```
vicb: We already use s in build to skip next build, better pick something else

packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -268,6 +268,13 @@ async function populateR2IncrementalCache(
 
 	fs.rmSync(listFile, { force: true });
 
+	if (!result.success) {
+		process.stdout.write(result.stdout);
```
vicb: Question: Don't we have an option in run Wrangler to forward std/err?

dario-piotrowicz: When we pipe it (when `logging === 'error'` (https://github.com/opennextjs/opennextjs-cloudflare/blob/63e29f24bd539ed8b915aa18fd0c85378a153c19/packages/cloudflare/src/cli/utils/run-wrangler.ts#L60)) no, at least I don't think so (@james-elicx can correct me if I'm wrong as he worked on this part of the codebase)

I can add an option to also write to the `process.stderr` if you want?

packages/cloudflare/src/cli/commands/migrate.ts
dario-piotrowicz: I've replaced this with `--yes|-y` (https://github.com/opennextjs/opennextjs-cloudflare/pull/1129/commits/411b64d399f2ce0e8cefecae911fb2b24d0dbf03) I hope this works for you

vicb:
Why should we create the bucket in build and migrate rather than deploy and upload? 

I would prefer the second set of commands rather than the first one if possible

dario-piotrowicz:
> I would prefer the second set of commands rather than the first one if possible

`build` and `migrate` include the logic that creates the wrangler config file, that's why this is part of that, how do you suggest that `deploy` and `upload` should be updated?

should `deploy` and `upload` _also_ create the wrangler config file? and should `build` and `migrate` not do that anymore?

It feels quite weird to me to have `deploy` and `upload` do some project settings tweaking... arguably it's also a bit weird to do that in `build`, so I'd be happy to remove the logic from there and only have `migrate` attempt the resources creation

anyways let me know...

vicb:
You can build and migrate for local dev so there should be no need to create a remote bucket a this point. 

The remote bucket is only needed when populating the remote cache, while deploying or uploading a version. That's why I think this is where the creation fits.

At that point ON has access to (and already uses) the config. So that seems like where it should be? 

It's probably best to update the populate cache command that is used by both deploy and upload.

MattieTK:
One of the challenges we have is working out if the user is going to be an R2 user or a KV user based on their account status (see internal discussion).

If we don't do that at this stage we're going to be making an assumption to create a binding and set a cache setting that might not be possible when we come to deploy the service (as its only at that point that we'd find out the user can't use R2).

In this case what would you suggest? Do we move the logic that infers the user's status to this command such that we can migrate to a resource type they have access to, but defer the resource creation? Or do we in the migration command make a optimistic assumption they have access to R2, and then call back to migrate on deploy if they do not with a different codepath?

I think we should treat `migrate` at this stage as being a quite changeable autoconfig specific command, and not worth getting too worried about at this point. It's likely we'll rework this.

dario-piotrowicz:
> You can build and migrate for local dev so there should be no need to create a remote bucket a this point.
> 
> The remote bucket is only needed when populating the remote cache, while deploying or uploading a version. That's why I think this is where the creation fits.
>
> At that point ON has access to (and already uses) the config. So that seems like where it should be?
> 
> It's probably best to update the populate cache command that is used by both deploy and upload.

Your suggestion doesn't take into account the KV fallback and the fact that we create a different config based on whether the user doesn't have R2 enabled.

Are you suggesting to remove the KV fallback logic? or to move it to the `deploy`/`upload` commands? (and they then update the config?)

packages/cloudflare/src/cli/commands/populate-cache.ts
james-elicx: Shouldn't this be being piped already then?
https://github.com/opennextjs/opennextjs-cloudflare/blob/main/packages/cloudflare/src/cli/utils/run-wrangler.ts#L93-L96

james-elicx:
Can we prompt the user to input if they want an R2 or KV cache as part of the `migrate` command and create based on their choice? Or would that introduce too much friction given the intended use case?

MattieTK:
> Can we prompt the user to input if they want an R2 or KV cache as part of the `migrate` command and create based on their choice? Or would that introduce too much friction given the intended use case?

We know that we want to give the user an R2 bucket. We need the user to auth to know this, as only through an error on creation will we know they don't have access. In those cases we would fall back to KV. So it's not really user choice so much as a fallback to not have them churn halfway through the process. We don't really want to give them the choice.

packages/cloudflare/src/cli/commands/populate-cache.ts
dario-piotrowicz: oh yeah you're right! 🙏  open-code removed that, sorry, I put the thing back: https://github.com/opennextjs/opennextjs-cloudflare/pull/1129/commits/a39462df55326aceb7cdd2ee3bc9fc2af9d1d1cf

dario-piotrowicz:
Closing in favour of https://github.com/opennextjs/opennextjs-cloudflare/pull/1132

PR #1132, opened 2026-02-17
Remove R2 from the templates used when creating `wrangler.jsonc` and `open-next.config.ts` files

dario-piotrowicz:
Fixes https://jira.cfdata.org/browse/DEVX-2491

This PR removes R2 from the templates used when creating `wrangler.jsonc` and `open-next.config.ts` files since this generally results on an application that can't actually be deployed.

For more details please see the included changesets.

### DIFF
--- a/.changeset/rotten-foxes-nail.md
+++ b/.changeset/rotten-foxes-nail.md
@@ -1,0 +1,9 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+Remove R2 caching from the `wrangler.jsonc` and `open-next.config.ts` files created by the `migrate` command
+
+The `migrate` command no longer sets up the R2 caching in the `wrangler.jsonc` and `open-next.config.ts` files it creates, this allows newly migrated applications to be deployed right away without forcing the user to enable R2 nor set an R2 bucket.
+
+Ideally applications should use caching for optimal results, so a warning is now also presented at the end of the migration recommending users to set up caching.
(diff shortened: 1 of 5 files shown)


.changeset/rotten-foxes-nail.md
```diff
@@ -0,0 +1,9 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+Remove R2 caching from the `wrangler.jsonc` and `open-next.config.ts` files created by the `migrate` command
+
+The `migrate` command no longer sets up the R2 caching in the `wrangler.jsonc` and `open-next.config.ts` files it creates, this allows newly migrated applications to be deployed right away without forcing the user to enable R2 nor set an R2 bucket.
+
+Ideally applications should use caching for optimal results, so a warning is now also presented at the end of the migration recommending users to set up caching.
```
james-elicx, line 9: should we default to the staticAssetsIncrementalCache in this case?

dario-piotrowicz: we discussed this and we think that it's better not to set any caching at all, this will make it easier to users (or us if we want to provide an automated way to then configure caching) to apply a caching implementation.

If we were to set a different caching solution, like the static or kv ones, then if people later wanted to move onto R2 there'd be more steps to do that, so it seems better to just not set anything at all here (instead of setting something sub-par).

what do you think? 🙂

james-elicx: i mean, i think we should be serving the stuff generated at build-time from static assets for the incremental cache, shouldn't we? i guess it falls back to the dummy one otherwise. up to you /shrug

dario-piotrowicz:
Closing in favour of https://github.com/opennextjs/opennextjs-cloudflare/pull/1133

PR #1139, opened 2026-02-21
fix: dynamically resolve Turbopack external module mappings

james-elicx:
_Disclaimer that this was 100% debugged and patched by Opus 4.6 agents in a different project that I was trying to get running on Workers. I've moved the patch back into this repo and added a regression test._

Some packages that are externalised by turbopack fail to be loaded/resolved on workerd, and esbuild can't statically analyse them properly when bundling from what I understand.

This PR does the following:
- Discovers externalised modules through symlinked packages and trace files.
- Re-maps the imports so that the bundler can statically analyse / handle them.
- Regression test with shiki (which was the library I had an issue with in my own project).

Without this patch:

image


vicb:
Thanks for the PR James!

When I implemented support for OG, I knew it was brittle and would probably need to be handled in a better way.

I'll take a look later this week.

examples/e2e/app-router/package.json
```diff
@@ -21,7 +21,8 @@
 		"@example/shared": "workspace:*",
 		"next": "catalog:e2e",
 		"react": "catalog:e2e",
-		"react-dom": "catalog:e2e"
+		"react-dom": "catalog:e2e",
+		"shiki": "^3.22.0"
```
conico974: I don't think this should be in app-router. Either in a new repo or an existing one that is specific to cloudflare. This issue likely does not happen on aws and app-router should be kept synced between both repos

conico974 (commented):
LGTM but I think the test is in the wrong place

.changeset/five-gifts-hope.md
```diff
@@ -0,0 +1,13 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+Fix Turbopack external module resolution on workerd by dynamically discovering external imports at build time.
+
+When packages are listed in `serverExternalPackages`, Turbopack externalizes them via `externalImport()` which uses dynamic `await import(id)`. On workerd, the bundler can't statically analyze `import(id)` with a variable, so these modules aren't included in the worker bundle.
```
vicb: "On workerd" is probably not the most accurate here. It is more ESBuild (the build system) that worked

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
```diff
@@ -10,6 +13,189 @@ fix:
...
+/**
+ * Discover Turbopack external module mappings by reading symlinks in .next/node_modules/.
+ *
+ * Turbopack externalizes packages listed in serverExternalPackages and creates hashed
+ * identifiers (e.g. "shiki-43d062b67f27bbdc") with symlinks in .next/node_modules/ pointing
+ * to the real packages (e.g. ../../node_modules/shiki). At runtime, externalImport() does
+ * `await import("shiki-43d062b67f27bbdc/wasm")` which fails on workerd because those hashed
+ * names are not real modules. This function discovers the mappings so we can intercept them.
+ */
+function discoverExternalModuleMappings(filePath: string): Map<string, string> {
```
vicb, line 32: Could you please add JSDoc for `filepath` and the return value

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
```diff
@@ -10,6 +13,189 @@ fix:
...
+	const dotNextDir = filePath.replace(/\/server\/chunks\/.*$/, "");
+	const nodeModulesDir = path.join(dotNextDir, "node_modules");
+
+	const mappings = new Map<string, string>();
+
+	if (!fs.existsSync(nodeModulesDir)) {
+		return mappings;
+	}
+
+	for (const entry of fs.readdirSync(nodeModulesDir)) {
```
vicb: You can use `withFileTypes` to avoid having to `lStatSync`

ref: https://nodejs.org/api/fs.html#fsreaddirsyncpath-options

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
```diff
@@ -10,6 +13,189 @@ fix:
   requireChunk(chunkPath)
 `;
 
+/**
+ * Discover Turbopack external module mappings by reading symlinks in .next/node_modules/.
+ *
+ * Turbopack externalizes packages listed in serverExternalPackages and creates hashed
+ * identifiers (e.g. "shiki-43d062b67f27bbdc") with symlinks in .next/node_modules/ pointing
```
vicb, line 20: Thought while reading this JSDoc: can this be implemented as an ESBuild plugin?
i.e. when we try to resolve `shiki-43d062b67f27bbdc`, we would replace with the actual path

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
```diff
@@ -10,6 +13,189 @@ fix:
...
+ */
+function buildExternalImportRule(
+	mappings: Map<string, string>,
+	tracedFiles: string[],
+	runtimeCode: string
+): string {
+	const cases: string[] = [];
+
+	// Always include the @vercel/og rewrite
+	cases.push(`    case "next/dist/compiled/@vercel/og/index.node.js":
```
vicb: nit: remove the leading spaces for the `case`s and `.map` at the end if they are really needed, adding a comment about why

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
```diff
@@ -10,6 +13,189 @@ fix:
...
+	for (const [hashedSubpath, realSubpath] of subpathCases) {
+		cases.push(`    case "${hashedSubpath}":
+      $RAW = await import("${realSubpath}");
+      break;`);
+	}
+
+	// Discover bare external imports from chunk files (e.g. externalImport("shiki")).
+	// These need explicit switch cases so the bundler can statically resolve them.
+	const bareImports = discoverBareExternalImports(tracedFiles, runtimeCode);
+	const alreadyCased = new Set(cases.map((c) => c.match(/case "([^"]+)"/)?.[1]).filter(Boolean));
```
vicb: wouldn't it be simpler to accumulate the `case`s added earlier rather than trying to extract the values with a regexp here?

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
```diff
@@ -10,6 +13,189 @@ fix:
...
+      $RAW = await import($ID);
+  }
+`;
+}
+
+/**
+ * Scan traced chunk files for bare external module imports (e.g. `externalImport("shiki")`).
+ *
+ * In some Turbopack versions, externalized packages are referenced by their real names
+ * (not hashed). On workerd, the default `await import(id)` with a variable `id` can't be
```
vicb: Same as earlier, this has nothing do do with workerd - only with the build/bundle step

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
```diff
@@ -10,6 +13,189 @@ fix:
...
+ * statically analyzed by the bundler. By adding explicit switch cases with string literals,
+ * we make these imports statically discoverable so they get bundled into the worker.
+ */
+function discoverBareExternalImports(tracedFiles: string[], runtimeCode: string): Map<string, string> {
+	const bareImports = new Map<string, string>();
+
+	// Turbopack assigns `externalImport` to a single-letter property on the context prototype,
+	// e.g. `contextPrototype.y = externalImport`. The property name could change between versions,
```
vicb, line 143: is "single-letter" always the case?
The code says `\w+`

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
```diff
@@ -10,6 +13,189 @@ fix:
...
+ * Only scans files with "[externals]" in the name since those are the chunks that
+ * contain externalImport calls.
+ */
+function discoverExternalSubpaths(mappings: Map<string, string>, tracedFiles: string[]): Map<string, string> {
+	const subpaths = new Map<string, string>();
+
+	const externalChunks = tracedFiles.filter((f) => f.includes("[externals]"));
+
+	for (const [hashedName, realName] of mappings) {
+		const pattern = new RegExp(`"(${hashedName.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")}/[^"]*)"`, "g");
```
vicb: Can you expand on what this is doing and why?

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
james-elicx: I have a feeling it might not solve all the cases where this can happen with dynamic imports

gavinying:
is it planned to be merged? thx.

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
```diff
@@ -10,6 +13,199 @@ fix:
...
+ * Discover Turbopack external module mappings by reading symlinks in .next/node_modules/.
+ *
+ * Turbopack externalizes packages listed in serverExternalPackages and creates hashed
+ * identifiers (e.g. "shiki-43d062b67f27bbdc") with symlinks in .next/node_modules/ pointing
+ * to the real packages (e.g. ../../node_modules/shiki). At runtime, externalImport() does
+ * `await import("shiki-43d062b67f27bbdc/wasm")` which fails because the bundler can't
+ * statically analyze those hashed names. This function discovers the mappings so we can
+ * generate explicit switch cases for the bundler.
+ *
+ * @param filePath Absolute path to the Turbopack runtime file (e.g. `.next/server/chunks/ssr/[turbopack]_runtime.js`),
```
vicb: `.next/server/chunks/ssr/[turbopack]_runtime.js` does not look like an absolute path. Could we pass the build options instead?

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
```diff
@@ -10,6 +13,199 @@ fix:
...
+
+	// Discover bare external imports from chunk files (e.g. externalImport("shiki")).
+	// These need explicit switch cases so the bundler can statically resolve them.
+	const bareImports = discoverBareExternalImports(tracedFiles, runtimeCode);
+	for (const [moduleName, realName] of bareImports) {
+		addCase(moduleName, realName);
+	}
+
+	// Indent each case line by 4 spaces to align with the switch body in the YAML fix block.
+	const indentedCases = cases.map((c) => c.replace(/^/gm, "    ")).join("\n");
```
vicb: Is there a need for the regexp here or should all lines be updated?

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
```diff
@@ -10,6 +13,199 @@ fix:
...
+	const bareImports = new Map<string, string>();
+
+	// Turbopack assigns `externalImport` to a property on the context prototype,
+	// e.g. `contextPrototype.y = externalImport`. The property name could change between versions,
+	// so we extract it dynamically from the runtime code rather than hardcoding it.
+	const propMatch = runtimeCode.match(/contextPrototype\.(\w+)\s*=\s*externalImport/);
+	if (!propMatch?.[1]) {
+		return bareImports;
+	}
+	const prop = propMatch[1];
```
vicb: nit: prop -> externalImportAlias ?

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
```diff
@@ -10,6 +13,199 @@ fix:
...
+
+	// Chunks call externalImport as e.g. `.y("shiki")` — build a regex using the discovered property name.
+	const callPattern = new RegExp(`\\.${prop}\\("([^"]+)"\\)`, "g");
```
vicb: nit: call -> externalImportRegexp

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
```diff
@@ -10,6 +13,199 @@ fix:
...
+
+	const chunkFiles = tracedFiles.filter((f) => f.includes(".next/server/chunks/"));
+
+	for (const filePath of chunkFiles) {
+		try {
+			const content = fs.readFileSync(filePath, "utf-8");
+			for (const match of content.matchAll(callPattern)) {
```
vicb: nit: match -> extenlImportMatch

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
```diff
@@ -10,6 +13,199 @@ fix:
...
+	const subpaths = new Map<string, string>();
+
+	const externalChunks = tracedFiles.filter((f) => f.includes("[externals]"));
+
+	for (const [hashedName, realName] of mappings) {
+		// Build a regex to find quoted subpath imports of this hashed module in chunk source code.
+		// E.g. for hashedName "shiki-43d062b67f27bbdc", this matches strings like
+		// "shiki-43d062b67f27bbdc/wasm" or "shiki-43d062b67f27bbdc/engine/javascript".
+		// The hashedName is escaped to safely use it as a literal in the regex pattern.
+		const escaped = hashedName.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
```
vicb: nit: don't we have an util for that? (IIRC it's coming to latest v8)

james-elicx: I believe we still support Node v22, and it looks like `RegExp.escape(...)` is only available from v24 onwards according to mdn

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
james-elicx: replaced with split -> map -> join

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
james-elicx: It's an absolute path

https://github.com/opennextjs/opennextjs-aws/blob/4bdfd292f985913f3298eb3f30a7790b877e0464/packages/open-next/src/build/copyTracedFiles.ts#L151
https://github.com/opennextjs/opennextjs-aws/blob/4bdfd292f985913f3298eb3f30a7790b877e0464/packages/open-next/src/build/copyTracedFiles.ts#L221

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
vicb: Iirc we have a util we use for i.e. path escaping (in aws)?

vicb:
Thanks for the updates James, I'll review tomorrow

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
james-elicx: Oh sorry I thought you meant a runtime util when I read V8. My bad. There's the cross platform path one that handles escaping.

vicb (approved):
LGTM, thanks James!

PR #1142, opened 2026-02-23
Use the c3 <COMPATIBILITY_DATE> placeholder in the template

vicb:
Support was added in c3 in [this PR](https://github.com/cloudflare/workers-sdk/pull/11520).

It will make things consistent with:
- other placeholders in the same file
- placeholders in `packages/cloudflare/templates/wrangler.jsonc`

Tested on a local build of c3 pointing to this branch


vicb:
Thanks for the review James!

PR #1145, opened 2026-02-24
fix: new Request(request, init) crashing in route handlers on workerd          

### LINKED ISSUE #1037: [BUG] Expo Authorization Proxy returns 400 instead of redirect when deployed on Cloudflare via opennextjs-cloudflare
### Describe the bug

Hi, thanks for this project!

I’m deploying a Next.js 15 + better-auth + expo project using @opennextjs/cloudflare. Everything works fine when accessing /api/auth/... from a web client, including login/logout/check.

However, when using Expo together with better-auth/expo-plugin, Expo tries to request:

/api/auth/expo-authorization-proxy?authorizationURL=...

and Cloudflare always returns:

{
  "code": "UNEXPECTED_ERROR",
  "message": "Unexpected error"
}

instead of the expected 302 redirect.

What I tested

❌ production (Cloudflare)
fails with 400

❌localhost run command： opennextjs-cloudflare preview
same behavior as production (400)

✅ next dev
works perfectly (302 redirect)

This means the issue only happens on the Cloudflare adapter runtime, not in Next.js dev or better-auth itself.

Expected behavior

/api/auth/expo-authorization-proxy should perform a redirect (302), same as in development.

Actual behavior

400 Unexpected Error inside Cloudflare runtime

Versions

"@opennextjs/cloudflare": "^1.14.4"
"next": "15.5.7"
"better-auth": "^1.4.5"

Extra note

This only happens with Expo authorization plugin integration. Other authentication routes behave normally.

Thanks! (The redirect ghosts seem to vanish only in dev mode 😅)

### Steps to reproduce

Image

Image

### Expected behavior

Image

1.14.4

### Wrangler version

4.53.0

### next info output

```shell
pnpm next info
npm warn Unknown env config "verify-deps-before-run". This will stop working in the next major version of npm.
npm warn Unknown env config "verify-deps-before-run". This will stop working in the next major version of npm.
This project is configured to use pnpm because /Users/imalec/workspaces/resolvai/resolvai-auth/package.json has a "packageManager" field
This project is configured to use pnpm because /Users/imalec/workspaces/resolvai/resolvai-auth/package.json has a "packageManager" field

Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.1.0: Mon Oct 20 19:34:05 PDT 2025; root:xnu-12377.41.6~2/RELEASE_ARM64_T6041
  Available memory (MB): 65536
  Available CPU cores: 14
Binaries:
  Node: 22.17.0
  npm: 11.6.0
  Yarn: N/A
  pnpm: 10.25.0
Relevant Packages:
  next: 15.5.7 // An outdated version detected (latest is 16.0.8), upgrade is highly recommended!
  eslint-config-next: 15.5.4
  react: 19.2.1
  react-dom: 19.2.1
  typescript: 5.9.3
Next.js Config:
  output: standalone
 ⚠ An outdated version detected (latest is 16.0.8), upgrade is highly recommended!
```

### Additional context

_No response_

### LINKED ISSUE #1144: [BUG] new Request(request, init) crashes in route handlers with Invalid URL: [object Request]
### Description

OpenNext replaces `globalThis.Request` with `CustomRequest`, which passes `input` straight through to workerd's native `Request` via `super(input, init)`.

workerd's `Request` constructor only accepts a string, URL, or Request instance. If it gets a plain object, it coerces it to a string (`"[object Object]"`), tries to parse that as a URL, and crashes. Node.js is more lenient and reads `.url` from plain objects, which is why this is workerd-specific.

Any code that does `new Request({ url, method, headers, body })` triggers the crash (e.g. third-party middleware like Clerk).

### Reproduction

The bug in two lines (works in Node.js, crashes in workerd):

```ts
const obj = { url: "https://example.com", method: "GET", headers: {}, body: null };
new Request(obj); // TypeError: Invalid URL: [object Object]
```

<details>
<summary>Full runnable worker</summary>

`worker.js`:
```js
export default {
  async fetch(request) {
    const obj = { url: request.url, method: "GET", headers: {}, body: null };
    return new Response(new Request(obj).url);
  },
};
```

`wrangler.jsonc`:
```jsonc
{ "name": "repro", "main": "worker.js", "compatibility_date": "2024-09-23" }
```

```bash
npx wrangler dev && curl http://localhost:8787/
# → TypeError: Invalid URL: [object Object]
```
</details>

### Fix

Before calling `super()`, check what `input` is. Strings, URLs, and real Request instances pass through directly. For plain objects, extract `.url` and merge `method`/`headers`/`body` into init so properties aren't lost:

```ts
const OriginalRequest = globalThis.Request;
const CustomRequest = class extends OriginalRequest {
  constructor(input: RequestInfo | URL, init?: RequestInit) {
    // ... existing cache/body fixes ...
    if (typeof input === "string" || input instanceof URL || input instanceof OriginalRequest) {
      super(input, init);
    } else {
      const req = input as unknown as Request;
      const merged = {
        method: req.method,
        headers: req.headers,
        body: req.body,
        ...(req.body ? { duplex: "half" as const } : {}),
        ...init,
      };
      super(req.url, merged as RequestInit);
    }
  }
};
```

### Environment

- `@opennextjs/cloudflare`: 1.17.0
- Cloudflare Workers runtime (workerd)

ama2369:
## Summary

Fixes #1144. Likely also fixes #1037.

`new Request(request, { headers })` in route handlers crashes with `TypeError: Invalid URL: [object Request]` because the `request` object isn't recognized as a native `Request` by workerd's constructor.

The fix captures `OriginalRequest` before overwriting `globalThis.Request`, adds an `instanceof` guard, and extracts `.url` with merged properties for non-native objects.

### What changed

`packages/cloudflare/src/cli/templates/init.ts`

- Capture `OriginalRequest` before replacement for reliable `instanceof` checks
- String, URL, and `OriginalRequest` instances pass through to `super()` directly
- Non-native objects: extract `.url`, merge `method`/`headers`/`body` into init, set `duplex: "half"` when body is present

## Test plan

- [ ] String and URL inputs still work
- [ ] `new Request(request, { headers })` works in route handlers (the bug)
- [ ] Method, headers, and body are preserved through the non-native path
- [ ] `stream.Readable` body conversion still works (existing behavior)

### DIFF
--- a/.changeset/fix-custom-request-constructor.md
+++ b/.changeset/fix-custom-request-constructor.md
@@ -1,0 +1,7 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+fix: handle plain Request-like objects in `CustomRequest` constructor
+
+workerd's `Request` constructor coerces plain objects to `"[object Object]"` instead of reading `.url`, causing `TypeError: Invalid URL`. The fix extracts `.url` and merges method/headers/body from plain objects before passing to `super()`.
--- a/packages/cloudflare/src/cli/templates/init.ts
+++ b/packages/cloudflare/src/cli/templates/init.ts
@@ -77,7 +77,10 @@ function initRuntime() {
 		return __original_fetch(input, init);
 	};
 
-	const CustomRequest = class extends globalThis.Request {
+	// workerd calls toString() on plain objects instead of reading .url,
+	// so capture the native constructor for instanceof checks.
+	const OriginalRequest = globalThis.Request;
+	const CustomRequest = class extends OriginalRequest {
 		constructor(input: RequestInfo | URL, init?: RequestInit) {
 			if (init) {
 				delete (init as { cache: unknown }).cache;
@@ -88,7 +91,21 @@ function initRuntime() {
 					value: init.body instanceof stream.Readable ? ReadableStream.from(init.body) : init.body,
 				});
 			}
-			super(input, init);
+			if (typeof input === "string" || input instanceof URL || input instanceof OriginalRequest) {
+				super(input, init);
+			} else {
+				// Plain Request like object from third-party middleware or edge adapters
+				// extract .url and merge properties so they aren't lost.
+				const req = input as unknown as Request;
+				const merged = {
+					method: req.method,
+					headers: req.headers,
+					body: req.body,
+					...(req.body ? { duplex: "half" as const } : {}),
+					...init,
+				};
+				super(req.url, merged as RequestInit);
+			}
 		}
 	};
 


vicb:
I don't believe the fix here is correct because we would deviate from Node.js behavior.

I do think however that it is a bug in the workerd runtime.
I'll create a GH issue there and link that to the original issue #1144

PR #1146, opened 2026-02-24
refactor: creation of config files

vicb:
Delete `packages/cloudflare/src/cli/utils/create-config-files.ts` to inline the 2 functions that were only called once each.

`packages/cloudflare/src/cli/utils/` already has a `create-wrangler-config.ts` and a `create-open-next-config`.ts

Also add test for both config creation.


dario-piotrowicz (approved):

vicb:
Thanks for the review Dario!

PR #1151, opened 2026-02-25
fix(patches): return empty object for unhandled manifests in loadManifest

### LINKED ISSUE #1141: [BUG] Unexpected loadManifest calls for new manifest files introduced in Next.js 16.2.0-canary
### Describe the bug

When building and previewing a Next.js app with next@canary (16.2.0-canary.53+) using @opennextjs/cloudflare, all dynamic
  routes return 500 Internal Server Error at runtime. The error is thrown by the CF Workers runtime:

  Error: Unexpected loadManifest(/.next/server/subresource-integrity-manifest.json) call!
  Error: Unexpected loadManifest(/.next/server/app/api/[route]/react-loadable-manifest.json) call!

  These manifest files were introduced in Next.js 16.2.0-canary and are not yet handled by @opennextjs/cloudflare, causing
  the worker to crash on any request that triggers loadManifests().

  **Environment**

  - @opennextjs/cloudflare: 1.16.5
  - next: 16.2.0-canary.53
  - wrangler: 4.67.0

### Steps to reproduce

 1. Create a Next.js app router project:
  pnpm create next-app my-app --app
  cd my-app
  2. Add @opennextjs/cloudflare and upgrade Next.js to canary:
  pnpm add @opennextjs/cloudflare
  pnpm add next@canary
  3. Create an open-next.config.ts and wrangler.jsonc as per the OpenNext Cloudflare docs.
  4. Build and preview:
  pnpm opennextjs-cloudflare build
  pnpm opennextjs-cloudflare preview
  5. Navigate to any dynamic/SSR route — you'll receive a 500 with the following in wrangler logs:
  ✘ [ERROR] ⨯ Error: Unexpected loadManifest(/.next/server/subresource-integrity-manifest.json) call!

### Expected behavior

Dynamic routes should render successfully. The new manifest files introduced in the canary
  (subresource-integrity-manifest.json, per-route react-loadable-manifest.json) should either be bundled into the worker or
  gracefully handled when not present.

1.16.5

### Wrangler version

4.67.0

### next info output

```shell
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.5.0: Tue Apr 22 19:54:49 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T6000
  Available memory (MB): 32768
  Available CPU cores: 10
Binaries:
  Node: 20.19.2
  npm: 10.8.2
  Yarn: 1.22.22
  pnpm: 9.15.9
Relevant Packages:
  next: 16.2.0-canary.58 // Latest available version is detected (16.2.0-canary.58).
  eslint-config-next: N/A
  react: 19.1.0
  react-dom: 19.1.0
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

### Additional context

_No response_

nathanschram:
## Summary
- Return `{}` instead of throwing for manifest paths not found during the build-time glob scan in `loadManifest()` and `evalManifest()`
- Handles Next.js 16.2.0-canary.53+ which introduces new `loadManifest()` calls for optional/phase-dependent manifests

## Problem
Next.js canary adds `loadManifest()` calls for:
1. `subresource-integrity-manifest.json` - only generated when `experimental.sri` is configured, but loaded unconditionally
2. Per-route `react-loadable-manifest.json` - generated by Turbopack, possibly in a different build phase

The adapter's build-time glob scan doesn't find these files, so the patched function throws at runtime, crashing all dynamic routes with 500.

## Fix
Replace the `throw new Error('Unexpected loadManifest...')` fallback with `return {}`. This follows the same defensive pattern used by other adapter plugins (`instrumentation.ts`, `find-dir.ts`) that return safe defaults for missing files.

## Test plan
- [ ] Build a Next.js app with `next@canary` (16.2.0-canary.53+) using `@opennextjs/cloudflare`
- [ ] Verify dynamic/SSR routes no longer return 500
- [ ] Verify existing manifests still load correctly (app-manifest, build-manifest, etc.)
- [ ] Verify SRI works correctly when `experimental.sri` IS configured

Fixes #1141

### DIFF
--- a/packages/cloudflare/src/cli/build/patches/plugins/load-manifest.ts
+++ b/packages/cloudflare/src/cli/build/patches/plugins/load-manifest.ts
@@ -69,12 +69,19 @@ function loadManifest($PATH, $$$ARGS) {
   // Known optional manifests \u2014 Next.js loads these with handleMissing: true
   // (see vercel/next.js packages/next/src/server/route-modules/route-module.ts).
   // Return {} to match Next.js behaviour instead of crashing the worker.
-  if ($PATH.endsWith("react-loadable-manifest.json") ||
-      $PATH.endsWith("subresource-integrity-manifest.json") ||
-      $PATH.endsWith("server-reference-manifest.json") ||
-      $PATH.endsWith("dynamic-css-manifest.json") ||
-      $PATH.endsWith("fallback-build-manifest.json")) {
-    return {};
+  // Note: Some manifest constants in Next.js omit the .json extension
+  // (e.g. SUBRESOURCE_INTEGRITY_MANIFEST, DYNAMIC_CSS_MANIFEST), so we
+  // strip .json before matching to handle both forms.
+  {
+    const p = $PATH.replace(/\\.json$/, "");
+    if (p.endsWith("react-loadable-manifest") ||
+        p.endsWith("subresource-integrity-manifest") ||
+        p.endsWith("server-reference-manifest") ||
+        p.endsWith("dynamic-css-manifest") ||
+        p.endsWith("fallback-build-manifest") ||
+        p.endsWith("prefetch-hints")) {
+      return {};
+    }
   }
   throw new Error(\`Unexpected loadManifest(\${$PATH}) call!\`);
 }`,
--- a/.changeset/fix-load-manifest-graceful-fallback.md
+++ b/.changeset/fix-load-manifest-graceful-fallback.md
@@ -10,12 +10,17 @@ generated, so the patched function threw at runtime, crashing dynamic routes wit
 
 Instead of a blanket catch-all, handle only the specific optional manifests from Next.js
 `route-module.ts`:
-- `react-loadable-manifest.json` (Turbopack per-route, not all routes have dynamic imports)
-- `subresource-integrity-manifest.json` (only when `experimental.sri` configured)
-- `server-reference-manifest.json` (App Router only)
-- `dynamic-css-manifest.json` (Pages Router + Webpack only)
-- `fallback-build-manifest.json` (only for `/_error` page)
-- `_client-reference-manifest.js` (optional for static metadata routes)
+
+- `react-loadable-manifest` (Turbopack per-route, not all routes have dynamic imports)
+- `subresource-integrity-manifest` (only when `experimental.sri` configured)
+- `server-reference-manifest` (App Router only)
+- `dynamic-css-manifest` (Pages Router + Webpack only)
+- `fallback-build-manifest` (only for `/_error` page)
+- `prefetch-hints` (new in Next.js 16.2)
+- `_client-reference-manifest.js` (optional for static metadata routes, evalManifest)
+
+Manifest matching strips `.json` before comparison since some Next.js constants omit
+the extension (`SUBRESOURCE_INTEGRITY_MANIFEST`, `DYNAMIC_CSS_MANIFEST`, etc.).
 
 Unknown manifests still throw to surface genuine errors.
 


nathanschram:
Hey @vicb - just wanted to check in on this. No rush at all, I know you're probably busy. Just wanted to make sure the approach looks reasonable to you, or if you'd prefer I take it in a different direction. Happy to adjust anything.

Rovak:
Can confirm that is working for me in production after copying this patch into the project using `pnpm patch`

bonadio:
Need this fix too. 
Thanks

sommeeeer:
LGTM, until this is merged you can use `npm i https://pkg.pr.new/@opennextjs/cloudflare@1151`.

packages/cloudflare/src/cli/build/patches/plugins/load-manifest.ts
```diff
@@ -64,9 +64,13 @@ function loadManifest($PATH, $$$ARGS) {
...
     return process.env.NEXT_BUILD_ID;
-	}
+\t}
   ${returnManifests}
-  throw new Error(\`Unexpected loadManifest(\${$PATH}) call!\`);
+  // Return an empty object for manifests not found during the build-time scan.
+  // This handles optional manifests (e.g. subresource-integrity-manifest.json when
+  // experimental.sri is not configured) and manifests generated in a different build
+  // phase (e.g. per-route react-loadable-manifest.json with Turbopack).
+  return {};
```
conico974: I don't think this is the right solution, this will just hide other potential errors, if we know the one that could be null, that's the one we should return an empty object, not everything

packages/cloudflare/src/cli/build/patches/plugins/load-manifest.ts
```diff
@@ -110,7 +114,7 @@ function evalManifest($PATH, $$$ARGS) {
 function evalManifest($PATH, $$$ARGS) {
   $PATH = $PATH.replaceAll(${JSON.stringify(sep)}, ${JSON.stringify(posix.sep)});
   ${returnManifests}
-  throw new Error(\`Unexpected evalManifest(\${$PATH}) call!\`);
+  return {};
```
conico974: Same here

packages/cloudflare/src/cli/build/patches/plugins/load-manifest.ts
nathanschram: Good call - updated to handle only the specific manifests that Next.js loads with `handleMissing: true` (traced through `route-module.ts` in vercel/next.js).

The five optional manifests for `loadManifest`:
- `react-loadable-manifest.json` - Turbopack per-route, not all routes have dynamic imports
- `subresource-integrity-manifest.json` - only when `experimental.sri` configured
- `server-reference-manifest.json` - App Router only, Pages Router never generates
- `dynamic-css-manifest.json` - Pages Router + Webpack only
- `fallback-build-manifest.json` - only for `/_error` page

Everything else still throws. Same approach for `evalManifest` - only `_client-reference-manifest.js` gets a safe default (`{ __RSC_MANIFEST: {} }`) since it's optional for static metadata routes.

packages/cloudflare/src/cli/build/patches/plugins/load-manifest.ts
nathanschram: Addressed - `evalManifest` now only handles `_client-reference-manifest.js` (optional for static metadata routes per `route-module.ts` line 343). Returns `{ __RSC_MANIFEST: {} }` as the safe default structure. Everything else still throws.

bonadio:
> LGTM, until this is merged you can use `npm i https://pkg.pr.new/@opennextjs/cloudflare@1151`.

This did not work for me, trying to run next 16.2 still receives "Unexpected loadManifest(/.next/server/prefetch-hints.json) call!"  running with wrangler dev locally

matthewvolk:
Hey @nathanschram — we've been testing Next.js 16.2.0 (stable, not canary) on Cloudflare Workers and can confirm this PR is needed alongside #1160 for things to work.

We found one issue during testing (after manually applying the patch from this PR locally): **some manifests are requested without the `.json` extension**, so the `$PATH.endsWith("...-manifest.json")` checks don't match them.

### Root cause

Next.js defines some manifest constants without `.json`:
- `SUBRESOURCE_INTEGRITY_MANIFEST = "subresource-integrity-manifest"` ([source](https://github.com/vercel/next.js/blob/canary/packages/next/src/shared/lib/constants.ts))
- `DYNAMIC_CSS_MANIFEST = "dynamic-css-manifest"` (same file)

When `loadManifest()` is called with these constants, the path ends up as e.g. `/.next/server/dynamic-css-manifest` — no `.json` extension.

### My suggestion for this PR

For each manifest in the fallback list, check both with and without `.json`:

```typescript
$PATH.endsWith("dynamic-css-manifest.json") || $PATH.endsWith("dynamic-css-manifest")
```

Or strip `.json` from the path before matching:

```typescript
const normalizedPath = $PATH.replace(/\.json$/, "");
if (normalizedPath.endsWith("dynamic-css-manifest")) return {};
```

After manually applying (locally) #1160 + this PR + the extensionless fix above, I can confirm that our Next.js 16.2.0 app renders successfully on Cloudflare Workers (for Platforms, in our case):

next-16-2-0-running-after-manual-patches

Happy to help with the fix or open a follow-up PR if that's easier. Thanks for this PR!

conico974:
@nathanschram Could You fix the prettier issue please ?

nathanschram:
@conico974 Prettier issue should be fixed now - sorry about that.

@matthewvolk Great catch on the extensionless manifests - I've updated the matching to strip `.json` before comparison, so it now handles both `dynamic-css-manifest` and `dynamic-css-manifest.json`. Verified against the Next.js constants in `shared/lib/constants.ts` where `SUBRESOURCE_INTEGRITY_MANIFEST`, `DYNAMIC_CSS_MANIFEST`, and `SERVER_REFERENCE_MANIFEST` are all defined without the extension.

@bonadio Added `prefetch-hints` to the optional manifest list as well - it's a new constant (`PREFETCH_HINTS = 'prefetch-hints.json'`) in Next.js 16.2.

Summary of changes in the latest push:
- Prettier fix on changeset
- Extensionless manifest matching (strips `.json` before checking)
- `prefetch-hints` added to the optional list
- All 236 tests pass, tsc clean

PR #1189, opened 2026-04-10
fix(tag-cache): forward `isStale()` in `withFilter`

### LINKED ISSUE #1187: withFilter does not forward isStale(), breaking SWR support
## Bug Description

The `withFilter` utility in `tag-cache-filter.js` does not forward the `isStale()` method from the underlying tag cache implementation. This means SWR (stale-while-revalidate) support added in v1.19.0 is completely broken when using `withFilter`.

## How it breaks

`withFilter` returns an object with only these methods:
- `getLastRevalidated` ✓
- `hasBeenRevalidated` ✓
- `writeTags` ✓
- `getPathsByTags` ✓
- **`isStale` ✗ — missing**

When Next.js 16 calls `globalThis.tagCache.isStale?.()`, it gets `undefined` because the filtered wrapper doesn't have that method:

```js
// utils/cache.js
await globalThis.tagCache.isStale?.(tags, lastModified) ?? false
//                        ^^^^^^^^ undefined when using withFilter → always false
```

This means `isStale` always returns `false`, so entries are never marked as "stale but serveable." Instead of SWR behavior (serve stale immediately, revalidate in background), every revalidated entry is treated as expired and must be re-fetched synchronously.

## Suggested Fix

Add `isStale` forwarding to `withFilter` in `src/api/overrides/tag-cache/tag-cache-filter.ts`:

```ts
export function withFilter({ tagCache, filterFn }) {
    return {
        // ... existing methods ...
        isStale: tagCache.isStale
            ? async (tags, lastModified) => {
                const filteredTags = tags.filter(filterFn);
                if (filteredTags.length === 0) {
                    return false;
                }
                return tagCache.isStale(filteredTags, lastModified);
            }
            : undefined,
    };
}
```

## Workaround

Replace `withFilter` with a custom wrapper that forwards all methods including `isStale`:

```ts
const inner = doShardedTagCache({ ... })
const filter = (tags) => tags.filter(softTagFilter)
const tagCache = {
    ...inner,
    name: `filtered-${inner.name}`,
    getLastRevalidated: (tags) => { /* filter + delegate */ },
    hasBeenRevalidated: (tags, lm) => { /* filter + delegate */ },
    isStale: (tags, lm) => { /* filter + delegate */ },
    writeTags: (tags) => { /* filter + delegate */ },
}
```

## Environment

- `@opennextjs/cloudflare`: 1.19.0
- `next`: 16.2.3
- Tag cache: `doShardedTagCache` with `softTagFilter`

fatihy101:
## Summary

The `withFilter` utility does not forward the optional `isStale()` method from the underlying tag cache implementation. This breaks SWR (stale-while-revalidate) support added in v1.19.0.

When using `withFilter`, `globalThis.tagCache.isStale?.()` always returns `undefined` → `false`, causing every revalidated entry to be treated as expired (immediate re-fetch) instead of stale (serve stale, revalidate in background).

## Changes

- Forward `isStale()` in `withFilter`, following the same conditional pattern used for `getPathsByTags` (only create the wrapper when the underlying tag cache provides it)
- Add tests for `isStale` filtering, empty tag short-circuit, and undefined forwarding

## How it was found

After upgrading to v1.19.0 and using `withFilter` + `softTagFilter` with `doShardedTagCache`, we observed that SWR never triggered. Tracing through the code revealed that `withFilter`'s return object omits `isStale`, so the optional chain in `@opennextjs/aws/utils/cache.js` always falls back to `false`.

Closes #1187

### DIFF
--- a/.changeset/major-actors-make.md
+++ b/.changeset/major-actors-make.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+fix(tag-cache): forward `isStale()` in `withFilter`
--- a/packages/cloudflare/src/api/overrides/tag-cache/tag-cache-filter.spec.ts
+++ b/packages/cloudflare/src/api/overrides/tag-cache/tag-cache-filter.spec.ts
@@ -10,6 +10,7 @@ const mockedTagCache = {
 	getPathsByTags: vi.fn(),
 	hasBeenRevalidated: vi.fn(),
 	writeTags: vi.fn(),
+	isStale: vi.fn(),
 } satisfies NextModeTagCache;
 
 const filterFn = (tag: string) => tag.startsWith("valid_");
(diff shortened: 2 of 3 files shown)


vicb:
Thanks for the issue and the fix @fatihy101 🙏

I'll add a changeset to the PR (running `pnpm changeset`) and merge.

vicb (approved):
Looks good to me 🙏

PR #1193, opened 2026-04-11
Fix tag cache stale logic and update tests

conico974:
Update the stale logic in the tag cache to consider the `revalidatedAt` timestamp, ensuring accurate cache validation. 

It doesn't technically require the bump of aws


packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
```diff
@@ -45,6 +45,10 @@ type Options = {
 	 * Note: When this is enabled, make sure that the cache gets purged
 	 *       either by enabling the auto cache purging feature or manually.
 	 *
+	 * This is currently not compatible with swr types of revalidateTag (i.e. on Next 16+, anything different than `revalidateTag("tag", { expire: 0 })`),
```
vicb: @conico974 Wan we try to clarify that?

- Does "This is currently not compatible" mean that `true` should not be used?
- IIUC if `shouldLazilyUpdateOnCacheHit` is true, this can also be `true` but users still have to use `{ expire: 0 }` for data cache?

By "not compatible" do you mean that cache will not work? in what ways?

What about creating an issue to follow up?
Can we log an error at runtime?

packages/cloudflare/src/api/overrides/tag-cache/d1-next-tag-cache.ts
```diff
@@ -105,8 +105,13 @@ export class D1NextModeTagCache implements NextModeTagCache {
 
 			const isStale = [...result.values()].some((v) => {
 				if (v == null) return false;
-				const { stale, expire } = v;
-				if (stale == null || stale <= (lastModified ?? now)) return false;
+				const { revalidatedAt, stale, expire } = v;
+				// A tag is stale when both its stale timestamp and its revalidatedAt are newer than the page.
```
vicb:
```suggestion
				// A tag is stale when both its stale and revalidatedAt timestamps are newer than the page.
```

packages/cloudflare/src/api/overrides/tag-cache/d1-next-tag-cache.ts
```diff
@@ -105,8 +105,13 @@ export class D1NextModeTagCache implements NextModeTagCache {
...
+				// revalidatedAt > lastModified ensures the revalidation that set this stale window happened
+				// after the page was generated, preventing a stale signal from a previous ISR cycle.
+				const isInStaleWindow =
+					stale != null && revalidatedAt > (lastModified ?? now) && (lastModified ?? now) <= stale;
```
vicb: nit: store `lastModified ?? now` in a const for readability

vicb: (everywhere applicable)

vicb (commented):
@conico974 I have added a few minor comments, could you please take a look and let me know what you think.

Changes otherwise look good so I can update the PR when I get your feedback and get that merged.

Thanks!

packages/cloudflare/src/api/overrides/incremental-cache/regional-cache.ts
conico974: The problem with `bypassTagCacheOnCacheHit` is that it bypass the tag cache :smile:  
So on `revalidateTag` the cache is not updated immediately, and even if we purge the regional cache, next time it is hit, the value might still be the stale one, and we'll repopulate the regional cache with this stale value, which will bypass the tag cache check next time and keep serving that stale value.
And with the data cache it can become an absolute mess since some entry might have been revalidated in another region and some not, and you end up with a potentially very inconsistent state.

> IIUC if shouldLazilyUpdateOnCacheHit is true, this can also be true but users still have to use { expire: 0 } for data cache?

This would work for page entry for any swr revalidateTag, for the data cache you always risk being in an inconsistent state.

> Can we log an error at runtime?

We could do that on writeTags (or a warning), and at build time we could print a warning as well

Not sure if I made myself clear here, it's always a mess to explain these race conditions

examples/e2e/app-router/open-next.config.ts
```diff
@@ -3,9 +3,10 @@ import r2IncrementalCache from "@opennextjs/cloudflare/overrides/incremental-cac
 import shardedTagCache from "@opennextjs/cloudflare/overrides/tag-cache/do-sharded-tag-cache";
 import doQueue from "@opennextjs/cloudflare/overrides/queue/do-queue";
 import queueCache from "@opennextjs/cloudflare/overrides/queue/queue-cache";
+import { withRegionalCache } from "@opennextjs/cloudflare/overrides/incremental-cache/regional-cache";
 
 const baseConfig = defineCloudflareConfig({
-	incrementalCache: r2IncrementalCache,
+	incrementalCache: withRegionalCache(r2IncrementalCache, { mode: "long-lived" }),
```
vicb: This needs to be reverted

vicb (dismissed):
The build is broken by my last commit (looks like nextVersion is undefined).
I'll take a look and fix

vicb:
The build issue was related to `globalThis.nextVersion` not being defined at build time. The code has been updated to fix that.
_The Open Next configuration is evaluated at build time_

vicb (approved):
LGTM, thanks a lot @conico974 !

PR #1196, opened 2026-04-13
fix(turbopack): handle Turbopack hashed external module IDs in workerd

Thereallo1026:
## Problem

Turbopack externalizes some packages using content-hashed module IDs like `shiki-43d062b67f27bbdc/core`. These are generated by Turbopack's module ID hashing and appear as `a.y("shiki-43d062b67f27bbdc/core")` calls inside the `[externals]_*` chunks.

In **workerd**, `await import(id)` only works for registered named modules. A hashed ID like `shiki-43d062b67f27bbdc/core` is not a valid module name, so the call fails with:

```
Failed to load external module shiki-43d062b67f27bbdc/core:
Error: No such module "shiki-43d062b67f27bbdc/core"
```

The actual package (`shiki`) **is** bundled in the worker's node_modules and is accessible via `require("shiki/core")` — the hash just needs to be stripped.

## Fix

The `inlineExternalImportRule` AST patch already wraps `externalImport` in a switch. I've extended the `default` case to detect Turbopack's hash pattern at runtime:

```
<packagename>-<16+ hex chars>[/subpath]
```

When detected, it strips the hash and falls back to `require()`:

```js
default: {
  const __dehashedId = $ID.replace(/^((?:@[^/]+\/)?[^/]+?)-[0-9a-f]{16,}(\/[^]*)?$/, "$1$2");
  if (__dehashedId !== $ID) {
    $RAW = require(__dehashedId || $ID);
  } else {
    $RAW = await import($ID);
  }
}
```

This handles:
- `shiki-43d062b67f27bbdc` → `shiki`
- `shiki-43d062b67f27bbdc/core` → `shiki/core`
- `shiki-43d062b67f27bbdc/wasm` → `shiki/wasm`
- `@shikijs/core-43d062b67f27bbdc/dist/index.js` → `@shikijs/core/dist/index.js`
- Normal IDs (`react`, `next/dist/...`) → unchanged, use `await import()` as before

## Tests

Added `turbopack.spec.ts` covering the regex behaviour for all these cases. All 302 existing tests continue to pass.

## Reproduction

Any Next.js 16 (Turbopack) app that uses `shiki` (e.g. via `fumadocs-core`) will hit this when deployed to Cloudflare Workers.

### DIFF
--- a/packages/cloudflare/src/cli/build/patches/plugins/turbopack.spec.ts
+++ b/packages/cloudflare/src/cli/build/patches/plugins/turbopack.spec.ts
@@ -1,85 +1,48 @@
 import { describe, expect, test } from "vitest";
 
-import { computePatchDiff } from "../../utils/test-patch.js";
+import { patchHashedExternalImports } from "./turbopack.js";
 
-// Re-export the rule for testing (it's not exported from turbopack.ts, so we inline it here)
-const inlineExternalImportRule = `
-rule:
-  pattern: "$RAW = await import($ID)"
-  inside:
-    regex: "externalImport"
-    kind: function_declaration
-    stopBy: end
-fix: |-
-  switch ($ID) {
-    case "next/dist/compiled/@vercel/og/index.node.js":
-      $RAW = await import("next/dist/compiled/@vercel/og/index.edge.js");
-      break;
-    default: {
-      // Turbopack hashes external package IDs: e.g. "shiki-43d062b67f27bbdc/core"
-      // Strip the hash suffix to recover the real package name, then use require().
-      const __dehashedId = $ID.replace(/^((?:@[^/]+\\/)?[^/]+?)-[0-9a-f]{16,}(\\/[^]*)?$/, "$1$2");
-      if (__dehashedId !== $ID) {
-        $RAW = require(__dehashedId || $ID);
-      } else {
-        $RAW = await import($ID);
-      }
-    }
-  }
-`;
-
-const externalImportFn = `
-async function externalImport(id) {
-    let raw;
-    try {
-        raw = await import(id);
-    } catch (err) {
-        throw new Error(\`Failed to load external module \${id}: \${err}\`);
-    }
-    return raw;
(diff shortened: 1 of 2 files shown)


vicb:
I realized that #1139 was not merged yet. 
Does this PR fix the same issue? 
@james-elicx ?

Thereallo1026:
Hi human here (Opus degrade is real), the previous commit was wrong

Patching `externalImport` in `[turbopack]_runtime.js` to use `require()` fails because by the time that function runs, wrangler has already finished bundling. Calling `require()` on an ESM-only package like `shiki` at that point throws:

```
Dynamic require of "shiki/core" is not supported
```

But the `a.y("shiki-<hash>/sub")` calls live inside the chunk files themselves, both `[externals]_*.js` and `[root-of-the-server]_*.js`

If we replace them with `Promise.resolve().then(() => require("shiki/sub"))` before wrangler bundles, wrangler can see the static `require()` at bundle time and inlines the real package correctly

I added a patch in the second commit targeting all `.next/server/chunks/**/*.js` files (filtered by `contentFilter: /a\.y\("/`) that does this, the `Promise.resolve().then()` wrapper preserves the async contract of the original `a.y()` call

The regex strips Turbopack's content hash (`<pkg>-<16+ hex chars>[/subpath]`) and handles scoped packages too

Tested working with a real fumadocs app on `opennextjs-cloudflare preview`

Thereallo1026:
@vicb yes, this fixes the same issue as #1139

I wasn't aware of it when I opened this PR, #1139 looks like a more robust solution

I'd suggest merging that one instead and closing this, apologies for the noise.

vicb:
> @vicb yes, this fixes the same issue as #1139
> 
> I wasn't aware of it when I opened this PR, #1139 looks like a more robust solution
> 
> I'd suggest merging that one instead and closing this, apologies for the noise.

No worries @Thereallo1026 and thanks for the PR.
We'll get 1139 merged

PR #1199, opened 2026-04-13
fix(cli): fail fast in non-TTY environments instead of hanging on config-creation prompts

### LINKED ISSUE #1198: [BUG] opennextjs-cloudflare build hangs on an interactive prompt in non-TTY environments
### Describe the bug

When `open-next.config.ts` is missing, `opennextjs-cloudflare build` prompts the user to auto-create it. In non-TTY environments (Cloudflare Workers Builds, Docker, GitHub Actions, etc.) the Enquirer-backed prompt can't read stdin, the build hangs on a half-rendered prompt and exits with a cryptic code, and nothing in the output tells the user what they should do.

The same class of bug exists for the missing-`wrangler.(toml|json|jsonc)` prompt in `buildCommand`. That path *does* have an escape hatch (`--skipWranglerConfigCheck` / `SKIP_WRANGLER_CONFIG_CHECK`), but users hit the prompt before they even know the flag exists.

Offending code paths:

- [`packages/cloudflare/src/cli/commands/utils/utils.ts`](https://github.com/opennextjs/opennextjs-cloudflare/blob/main/packages/cloudflare/src/cli/commands/utils/utils.ts):`compileConfig` calls `askConfirmation` unconditionally.
- [`packages/cloudflare/src/cli/commands/build.ts`](https://github.com/opennextjs/opennextjs-cloudflare/blob/main/packages/cloudflare/src/cli/commands/build.ts): `buildCommand` calls `askConfirmation` without checking TTY state.

### Steps to reproduce

1. Create a minimal Next.js 16 app (`next.config.ts`, `app/layout.tsx`, `app/page.tsx`, `wrangler.jsonc`). Do **not** create `open-next.config.ts`.
2. Install `@opennextjs/cloudflare` and `wrangler`.
3. Run the build with stdin redirected from `/dev/null` to simulate a non-TTY environment:

   ```bash
   pnpm opennextjs-cloudflare build < /dev/null
   ```

Actual output:

```
? Missing required `open-next.config.ts` file, do you want to create one? (Y/n) ‣ true
 ELIFECYCLE  Command failed with exit code 13.
Failed: error occurred while running build command
```

On the actual Cloudflare Workers Builds runner, the error is identical, we first hit this on a real deployment, then reproduced it locally by piping `/dev/null` into stdin.

---

### Expected behavior

In a non-interactive environment, the CLI should fail fast with an actionable error that tells the user exactly what to do, either showing the three-line default config they need to create, or pointing at a documented way to skip the prompt. It should never hang or exit with a code and a truncated prompt.

(Local/interactive behavior should stay the same: if there's a TTY, prompting is fine.)

`1.18.0` (also verified against `1.19.1` on `main`)

### Wrangler version

4.78.0

### next info output

```shell
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Home Single Language
  Available memory (MB): 20333
  Available CPU cores: 16
Binaries:
  Node: 22.21.1
  npm: 10.9.4
  Yarn: N/A
  pnpm: 10.28.1
Relevant Packages:
  next: 16.2.1 // There is a newer version (16.2.3) available, upgrade recommended!
  eslint-config-next: N/A
  react: 19.2.4
  react-dom: 19.2.4
  typescript: N/A
Next.js Config:
  output: N/A



> The failing build that originally surfaced this issue ran on Cloudflare Workers Builds' Linux runner, Node `22.16.0`, pnpm `10.11.1`, same Next.js `16.2.1`. The `next info` output above is from the same Next.js project run locally on Windows; the bug reproduces on both.
```

### Additional context

- The exit code the user sees (`exit code 13`) comes from Enquirer failing on a missing TTY, not from any clean error path inside `@opennextjs/cloudflare`. There's no logged context.
- The `wrangler.(toml|json|jsonc)` prompt has the same defect but already has a skip flag, so the fix for it is mostly a nicer error message pointing at that flag.
- I have a fix ready and will open a PR shortly that introduces an `isInteractive()` helper (mirroring wrangler's own TTY check at [`run-wrangler.ts:120`](https://github.com/opennextjs/opennextjs-cloudflare/blob/main/packages/cloudflare/src/cli/commands/utils/run-wrangler.ts)) and gates both prompts behind it, throwing actionable errors instead of hanging. Happy to close this issue with the PR if preferred.

SdSadat:
## Problem

On Cloudflare Workers Builds (and any other non-TTY environment — Docker, GitHub Actions, etc.), `opennextjs-cloudflare build` crashes with a truncated interactive prompt and a cryptic exit code when `open-next.config.ts` is missing:

```text
? Missing required `open-next.config.ts` file, do you want to create one? (Y/n) ‣ true
 ELIFECYCLE  Command failed with exit code 13.
Failed: error occurred while running build command
```

[`compileConfig`](https://github.com/opennextjs/opennextjs-cloudflare/blob/main/packages/cloudflare/src/cli/commands/utils/utils.ts) unconditionally calls `askConfirmation` (backed by Enquirer). Enquirer needs a readable stdin — in CI it errors out immediately, the user sees a half-rendered prompt, and there's nothing in the message telling them what to do.

The same class of bug exists in [`buildCommand`](https://github.com/opennextjs/opennextjs-cloudflare/blob/main/packages/cloudflare/src/cli/commands/build.ts) for the missing-`wrangler.(toml|json|jsonc)` prompt. That path already has an escape hatch (`--skipWranglerConfigCheck` / `SKIP_WRANGLER_CONFIG_CHECK`), but users hit the prompt before they even know the flag exists.

## Fix

Introduces an `isInteractive()` helper that matches the TTY check wrangler itself uses (`process.stdin.isTTY && process.stdout.isTTY`, noted in a comment in [`run-wrangler.ts:120`](https://github.com/opennextjs/opennextjs-cloudflare/blob/main/packages/cloudflare/src/cli/commands/utils/run-wrangler.ts)) and gates both prompts behind it.

- **`compileConfig`** — in non-interactive mode, throws with the exact 3-line template the user needs to paste into `open-next.config.ts`. Zero-friction recovery.
- **`buildCommand`** — in non-interactive mode, the wrangler-config prompt throws with a message pointing at the existing `--skipWranglerConfigCheck` / `SKIP_WRANGLER_CONFIG_CHECK` escape hatch. No new flag.
- **Interactive behavior is unchanged** — if there's a TTY, you still get the prompt.

Chose *fail fast with instructions* over *auto-create the file silently in CI* because:

1. Auto-creating mutates project state the user never committed — the next local build hits the same error, creating a confusing dev/CI divergence.
2. A clear error is actionable. The error message ships the exact 3-line template, so "copy-paste-commit" is the whole recovery path.
3. It's the least surprising change to ship: no new behavior, no new flags, no new tests beyond the two that cover the new branch.

## Before / After

**Before** (non-TTY, missing `open-next.config.ts`):

```text
? Missing required `open-next.config.ts` file, do you want to create one? (Y/n) ‣ true
 ELIFECYCLE  Command failed with exit code 13.
```

**After** (non-TTY, missing `open-next.config.ts`):

```text
Error: No `open-next.config.ts` file was found in the project root.

This file is required for OpenNext Cloudflare builds. Create it with:

    import { defineCloudflareConfig } from '@opennextjs/cloudflare';

    export default defineCloudflareConfig({});

Commit it and re-run the build. (To create it interactively, run
`opennextjs-cloudflare build` in a terminal with TTY access.)

 ELIFECYCLE  Command failed with exit code 1.
```

## Testing

Extended `packages/cloudflare/src/cli/commands/utils/utils.spec.ts`:

- Mocked `isInteractive` with a `beforeEach` that resets it to `true` so existing tests continue to exercise the TTY branch unchanged.
- **New test**: asserts that in a non-interactive environment the non-TTY branch fires, `askConfirmation` is **not** called, `createOpenNextConfigFile` is **not** called, and the error body matches the template (regex-checks for both `No open-next.config.ts` and `defineCloudflareConfig`).
- **New test**: explicit regression check that the TTY branch still reaches `askConfirmation` and `createOpenNextConfigFile` in interactive mode.

```
 Test Files  1 passed (1)
 Tests       7 passed (7)
```

Prettier, ESLint, and `tsc --noEmit` all clean on the touched files.

### End-to-end repro

To reproduce end-to-end against a real build, scaffold a minimal Next.js 16 app (`next.config.ts` + `app/layout.tsx` + `app/page.tsx` + `wrangler.jsonc`, intentionally **no** `open-next.config.ts`), install this branch, and run:

```bash
pnpm opennextjs-cloudflare build < /dev/null   # redirect stdin from /dev/null to simulate non-TTY
```

Hits the new error path immediately with exit code 1. Dropping in the 3-line config from the error message and re-running unblocks the build past `compileConfig`. This matches the symptom in the issue originally reported against Cloudflare Workers Builds (#1198).

## Files Changed

| File | Change |
| --- | --- |
| `.changeset/fix-non-tty-config-prompts.md` | **New** — patch changeset |
| `packages/cloudflare/src/cli/utils/is-interactive.ts` | **New** — `isInteractive()` helper |
| `packages/cloudflare/src/cli/commands/utils/utils.ts` | Import helper, add non-interactive branch in `compileConfig` |
| `packages/cloudflare/src/cli/commands/build.ts` | Import helper, add non-interactive branch for wrangler-config prompt |
| `packages/cloudflare/src/cli/commands/utils/utils.spec.ts` | Mock `isInteractive`, add `beforeEach`, +2 test cases |

+93/-1, no new dependencies.

Fixes #1198.

### DIFF
--- a/packages/cloudflare/src/cli/commands/build.ts
+++ b/packages/cloudflare/src/cli/commands/build.ts
@@ -4,7 +4,7 @@ import type yargs from "yargs";
 import { build as buildImpl } from "../build/build.js";
 import { askConfirmation } from "../utils/ask-confirmation.js";
 import { createWranglerConfigFile, findWranglerConfig } from "../utils/create-wrangler-config.js";
-import { isInteractive } from "../utils/is-interactive.js";
+import { isNonInteractiveOrCI } from "../utils/is-interactive.js";
 import type { WithWranglerArgs } from "./utils/utils.js";
 import {
 	compileConfig,
@@ -45,12 +45,9 @@ export async function buildCommand(
 			// In non-interactive environments (CI, Cloudflare Workers Builds,
 			// Docker, etc.) the prompt would hang or crash. Fail fast with a
 			// clear message pointing at the existing escape hatch.
-			if (!isInteractive()) {
+			if (isNonInteractiveOrCI()) {
 				throw new Error(
-					"No `wrangler.(toml|json|jsonc)` config file found.\n\n" +
-						"Create one at the project root before running the build, " +
-						"or skip this check with `--skipWranglerConfigCheck` or " +
-						"`SKIP_WRANGLER_CONFIG_CHECK=yes`."
+					"No `wrangler.(toml|json|jsonc)` config file found.\n\nCreate one at the project root before running the build, or skip this check with `--skipWranglerConfigCheck` or `SKIP_WRANGLER_CONFIG_CHECK=yes`."
 				);
 			}
 
--- a/packages/cloudflare/src/cli/commands/build.ts
+++ b/packages/cloudflare/src/cli/commands/build.ts
@@ -4,6 +4,7 @@ import type yargs from "yargs";
 import { build as buildImpl } from "../build/build.js";
 import { askConfirmation } from "../utils/ask-confirmation.js";
 import { createWranglerConfigFile, findWranglerConfig } from "../utils/create-wrangler-config.js";
+import { isInteractive } from "../utils/is-interactive.js";
 import type { WithWranglerArgs } from "./utils/utils.js";
 import {
 	compileConfig,
@@ -41,6 +42,18 @@ export async function buildCommand(
 	//       nor when `--skipWranglerConfigCheck` is used.
 	if (!projectOpts.wranglerConfigPath && !args.skipWranglerConfigCheck) {
 		if (!findWranglerConfig(projectOpts.sourceDir)) {
+			// In non-interactive environments (CI, Cloudflare Workers Builds,
+			// Docker, etc.) the prompt would hang or crash. Fail fast with a
+			// clear message pointing at the existing escape hatch.
+			if (!isInteractive()) {
+				throw new Error(
+					"No `wrangler.(toml|json|jsonc)` config file found.\n\n" +
+						"Create one at the project root before running the build, " +
+						"or skip this check with `--skipWranglerConfigCheck` or " +
+						"`SKIP_WRANGLER_CONFIG_CHECK=yes`."
+				);
+			}
+
 			const confirmCreate = "No `wrangler.(toml|json|jsonc)` config file found, do you want to create one?";
 			if (await askConfirmation(confirmCreate)) {
 				await createWranglerConfigFile(projectOpts.sourceDir);
--- a/packages/cloudflare/src/cli/commands/utils/utils.ts
+++ b/packages/cloudflare/src/cli/commands/utils/utils.ts
@@ -18,7 +18,7 @@ import {
 	findOpenNextConfig,
 	OPEN_NEXT_CONFIG_FILE_NAME,
 } from "../../utils/create-open-next-config.js";
-import { isInteractive } from "../../utils/is-interactive.js";
+import { isNonInteractiveOrCI } from "../../utils/is-interactive.js";
 
 export type WithWranglerArgs<T = unknown> = T & {
 	// Array of arguments that can be given to wrangler commands, including the `--config` and `--env` args.
@@ -66,12 +66,9 @@ export async function compileConfig(configPath: string | undefined) {
 		// In non-interactive environments (CI, Cloudflare Workers Builds,
 		// Docker, etc.) there is no TTY to answer a prompt — the build would
 		// hang or crash. Fail fast with an actionable message instead.
-		if (!isInteractive()) {
+		if (isNonInteractiveOrCI()) {
 			throw new Error(
-				`No \`${OPEN_NEXT_CONFIG_FILE_NAME}\` file was found in the project root.\n\n` +
-					"This file is required for OpenNext Cloudflare builds.\n" +
-					"Run `opennextjs-cloudflare migrate` to create it, or see https://opennext.js.org/cloudflare/get-started for setup guidance.\n" +
-					"Commit it and re-run the build."
+				`No \`${OPEN_NEXT_CONFIG_FILE_NAME}\` file was found in the project root.\n\nThis file is required for OpenNext Cloudflare builds.\nRun \`opennextjs-cloudflare migrate\` to create it, or see https://opennext.js.org/cloudflare/get-started for setup guidance.\nCommit it and re-run the build.`
 			);
 		}
 
--- a/packages/cloudflare/src/cli/commands/utils/utils.ts
+++ b/packages/cloudflare/src/cli/commands/utils/utils.ts
@@ -13,7 +13,11 @@ import type yargs from "yargs";
 import type { OpenNextConfig } from "../../../api/config.js";
 import { ensureCloudflareConfig } from "../../build/utils/ensure-cf-config.js";
 import { askConfirmation } from "../../utils/ask-confirmation.js";
-import { createOpenNextConfigFile, findOpenNextConfig } from "../../utils/create-open-next-config.js";
+import {
+	createOpenNextConfigFile,
+	findOpenNextConfig,
+	OPEN_NEXT_CONFIG_FILE_NAME,
+} from "../../utils/create-open-next-config.js";
 import { isInteractive } from "../../utils/is-interactive.js";
 
 export type WithWranglerArgs<T = unknown> = T & {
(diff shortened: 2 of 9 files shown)


packages/cloudflare/src/cli/commands/utils/utils.ts
```diff
@@ -58,6 +59,21 @@ export async function compileConfig(configPath: string | undefined) {
 	configPath ??= findOpenNextConfig(nextAppDir);
 
 	if (!configPath) {
+		// In non-interactive environments (CI, Cloudflare Workers Builds,
+		// Docker, etc.) there is no TTY to answer a prompt — the build would
+		// hang or crash. Fail fast with an actionable message instead.
+		if (!isInteractive()) {
+			throw new Error(
+				"No `open-next.config.ts` file was found in the project root.\n\n" +
```
vicb: A few comments:

Could you please extract "open-next.config.ts" to a constant and use it here (so that the message doesn't get outdated if it changes).

Similarly please do no include code snippets but rather a link to [the docs](https://opennext.js.org/cloudflare/get-started#existing-nextjs-apps) or a mention of the migrate command would be more robust.

Thanks!

packages/cloudflare/src/cli/utils/is-interactive.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Whether the current process is running in an interactive terminal.
+ *
+ * Used to gate prompts that would otherwise hang or error in non-TTY
+ * environments like CI, Docker builds, or Cloudflare Workers Builds.
+ *
+ * Matches the same check wrangler uses for its own interactive detection.
```
vicb: Why we use `isInteractive` is documented at the call sites, no need to repeat here.

```suggestion
```

vicb (commented):
Thanks for the PR.

I have added inline comments.

Maybe you should use https://github.com/cloudflare/workers-sdk/blob/409d23a09eabc5cb195307a6150bc9cad0cc30b4/packages/wrangler/src/is-interactive.ts#L38-L58 instead?

packages/cloudflare/src/cli/commands/utils/utils.ts
SdSadat: Thanks for the suggestion, onto it!

packages/cloudflare/src/cli/utils/is-interactive.ts
SdSadat: Sure, will remove it.

packages/cloudflare/src/cli/commands/utils/utils.ts
```diff
@@ -58,12 +63,24 @@ export async function compileConfig(configPath: string | undefined) {
 	configPath ??= findOpenNextConfig(nextAppDir);
 
 	if (!configPath) {
+		// In non-interactive environments (CI, Cloudflare Workers Builds,
+		// Docker, etc.) there is no TTY to answer a prompt — the build would
+		// hang or crash. Fail fast with an actionable message instead.
+		if (!isInteractive()) {
+			throw new Error(
+				`No \`${OPEN_NEXT_CONFIG_FILE_NAME}\` file was found in the project root.\n\n` +
```
vicb: nit: maybe use a single multi-line string instead and concatenating strings

packages/cloudflare/src/cli/commands/build.ts
```diff
@@ -41,6 +42,18 @@ export async function buildCommand(
 	//       nor when `--skipWranglerConfigCheck` is used.
 	if (!projectOpts.wranglerConfigPath && !args.skipWranglerConfigCheck) {
 		if (!findWranglerConfig(projectOpts.sourceDir)) {
+			// In non-interactive environments (CI, Cloudflare Workers Builds,
+			// Docker, etc.) the prompt would hang or crash. Fail fast with a
+			// clear message pointing at the existing escape hatch.
+			if (!isInteractive()) {
+				throw new Error(
+					"No `wrangler.(toml|json|jsonc)` config file found.\n\n" +
```
vicb: nit: ditto for multi-line string

vicb (commented):
Thanks for the updates

2 more nits.

What about my previous comment:

> Maybe you should use https://github.com/cloudflare/workers-sdk/blob/409d23a09eabc5cb195307a6150bc9cad0cc30b4/packages/wrangler/src/is-interactive.ts#L38-L58 instead?

SdSadat:
> Thanks for the updates
> 
> 2 more nits.
> 
> What about my previous comment:
> 
> > Maybe you should use https://github.com/cloudflare/workers-sdk/blob/409d23a09eabc5cb195307a6150bc9cad0cc30b4/packages/wrangler/src/is-interactive.ts#L38-L58 instead?

To align with your preference, should we:

Fully mirror Wrangler interactivity helpers, or
Keep this PR scoped to prompt suppression only?
Also, are you okay with adding ci-info for CI detection?

vicb:
> > Thanks for the updates
> > 2 more nits.
> > What about my previous comment:
> > > Maybe you should use https://github.com/cloudflare/workers-sdk/blob/409d23a09eabc5cb195307a6150bc9cad0cc30b4/packages/wrangler/src/is-interactive.ts#L38-L58 instead?
> 
> To align with your preference, should we:
> 
> Fully mirror Wrangler interactivity helpers, or Keep this PR scoped to prompt suppression only? Also, are you okay with adding ci-info for CI detection?

Probably adding a `isNonInteractiveOrCI` is enough.

IMO it should: use the current code and add the check from the `ci` package. IIRC the latest version of this package supports Cloudflare Workers Build so no need to add an explicit check

vicb (approved):
LGTM, Thanks for the updates!

vicb:
I have merged in the latest changes

PR #1208, opened 2026-04-17
replace NEXT_BUILD_ID with OPEN_NEXT_BUILD_ID

edmundhung:
This matches the changes in https://github.com/opennextjs/opennextjs-aws/pull/1144 to deprecate the `NEXT_BUILD_ID` environment variable in favor of `OPEN_NEXT_BUILD_ID`.


packages/cloudflare/src/cli/templates/init.ts
```diff
@@ -145,6 +145,7 @@ function populateProcessEnv(url: URL, env: CloudflareEnv) {
 	if (__DEPLOYMENT_ID__) {
 		process.env.DEPLOYMENT_ID = __DEPLOYMENT_ID__;
 	}
+	process.env.OPEN_NEXT_BUILD_ID = __OPEN_NEXT_BUILD_ID__;
```
james-elicx: I believe we'll probably need to bump opennext before we can make these changes ourselves as I believe the part of the template where this var would be set (their change in adapters/config/index.ts) is built as part of the worker in ours. So my belief is that we might not need to add this in the init.ts template in cloudflare, but i can't confirm that without testing a branch based on the new aws version.

alex-all3dp:
@edmundhung @james-elicx is this the fix for https://github.com/opennextjs/opennextjs-aws/issues/1136 and  https://github.com/opennextjs/opennextjs-cloudflare/issues/1183 respectively? Does it require any changes on the user's side or will `getDeploymentId()` already return `OPEN_NEXT_BUILD_ID` once these changes are merged?

packages/cloudflare/src/cli/templates/init.ts
vicb: I agree with @james-elicx here, `process.env.OPEN_NEXT_BUILD_ID` is set by aws (in the latest version) and we do not need to explicitely set that here.

I have verified with the playground15 on main (checking http://localhost:8787/api/env)

james-elicx:
I believe some of the patches might need updating as well

vicb:
> I believe some of the patches might need updating as well

Do you have anything specific in mind ?

FYI I have just created [a PR to aws](https://github.com/opennextjs/opennextjs-aws/pull/1149) and have further local commits that I will push after AWS is released but no update to patches yet.

Thanks

packages/cloudflare/src/cli/templates/init.ts
vicb: Pushed a commit to fix that [bebeac3](https://github.com/opennextjs/opennextjs-cloudflare/pull/1208/commits/bebeac3f7555e7a6f40992cfcaba051b156eb175)

james-elicx:
> > I believe some of the patches might need updating as well
> 
> Do you have anything specific in mind ?
> 
> FYI I have just created [a PR to aws](https://github.com/opennextjs/opennextjs-aws/pull/1149) and have further local commits that I will push after AWS is released but no update to patches yet.
> 
> Thanks

I'm assuming these parts would want to reference the new variable:

https://github.com/opennextjs/opennextjs-cloudflare/blob/a2679bf9549f620e1ab0e1900dcc7a6b6ac03e0a/packages/cloudflare/src/cli/build/patches/plugins/load-manifest.ts#L70-L72

https://github.com/opennextjs/opennextjs-cloudflare/blob/a2679bf9549f620e1ab0e1900dcc7a6b6ac03e0a/packages/cloudflare/src/cli/build/patches/plugins/next-server.ts#L71-L73

vicb:
> > > I believe some of the patches might need updating as well
> > 
> > 
> > Do you have anything specific in mind ?
> > FYI I have just created [a PR to aws](https://github.com/opennextjs/opennextjs-aws/pull/1149) and have further local commits that I will push after AWS is released but no update to patches yet.
> > Thanks
> 
> I'm assuming these parts would want to reference the new variable:
> 
> https://github.com/opennextjs/opennextjs-cloudflare/blob/a2679bf9549f620e1ab0e1900dcc7a6b6ac03e0a/packages/cloudflare/src/cli/build/patches/plugins/load-manifest.ts#L70-L72
> 
> https://github.com/opennextjs/opennextjs-cloudflare/blob/a2679bf9549f620e1ab0e1900dcc7a6b6ac03e0a/packages/cloudflare/src/cli/build/patches/plugins/next-server.ts#L71-L73

I believe not, this is Next code and behavior should not be altered.
@conico974 do you confirm?

vicb:
@conico974 @edmundhung @james-elicx Could you please take a quick look at the PR. IMO it's good to go (I'll release AWS and use that after the changes here are validated).

james-elicx (approved):

conico974:
> > > > I believe some of the patches might need updating as well
> > > 
> > > 
> > > Do you have anything specific in mind ?
> > > FYI I have just created [a PR to aws](https://github.com/opennextjs/opennextjs-aws/pull/1149) and have further local commits that I will push after AWS is released but no update to patches yet.
> > > Thanks
> > 
> > 
> > I'm assuming these parts would want to reference the new variable:
> > https://github.com/opennextjs/opennextjs-cloudflare/blob/a2679bf9549f620e1ab0e1900dcc7a6b6ac03e0a/packages/cloudflare/src/cli/build/patches/plugins/load-manifest.ts#L70-L72
> > 
> > https://github.com/opennextjs/opennextjs-cloudflare/blob/a2679bf9549f620e1ab0e1900dcc7a6b6ac03e0a/packages/cloudflare/src/cli/build/patches/plugins/next-server.ts#L71-L73
> 
> I believe not, this is Next code and behavior should not be altered. @conico974 do you confirm?

Yeah we don't want to modify that, that's also why I named the new env `OPEN_NEXT_BUILD_ID` instead

packages/cloudflare/package.json
```diff
@@ -54,7 +54,7 @@
 	"dependencies": {
 		"@ast-grep/napi": "^0.40.5",
 		"@dotenvx/dotenvx": "catalog:",
-		"@opennextjs/aws": "3.10.2",
+		"@opennextjs/aws": "https://pkg.pr.new/@opennextjs/aws@1149",
```
conico974: Don't forget to update this one 😉

conico974 (approved):

vicb:
Thanks for the reviews @conico974 and @james-elicx !

vicb (approved):
Thanks for the PR @edmundhung !

PR #1210, opened 2026-04-20
refactor: add explicit TypeScript return type annotations

314systems:


packages/cloudflare/src/api/durable-objects/queue.ts
```diff
@@ -23,11 +23,11 @@ export class DOQueueHandler extends DurableObject<CloudflareEnv> {
 	// Ongoing revalidations are deduped by the deduplication id
 	// Since this is running in waitUntil, we expect the durable object state to persist this during the duration of the revalidation
 	// TODO: handle incremental cache with only eventual consistency (i.e. KV or R2/D1 with the optional cache layer on top)
-	ongoingRevalidations = new Map<string, Promise<void>>();
+	ongoingRevalidations: Map<string, Promise<void>> = new Map<string, Promise<void>>();
```
vicb: I don't think this change (and the next one) is necessary.

I like the explicit return types, thanks for the update.

packages/cloudflare/src/api/overrides/incremental-cache/kv-incremental-cache.ts
```diff
@@ -27,7 +27,7 @@ export const PREFIX_ENV_NAME = "NEXT_INC_CACHE_KV_PREFIX";
  * when the constructor is called.
  */
 class KVIncrementalCache implements IncrementalCache {
-	readonly name = NAME;
+	readonly name: string = NAME;
```
vicb: I think this should be reverted (everywhere applicable)

packages/cloudflare/src/api/overrides/queue/memory-queue.ts
```diff
@@ -17,9 +17,13 @@ export const DEFAULT_REVALIDATION_TIMEOUT_MS = 10_000;
 export class MemoryQueue implements Queue {
 	readonly name = "memory-queue";
 
-	revalidatedPaths = new Set<string>();
+	revalidatedPaths: Set<string> = new Set<string>();
```
vicb: please revert

packages/cloudflare/src/api/overrides/queue/memory-queue.ts
```diff
@@ -17,9 +17,13 @@ export const DEFAULT_REVALIDATION_TIMEOUT_MS = 10_000;
...
 
-	constructor(private opts = { revalidationTimeoutMs: DEFAULT_REVALIDATION_TIMEOUT_MS }) {}
+	constructor(
```
vicb: I would revert this one too

vicb (commented):
I like the added explicit return types.

I think some of the changes in this PR are not necessary (see my inline comments).

packages/cloudflare/src/api/durable-objects/queue.ts
314systems: Fair point. I was over-annotating while testing isolatedDeclarations. Does this look like a good middle ground?

```ts
ongoingRevalidations: Map<string, Promise<void>> = new Map();
```

vicb: IMO the original code is the idiomatic TS way and does not need to be changed.

packages/cloudflare/src/api/durable-objects/queue.ts
```diff
@@ -27,7 +27,7 @@ export class DOQueueHandler extends DurableObject<CloudflareEnv> {
 
 	sql: SqlStorage;
 
-	routeInFailedState = new Map<string, FailedState>();
+	routeInFailedState: Map<string, FailedState> = new Map();
```
vicb: please revert

vicb (commented):
One last nit

examples/common/config-e2e.ts
```diff
@@ -1,13 +1,11 @@
...
 import { getAppPort, getInspectorPort, type AppName } from "./apps";
 
-declare const process: typeof nodeProcess;
-
 export function configurePlaywright(
 	app: AppName,
 	{
 		// Do we run on CI?
-		isCI = Boolean(process.env.CI),
+		isCI = !!process.env.CI,
```
vicb: Could you please revert this change

packages/cloudflare/src/cli/build/utils/test-patch.ts
```diff
@@ -9,7 +9,7 @@ import { createPatch } from "diff";
  * @param rule ASTgrep rule
  * @returns diff in unified diff format
  */
-export function computePatchDiff(filename: string, src: string, rule: string): string {
+export function computePatchDiff(filename: string, src: string, rule: string | RuleConfig): string {
```
vicb: Is `computePatchDiff` ever called with something that is not a string?

314systems: Not currently, but I think it should accept `RuleConfig` to stay consistent with the underlying `patchCode` signature. For example, a rule written as a YAML string:

```ts
export const buildIdRule = `
rule:
  pattern:
    selector: method_definition
    context: "class { getBuildId($$$PARAMS) { $$$_ } }"
fix: |-
  getBuildId($$$PARAMS) {
  }
`;
```

can be rewritten as:

```ts
export const buildIdRule = {
  rule: {
    pattern: {
      selector: "method_definition",
      context: "class { getBuildId($$$PARAMS) { $$$_ } }",
    },
  },
  fix: `getBuildId($$$PARAMS) {
}`,
} satisfies RuleConfig;
```

The `string | RuleConfig` type on `computePatchDiff` is there to support both forms.

vicb: We can change that if we ever need this.
My usual workflow is to tune rules on https://ast-grep.github.io/playground.html and copy the text into the source code.

314systems: The reason I made the change was that I tried writing tests for `getLoadManifestRule` and `getEvalManifestRule` using `computePatchDiff`, and hit a type error there since each of these functions returns `RuleConfig` (not `string`).

https://github.com/opennextjs/opennextjs-cloudflare/blob/d577521081365c6f9235d32959216f6db5e9268a/packages/cloudflare/src/cli/build/patches/plugins/load-manifest.ts#L36-L94

https://github.com/opennextjs/opennextjs-cloudflare/blob/d577521081365c6f9235d32959216f6db5e9268a/packages/cloudflare/src/cli/build/patches/plugins/load-manifest.ts#L96-L196

Happy to revert to string only if you'd prefer to keep things minimal for now.

vicb: > The reason I made the change was that I tried writing tests for getLoadManifestRule and getEvalManifestRule using computePatchDiff, and hit a type error there since each of these functions returns RuleConfig (not string).

Again you can change that when and if it is needed.
It's better to keep changes small and focused in a PR.
That is change related to the title/description of the PR.

Thanks.

packages/cloudflare/src/cli/build/patches/plugins/optional-deps.ts
```diff
@@ -24,7 +27,7 @@ export function handleOptionalDependencies(dependencies: string[]) {
 	return {
 		name,
 
-		setup: async (build: PluginBuild) => {
+		setup: async (build) => {
```
vicb: wrong change?
IMO this should be consistent with https://github.com/opennextjs/opennextjs-cloudflare/pull/1210/changes#diff-b23fad48f44b24ba602ee7e3fc4ed237515e8a257e6f317aa889b769e9769704L27

314systems: maybe wrong change

vicb:
@314systems I saw your review requests.

There are a few higher priority changes I'd like to get to before and I'll look into merging this after.

Thanks for your patience!

Issue #1211, opened 2026-04-20
Bundle size: unused @vercel/og dynamic import inflates Worker by ~2 MiB on free tier


mushan0x0:
### Summary

When an app does not use `next/og` (`ImageResponse`, `opengraph-image`, etc.), the built `.open-next/server-functions/default/handler.mjs` still contains a dynamic import:

```js
import("next/dist/compiled/@vercel/og/index.edge.js")
```

This comes from Next.js's internal `externalImport` helper and survives the existing `patchVercelOgImport` / `patchVercelOgFallbackFont` AST rules in `packages/cloudflare/src/cli/build/patches/ast/vercel-og.ts`, because those rules only rewrite the module specifier (`index.node.js` → `index.edge.js`), they don't strip the call site when `@vercel/og` is never actually reached at runtime.

Wrangler treats that dynamic import as a real module graph entry and pulls in:

- `@vercel/og` edge bundle (~800 KiB uncompressed / ~260 KiB gzip)
- `resvg.wasm` (~1.4 MiB)

…which, for a CSR-heavy app that doesn't use OG images at all, single-handedly pushes the Worker over the free-tier **3 MiB** limit.

### Current workaround

In our project we run a post-build script that rewrites the emitted handler before `wrangler deploy`:

```js
// scripts/patch-handler.mjs
source = source.replace(
  /import\(\"next\/dist\/compiled\/@vercel\/og\/index\.edge\.js\"\)/g,
  'Promise.reject(new Error(\"@vercel/og disabled\"))',
);
```

After this patch the Worker is ~1.6 MiB instead of ~3.8 MiB, and everything else (ProForm, admin UI, D1, R2, KV, Vectorize) works unchanged.

### Proposal

One of:

1. **Detect-and-stub automatically.** When the user's app does not import `next/og` (or more generally, when `ImageResponse` is not reachable), replace the `import(\"…@vercel/og/index.edge.js\")` call site with a rejected Promise so Wrangler no longer bundles the module + wasm.
2. **Opt-in flag** in `open-next.config.ts` / `@opennextjs/cloudflare` config, e.g. `cloudflare: { disableVercelOg: true }`, that performs the same rewrite. This keeps the default behavior safe and gives free-tier users a supported escape hatch.

I'd be happy to send a PR for option 2 (config flag + AST rule under `packages/cloudflare/src/cli/build/patches/ast/`) if maintainers are open to it — wanted to file the issue first to confirm the approach.

[...]

PR #1221, opened 2026-04-22
fix(cloudflare): stop bundling @vercel/og when it is unused

### LINKED ISSUE #1220: [FEATURE] Opt-in flag to strip unused @vercel/og dynamic import (saves ~2 MiB, keeps Worker under free-tier 3 MiB)

When an app does not use `next/og` (`ImageResponse`, `opengraph-image`, etc.), the built `.open-next/server-functions/default/handler.mjs` still contains a dynamic import:

```js
import("next/dist/compiled/@vercel/og/index.edge.js")
```

This comes from Next.js's internal `externalImport` helper and survives the existing `patchVercelOgImport` / `patchVercelOgFallbackFont` AST rules in `packages/cloudflare/src/cli/build/patches/ast/vercel-og.ts`, because those rules only rewrite the module specifier (`index.node.js` → `index.edge.js`); they don't strip the call site when `@vercel/og` is never actually reached at runtime.

Wrangler treats that dynamic import as a real module graph entry and pulls in:

- `@vercel/og` edge bundle (~800 KiB uncompressed / ~260 KiB gzip)
- `resvg.wasm` (~1.4 MiB)

For a CSR-heavy app that doesn't use OG images at all, this single dynamic import pushes the Worker over the Cloudflare free-tier **3 MiB** limit (observed ~3.8 MiB vs ~1.6 MiB after rewrite).

Add an opt-in flag to `@opennextjs/cloudflare`, e.g.:

```ts
// open-next.config.ts
  cloudflare: { disableVercelOg: true },
});
```

When enabled, an AST rule under `packages/cloudflare/src/cli/build/patches/ast/` rewrites the remaining `import("next/dist/compiled/@vercel/og/index.edge.js")` call sites to a rejected Promise (or a no-op stub), so Wrangler no longer bundles the `@vercel/og` edge module and `resvg.wasm`.

This keeps the default behavior safe and gives free-tier users a supported escape hatch without a post-build regex script.

Happy to send a PR (config flag + AST rule) if maintainers are open to it.

1. **Auto-detect and stub.** Scan user sources for `next/og` / `ImageResponse` / `opengraph-image` / `twitter-image` usage and, when none are found, automatically rewrite the call site. Safer defaults but more moving parts and edge cases.
2. **Post-build regex patch (current workaround).** We run a small script before `wrangler deploy`:

   ```js
   source = source.replace(
     /import\("next\/dist\/compiled\/@vercel\/og\/index\.edge\.js"\)/g,
     'Promise.reject(new Error("@vercel/og disabled"))',
   );
   ```

   After this patch our Worker is ~1.6 MiB instead of ~3.8 MiB and everything else (ProForm, admin UI, D1, R2, KV, Vectorize) works unchanged. But it's fragile — it depends on the exact emitted string and has to live in every project that hits the 3 MiB limit.
3. **Do nothing / upgrade to paid plan.** Not acceptable for free-tier / hobby projects that never use OG images.

1.19.2

### Additional context

- Next.js: `16.1.1`
- Wrangler: `4.83.0`
- Repro: any Next 15 / 16 app deployed via `@opennextjs/cloudflare` that does **not** import `next/og`. Compare `.open-next/server-functions/default/handler.mjs` and `wrangler deploy --dry-run` size with and without the regex rewrite above.
- Previous issue closed by maintainer asking to re-file with the template: #1211.

### Before submitting

- [x] I have checked that there isn't already a similar feature request
- [x] This is a single feature (not multiple features in one request)

mushan0x0:
Fixes #1220 (and the original #1211).

## Problem

Next.js's internal `externalImport` helper emits a dynamic

```js
import("next/dist/compiled/@vercel/og/index.edge.js")
```

into the app handler even for apps that never use `next/og` (`ImageResponse`, `opengraph-image`, `twitter-image`, …). Today, when `patchVercelOgLibrary` decides `useOg === false`, `bundle-server.ts` marks that module as `external`:

```ts
external: [
  "./middleware/handler.mjs",
  // Do not bundle og when it is not used
  ...(useOg ? [] : ["next/dist/compiled/@vercel/og/index.edge.js"]),
],
```

esbuild honors `external` and leaves the `import("…@vercel/og/index.edge.js")` call site intact. Wrangler then resolves that external module against `node_modules` and bundles:

- `@vercel/og` edge bundle (~800 KiB uncompressed / ~260 KiB gzip)
- `resvg.wasm` (~1.4 MiB)

…even though the code path is unreachable. For CSR-heavy apps that never use OG images this single dynamic import pushes the Worker over the **3 MiB** free-tier gzip limit.

Reproduced on a real Next 16 app with `@opennextjs/cloudflare@1.19.1` and `1.19.2`:

| build                              | `wrangler deploy --dry-run` |
|------------------------------------|-----------------------------|
| unpatched (`useOg=false`, external)| `14098.57 KiB / gzip 3119.42 KiB` ❌ |
| after this PR                      | `11875.51 KiB / gzip 2415.74 KiB` ✅ |

(The same savings were previously achievable only via a post-build regex script that rewrote the emitted handler.)

## Fix

When `useOg === false`, alias the edge entry to the existing `throw.js` shim instead of marking it external:

```ts
external: ["./middleware/handler.mjs"],
alias: {
  ...(useOg
    ? {}
    : {
        "next/dist/compiled/@vercel/og/index.edge.js": path.join(
          buildOpts.outputDir,
          "cloudflare-templates/shims/throw.js"
        ),
      }),
  // …
},
```

The unreachable dynamic import now resolves to a tiny `throw "OpenNext shim"` module, and Wrangler no longer drags the real `@vercel/og` library + `resvg.wasm` into the Worker bundle. If the dynamic import is ever executed at runtime (which would indicate a bug in `patchVercelOgLibrary`'s `useOg` detection), the shim throws immediately — the same failure mode users already get from `@ampproject/toolbox-optimizer` and other shimmed modules, rather than a silent runtime mismatch.

When `useOg === true` the behavior is unchanged: the real edge build is copied and patched by `patchVercelOgLibrary` as before.

## Why not a config flag?

The original issue proposed an opt-in flag (`cloudflare: { disableVercelOg: true }`), but the adapter already computes the exact signal it needs (`useOg`), so this can be done safely with no new config surface.

## Validation

- `pnpm --filter cloudflare test` — 31 files / 317 tests pass.
- `pnpm code:checks` (prettier + eslint + tsc) — clean.
- End-to-end: built a real app (Next 16.2.4, `app/` router, no `next/og`) against this fork, confirmed `handler.mjs` no longer contains `next/dist/compiled/@vercel/og/index.edge.js` and `wrangler deploy --dry-run` size dropped as shown above.

## Changeset

Included as a `patch` bump under `.changeset/drop-unused-vercel-og.md`.

### DIFF
--- a/.changeset/drop-unused-vercel-og.md
+++ b/.changeset/drop-unused-vercel-og.md
@@ -1,0 +1,9 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+Stop bundling `@vercel/og` (and its ~1.4 MiB `resvg.wasm`) when the app does not use it.
+
+Next.js's `externalImport` helper keeps a dynamic `import("next/dist/compiled/@vercel/og/index.edge.js")` in the emitted handler even for apps that never use `ImageResponse` / `opengraph-image`. Previously this module was marked as `external` when `useOg` was `false`, which left Wrangler to resolve and bundle it — pulling in ~800 KiB of JS plus `resvg.wasm` and pushing many Workers over the Cloudflare free-tier 3 MiB gzip limit.
+
+When `useOg` is `false`, the edge entry is now aliased to the existing `throw.js` shim, so the unreachable dynamic import resolves to a tiny module and the real `@vercel/og` library is no longer pulled into the Worker bundle.
--- a/packages/cloudflare/src/cli/build/bundle-server.ts
+++ b/packages/cloudflare/src/cli/build/bundle-server.ts
@@ -113,12 +113,20 @@ export async function bundleServer(buildOpts: BuildOptions, projectOpts: Project
 			// Apply updater updates, must be the last plugin
 			updater.plugin,
 		] as Plugin[],
-		external: [
-			"./middleware/handler.mjs",
-			// Do not bundle og when it is not used
-			...(useOg ? [] : ["next/dist/compiled/@vercel/og/index.edge.js"]),
-		],
+		external: ["./middleware/handler.mjs"],
 		alias: {
+			// When @vercel/og is not used, alias the edge entry to a throwing shim so the
+			// dynamic `import("next/dist/compiled/@vercel/og/index.edge.js")` call site
+			// emitted by Next.js does not drag the library (~800 KiB) and its
+			// `resvg.wasm` (~1.4 MiB) into the Worker bundle.
+			...(useOg
+				? {}
+				: {
+						"next/dist/compiled/@vercel/og/index.edge.js": path.join(
+							buildOpts.outputDir,
+							"cloudflare-templates/shims/throw.js"
+						),
+					}),
 			// Workers have `fetch` so the `node-fetch` polyfill is not needed
 			"next/dist/compiled/node-fetch": path.join(buildOpts.outputDir, "cloudflare-templates/shims/fetch.js"),
 			// Workers have builtin Web Sockets


vicb (approved):
Great investigation!

LGTM, thank you so much for the fix!

PR #1234, opened 2026-04-27
refactor: use native Node glob APIs

314systems:
Use Node.js native glob APIs for cache population and build-time patch discovery instead of the `glob` package. This keeps path handling based on explicit working directories and removes the direct `glob` dependency.

### DIFF
--- a/.changeset/glob.md
+++ b/.changeset/glob.md
@@ -1,0 +1,7 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+refactor: use native Node glob APIs
+
+Use Node.js native glob APIs for cache population and build-time patch discovery instead of the `glob` package. This keeps path handling based on explicit working directories and removes the direct `glob` dependency.
--- a/.changeset/glob.md
+++ b/.changeset/glob.md
@@ -5,3 +5,5 @@
 refactor: use native Node glob APIs
 
 Use Node.js native glob APIs for cache population and build-time patch discovery instead of the `glob` package. This keeps path handling based on explicit working directories and removes the direct `glob` dependency.
+
+Node.js `>=22.17.0 <23 || >=24.1.0` is now required because `fs.promises.glob` — used for cache population and patch discovery — was stabilized in those releases.
--- a/packages/cloudflare/package.json
+++ b/packages/cloudflare/package.json
@@ -59,7 +59,6 @@
 		"cloudflare": "^4.4.1",
 		"comment-json": "^4.5.1",
 		"enquirer": "^2.4.1",
-		"glob": "catalog:",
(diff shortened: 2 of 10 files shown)


vicb (commented):
Thanks for the PR!

What is the minimal version of Node required?
Please answer the question in the PR description and by adding an `engines` entry in package.json

Also please do not change async <> sync if there is no good reason to do so - if there are please do that in a follow up PR explaining the rationale in the description

.changeset/glob.md
```diff
@@ -0,0 +1,7 @@
+---
```
vicb, line 1: We do not really need a changeset for that if there are no user facing change

314systems:
> What is the minimal version of Node required?

https://nodejs.org/api/fs.html#fspromisesglobpattern-options

Added in: v22.0.0

Version | Changes
-- | --
v24.1.0, v22.17.0 | Add support for URL instances for **cwd** option.
v24.0.0, v22.17.0 | **Marking the API stable.**
v23.7.0, v22.14.0 | Add support for exclude option to accept glob patterns.
v22.2.0 | Add support for withFileTypes as an option.

v22.17.0 (or v24.1.0) is.

Node v20 is reaching EOL soon (April 2026), so I think we can safely drop support for it.

james-elicx:
> Node v20 is reaching EOL soon, so I think we can safely drop support for it.

If I recall correctly, I believe the plan was for the minimum Node version to be bumped as part of the next major. The AWS adapter still supports Node.js v20 as well.

For what it's worth, many companies will use a Node.js version right up until it EOLs, and continue using it beyond that.

.changeset/glob.md
314systems: I think there are no changes facing users directly.

However, since this PR defines `engines`, I thought it should be submitted as a patch.

vicb:
> > Node v20 is reaching EOL soon, so I think we can safely drop support for it.
> 
> If I recall correctly, I believe the plan was for the minimum Node version to be bumped as part of the next major. The AWS adapter still supports Node.js v20 as well.
> 
> For what it's worth, many companies will use a Node.js version right up until it EOLs, and continue using it beyond that.

Good point @james-elicx.

I'll move this PR to draft for now.
I don't think we need to rush this change before the next major as it might break users.

314systems:
(Note: Node 20 is now [EOL](https://nodejs.org/en/about/eol))

sommeeeer:
I can confirm what @james-elicx is saying. One of the companies I work for are still using Node v20 in AWS Lambdas even though it has reached end of support recently. Looking in AWS [docs](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html) I see that we should probably update soon.

image

PR #1238, opened 2026-04-28
refactor: remove unnecessary optional chaining, nullish coalescing, and type assertions

314systems:
Cleans up several spots in `packages/cloudflare` where optional chaining (`?.`),
nullish coalescing (`??`), or non-null/type assertions (`!`, `as`) were guarding
against states the types already preclude.

No behavioural changes intended.

(Note: Investigated by switching the ESLint config to `tseslint.configs.strictTypeChecked` and addressing the resulting warnings.)

---

<a href="https://app.devin.ai/review/opennextjs/opennextjs-cloudflare/pull/1238" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

PR #1239, opened 2026-04-29
misc minor refactoring

vicb:
Supersedes https://github.com/opennextjs/opennextjs-cloudflare/pull/1238

@314systems I have extracted a few "safe" changes from your PR.

I guess some of the changes there might not be compatible with the minimum version of TS from package.json.
So it would take a long time to check every single change.

I also don't think there is a urge to change working code when there is no clear benefit / readability gain. We can change the code when we have to update it for other reason. We don't have unlimited resources to work on this project and allocating them to higher priority bug fixes / features is a better thing to do IMO.

Thanks!

---

<a href="https://app.devin.ai/review/opennextjs/opennextjs-cloudflare/pull/1239" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


james-elicx (approved):

vicb:
Thanks for the review James!

PR #1248, opened 2026-05-04
fix: harden runWrangler against shell injection and silence DEP0190

kirkouimet:
## Summary

Closes #1182.

`runWrangler` in `packages/cloudflare/src/cli/commands/utils/run-wrangler.ts` previously called `spawnSync` with `shell: true` while interpolating flag values into single-string array entries:

```ts
spawnSync(options.packager, [
  options.packager === "bun" ? "x" : "exec",
  "wrangler",
  ...injectPassthroughFlagForArgs(options, [
    ...args,
    wranglerOpts.environment && `--env ${wranglerOpts.environment}`,  // ⚠️ single concatenated string
    wranglerOpts.configPath && `--config ${wranglerOpts.configPath}`,  // ⚠️ single concatenated string
    ...
  ].filter((v): v is string => !!v)),
], { shell: true, ... });
```

With `shell: true`, Node.js joins the args array into a shell command string without escaping, so a value containing shell metacharacters (`;`, `&&`, `$(...)`, backticks, etc.) can break out of its intended argument boundary and execute arbitrary commands. Node.js 22 added [`DEP0190`](https://nodejs.org/api/deprecations.html#DEP0190) specifically to flag this pattern, and that warning surfaces on every `opennextjs-cloudflare deploy` invoked through `wrangler-action` on Node 22+:

```
(node:NNNN) [DEP0190] DeprecationWarning: Passing args to a child process with shell option true
can lead to security vulnerabilities, as the arguments are not escaped, only concatenated.
```

## Fix

- Each flag/value pair is now passed as separate array entries (`["--env", value]`, `["--config", value]`, etc.) so values are never interpolated into a shell string.
- `shell` is set to `false` on POSIX, where the package manager binary can be invoked directly. This silences `DEP0190` and removes the injection vector entirely.
- `shell: true` is retained on Windows so the package manager's `.cmd` shim still resolves via `cmd.exe`. With each value as a separate array entry, Node escapes them correctly even when the shell is involved, so the injection vector is closed on every platform.

## Test plan

- [x] `pnpm code:checks` (prettier + eslint + tsc) passes
- [x] `pnpm -r test` — **32 files, 326 tests, all green**
- [x] End-to-end smoke test: invoking the patched `runWrangler` to call `wrangler --version` against a real project on **Node 24.14.1** returns `4.87.0` with **no `DEP0190` emitted** (previously emitted)
- [x] Security regression test: a value of `staging$(echo INJECTED > /tmp/pwned.txt)` no longer executes the injected subshell — the string is passed verbatim as an argument and the file is never created (with the old code, the file IS created)
- [ ] CI on Windows runners (covered by repo's existing matrix)

---

<a href="https://app.devin.ai/review/opennextjs/opennextjs-cloudflare/pull/1248" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/fix-runwrangler-shell-injection.md
+++ b/.changeset/fix-runwrangler-shell-injection.md
@@ -3,19 +3,3 @@
 ---
 
 fix: harden `runWrangler` against shell injection and silence Node.js DEP0190.
-
-`runWrangler` previously called `spawnSync` with `shell: true` while interpolating
-flag values into single-string array entries (e.g. `` `--env ${wranglerOpts.environment}` ``).
-With `shell: true`, Node.js joins the arguments into a shell command string without
-escaping, so a value containing shell metacharacters (`;`, `&&`, `$(...)`, backticks,
-etc.) could break out of its intended argument boundary. Node.js 22 added `DEP0190`
-specifically to flag this pattern, and that warning surfaced on every `opennextjs-cloudflare deploy`
-invoked through `wrangler-action` on Node 22+.
-
-Each flag/value pair is now passed as separate array entries, and `shell` is set to
-`false` on POSIX (where the package manager binary can be invoked directly). `shell: true`
-is retained on Windows so the package manager's `.cmd` shim still resolves via `cmd.exe`,
-but values are no longer interpolated into the shell command, so the injection vector
-is closed on every platform.
-
-Closes [#1182](https://github.com/opennextjs/opennextjs-cloudflare/issues/1182).
--- a/.changeset/fix-runwrangler-shell-injection.md
+++ b/.changeset/fix-runwrangler-shell-injection.md
@@ -1,0 +1,21 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+fix: harden `runWrangler` against shell injection and silence Node.js DEP0190.
+
+`runWrangler` previously called `spawnSync` with `shell: true` while interpolating
+flag values into single-string array entries (e.g. `` `--env ${wranglerOpts.environment}` ``).
+With `shell: true`, Node.js joins the arguments into a shell command string without
+escaping, so a value containing shell metacharacters (`;`, `&&`, `$(...)`, backticks,
+etc.) could break out of its intended argument boundary. Node.js 22 added `DEP0190`
+specifically to flag this pattern, and that warning surfaced on every `opennextjs-cloudflare deploy`
+invoked through `wrangler-action` on Node 22+.
+
+Each flag/value pair is now passed as separate array entries, and `shell` is set to
+`false` on POSIX (where the package manager binary can be invoked directly). `shell: true`
+is retained on Windows so the package manager's `.cmd` shim still resolves via `cmd.exe`,
+but values are no longer interpolated into the shell command, so the injection vector
+is closed on every platform.
+
+Closes [#1182](https://github.com/opennextjs/opennextjs-cloudflare/issues/1182).
--- a/packages/cloudflare/src/cli/commands/deploy.ts
+++ b/packages/cloudflare/src/cli/commands/deploy.ts
@@ -4,7 +4,7 @@ import type yargs from "yargs";
 import { DEPLOYMENT_MAPPING_ENV_NAME } from "../templates/skew-protection.js";
 import { populateCache, withPopulateCacheOptions } from "./populate-cache.js";
 import { getDeploymentMapping } from "./skew-protection.js";
-import { getEnvFromPlatformProxy, quoteShellMeta } from "./utils/helpers.js";
+import { getEnvFromPlatformProxy } from "./utils/helpers.js";
 import { runWrangler } from "./utils/run-wrangler.js";
 import type { WithWranglerArgs } from "./utils/utils.js";
 import {
@@ -57,16 +57,8 @@ export async function deployCommand(args: WithWranglerArgs<{ cacheChunkSize?: nu
 		[
 			"deploy",
 			...args.wranglerArgs,
-			// The `--var` value is `KEY:JSON` where the JSON object contains `{`, `}`, `"`, etc.
-			// On Windows `runWrangler` keeps `shell: true`, so the value must be quoted to survive
-			// `cmd.exe` tokenization. On POSIX `shell: false` means it is passed verbatim.
 			...(deploymentMapping
-				? [
-						"--var",
-						process.platform === "win32"
-							? quoteShellMeta(`${DEPLOYMENT_MAPPING_ENV_NAME}:${JSON.stringify(deploymentMapping)}`)
-							: `${DEPLOYMENT_MAPPING_ENV_NAME}:${JSON.stringify(deploymentMapping)}`,
-					]
+				? ["--var", `${DEPLOYMENT_MAPPING_ENV_NAME}:${JSON.stringify(deploymentMapping)}`]
 				: []),
 		],
 		{
--- a/packages/cloudflare/src/cli/commands/deploy.ts
+++ b/packages/cloudflare/src/cli/commands/deploy.ts
@@ -4,7 +4,7 @@ import type yargs from "yargs";
 import { DEPLOYMENT_MAPPING_ENV_NAME } from "../templates/skew-protection.js";
 import { populateCache, withPopulateCacheOptions } from "./populate-cache.js";
 import { getDeploymentMapping } from "./skew-protection.js";
-import { getEnvFromPlatformProxy, quoteShellMeta } from "./utils/helpers.js";
+import { getEnvFromPlatformProxy } from "./utils/helpers.js";
 import { runWrangler } from "./utils/run-wrangler.js";
 import type { WithWranglerArgs } from "./utils/utils.js";
 import {
@@ -58,7 +58,7 @@ export async function deployCommand(args: WithWranglerArgs<{ cacheChunkSize?: nu
 			"deploy",
 			...args.wranglerArgs,
 			...(deploymentMapping
-				? [`--var ${DEPLOYMENT_MAPPING_ENV_NAME}:${quoteShellMeta(JSON.stringify(deploymentMapping))}`]
+				? ["--var", `${DEPLOYMENT_MAPPING_ENV_NAME}:${JSON.stringify(deploymentMapping)}`]
 				: []),
 		],
 		{
--- a/packages/cloudflare/src/cli/commands/deploy.ts
+++ b/packages/cloudflare/src/cli/commands/deploy.ts
@@ -4,7 +4,7 @@ import type yargs from "yargs";
 import { DEPLOYMENT_MAPPING_ENV_NAME } from "../templates/skew-protection.js";
 import { populateCache, withPopulateCacheOptions } from "./populate-cache.js";
 import { getDeploymentMapping } from "./skew-protection.js";
-import { getEnvFromPlatformProxy } from "./utils/helpers.js";
+import { getEnvFromPlatformProxy, quoteShellMeta } from "./utils/helpers.js";
 import { runWrangler } from "./utils/run-wrangler.js";
 import type { WithWranglerArgs } from "./utils/utils.js";
 import {
@@ -57,8 +57,16 @@ export async function deployCommand(args: WithWranglerArgs<{ cacheChunkSize?: nu
 		[
 			"deploy",
 			...args.wranglerArgs,
+			// The `--var` value is `KEY:JSON` where the JSON object contains `{`, `}`, `"`, etc.
+			// On Windows `runWrangler` keeps `shell: true`, so the value must be quoted to survive
+			// `cmd.exe` tokenization. On POSIX `shell: false` means it is passed verbatim.
 			...(deploymentMapping
-				? ["--var", `${DEPLOYMENT_MAPPING_ENV_NAME}:${JSON.stringify(deploymentMapping)}`]
+				? [
+						"--var",
+						process.platform === "win32"
+							? quoteShellMeta(`${DEPLOYMENT_MAPPING_ENV_NAME}:${JSON.stringify(deploymentMapping)}`)
+							: `${DEPLOYMENT_MAPPING_ENV_NAME}:${JSON.stringify(deploymentMapping)}`,
+					]
 				: []),
 		],
 		{
--- a/packages/cloudflare/src/cli/commands/populate-cache.ts
+++ b/packages/cloudflare/src/cli/commands/populate-cache.ts
@@ -43,7 +43,7 @@ import {
 import { ensureR2Bucket } from "../utils/ensure-r2-bucket.js";
 import { normalizePath } from "../utils/normalize-path.js";
 import type { R2Response } from "../workers/r2-cache-types.js";
-import { getEnvFromPlatformProxy, type WorkerEnvVar } from "./utils/helpers.js";
+import { getEnvFromPlatformProxy, quoteShellMeta, type WorkerEnvVar } from "./utils/helpers.js";
 import type { WranglerTarget } from "./utils/run-wrangler.js";
 import { runWrangler } from "./utils/run-wrangler.js";
 import type { WithWranglerArgs } from "./utils/utils.js";
@@ -533,7 +533,10 @@ async function populateKVIncrementalCache(
 				"kv",
 				"bulk",
 				"put",
-				chunkPath,
+				// chunkPath may contain spaces on Windows (e.g. `C:\Users\Some User\...`), and
+				// `runWrangler` keeps `shell: true` on Windows for `.cmd` resolution, so quote
+				// metacharacters there. On POSIX `shell: false` means the value is passed verbatim.
+				process.platform === "win32" ? quoteShellMeta(chunkPath) : chunkPath,
 				"--binding",
 				KV_CACHE_BINDING_NAME,
 				populateCacheOptions.shouldUsePreviewId ? "--preview" : "--no-preview",
@@ -570,6 +573,11 @@ function populateD1TagCache(
 		throw new Error(`No D1 binding "${D1_TAG_BINDING_NAME}" found!`);
 	}
 
+	// SQL values contain parens, semicolons, and spaces. On Windows `runWrangler` keeps
+	// `shell: true` so `cmd.exe` would tokenize unquoted SQL as separate args. Quote on
+	// Windows; pass through verbatim on POSIX where `shell: false` is set.
+	const quoteForShell = (sql: string) => (process.platform === "win32" ? quoteShellMeta(sql) : sql);
+
 	const result = runWrangler(
 		buildOpts,
 		[
@@ -582,7 +590,9 @@ function populateD1TagCache(
 			//   stale         - Timestamp (ms) when the cached entry becomes stale. Added in v1.19.
 			//   expire        - Timestamp (ms) when the cached entry expires. NULL means no expire. Added in v1.19.
 			"--command",
-			"CREATE TABLE IF NOT EXISTS revalidations (tag TEXT NOT NULL, revalidatedAt INTEGER NOT NULL, stale INTEGER, expire INTEGER default NULL, UNIQUE(tag) ON CONFLICT REPLACE);",
+			quoteForShell(
+				"CREATE TABLE IF NOT EXISTS revalidations (tag TEXT NOT NULL, revalidatedAt INTEGER NOT NULL, stale INTEGER, expire INTEGER default NULL, UNIQUE(tag) ON CONFLICT REPLACE);"
+			),
 			populateCacheOptions.shouldUsePreviewId ? "--preview" : "--no-preview",
 		],
 		{
@@ -607,7 +617,9 @@ function populateD1TagCache(
 			"execute",
 			D1_TAG_BINDING_NAME,
 			"--command",
-			"ALTER TABLE revalidations ADD COLUMN stale INTEGER; ALTER TABLE revalidations ADD COLUMN expire INTEGER default NULL",
+			quoteForShell(
+				"ALTER TABLE revalidations ADD COLUMN stale INTEGER; ALTER TABLE revalidations ADD COLUMN expire INTEGER default NULL"
+			),
 			populateCacheOptions.shouldUsePreviewId ? "--preview" : "--no-preview",
 		],
 		{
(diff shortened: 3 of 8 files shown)


kirkouimet:
Great catch from Devin Review — that was a real bug. Pushed [5df6c61](https://github.com/opennextjs/opennextjs-cloudflare/pull/1248/commits/5df6c61) splitting every multi-word arg and `--flag value` pair at the three affected call sites:

- `populate-cache.ts`: `"kv bulk put"` → `"kv", "bulk", "put"`; `"--binding ${X}"` → `"--binding", X`; `"--preview ${X}"` → `"--preview", String(X)`; same treatment for the two `d1 execute` calls (and removed the stray quotes that were wrapping the `--command` SQL).
- `deploy.ts` and `upload.ts`: `"versions upload"` → `"versions", "upload"`; `--var` pair split; `quoteShellMeta` wrapper dropped since `shell: false` means args go to the process verbatim and shell-escaping would inject literal characters into the value.
- Dropped the now-unused `quoteShellMeta` imports from those three files. Left the function in `helpers.ts` since it might be relied on externally — happy to remove that too if maintainers prefer.

Re-verified locally: `pnpm code:checks` (prettier + eslint + tsc) green, `pnpm -r test` 326/326 pass, and an end-to-end invocation of the rebuilt `runWrangler` against a real wrangler CLI with `["versions", "list", "--name", "..."]` reaches wrangler's auth check (i.e. arg parsing succeeds) with no `DEP0190` emitted.

packages/cloudflare/src/cli/commands/utils/run-wrangler.ts
```diff
@@ -102,19 +102,22 @@ export function runWrangler(
...
 		],
 		{
-			shell: true,
+			// Each flag value is now a separate array entry, so `shell: true` is no longer
+			// needed for value interpolation and would only re-introduce the injection vector
+			// flagged by Node.js DEP0190 (and the related security advisory). `shell: true` is
+			// still required on Windows so that the package manager's `.cmd` shim is resolved
+			// via cmd.exe; with each arg as a separate entry, Node escapes them correctly even
+			// when the shell is involved.
+			shell: process.platform === "win32",
```
vicb: Could we still use `shell: true` now that args are separated?

packages/cloudflare/src/cli/commands/deploy.ts
```diff
@@ -4,7 +4,7 @@ import type yargs from "yargs";
 import { DEPLOYMENT_MAPPING_ENV_NAME } from "../templates/skew-protection.js";
 import { populateCache, withPopulateCacheOptions } from "./populate-cache.js";
 import { getDeploymentMapping } from "./skew-protection.js";
-import { getEnvFromPlatformProxy, quoteShellMeta } from "./utils/helpers.js";
```
vicb, line 7: is `quoteShellMeta` still use anywhere? If not, we can remove that

packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -570,15 +574,18 @@ function populateD1TagCache(
...
 			//   tag           - The cache tag.
 			//   revalidatedAt - Timestamp (ms) when the tag was last revalidated.
 			//   stale         - Timestamp (ms) when the cached entry becomes stale. Added in v1.19.
 			//   expire        - Timestamp (ms) when the cached entry expires. NULL means no expire. Added in v1.19.
-			`--command "CREATE TABLE IF NOT EXISTS revalidations (tag TEXT NOT NULL, revalidatedAt INTEGER NOT NULL, stale INTEGER, expire INTEGER default NULL, UNIQUE(tag) ON CONFLICT REPLACE);"`,
-			`--preview ${populateCacheOptions.shouldUsePreviewId}`,
+			"--command",
+			"CREATE TABLE IF NOT EXISTS revalidations (tag TEXT NOT NULL, revalidatedAt INTEGER NOT NULL, stale INTEGER, expire INTEGER default NULL, UNIQUE(tag) ON CONFLICT REPLACE);",
+			"--preview",
+			String(populateCacheOptions.shouldUsePreviewId),
```
vicb: What about:
```suggestion
			populateCacheOptions.shouldUsePreviewId ? "--preview" : "--no-preview",
```

ref: https://yargs.js.org/docs/#api-reference-booleankey

packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -598,10 +605,13 @@ function populateD1TagCache(
...
-			"d1 execute",
+			"d1",
+			"execute",
 			D1_TAG_BINDING_NAME,
-			`--command "ALTER TABLE revalidations ADD COLUMN stale INTEGER; ALTER TABLE revalidations ADD COLUMN expire INTEGER default NULL"`,
-			`--preview ${populateCacheOptions.shouldUsePreviewId}`,
+			"--command",
+			"ALTER TABLE revalidations ADD COLUMN stale INTEGER; ALTER TABLE revalidations ADD COLUMN expire INTEGER default NULL",
+			"--preview",
+			String(populateCacheOptions.shouldUsePreviewId),
```
vicb: ditto

vicb (commented):
Thanks for the PR!

I have added some inline comments.

@james-elicx has done a lot of work on the CLI, I'd love to get some insights from him too

packages/cloudflare/src/cli/commands/utils/run-wrangler.ts
vicb: Oh now I understand this would trigger DEP0190.

Could we please reword the comment to something like:

Setting `shell` to `false` to avoid DEP0190
See https://nodejs.org/api/deprecations.html#dep0190-passing-args-to-nodechild-process-execfilespawn-with-shell-option

Note that `shell: true` is required on Windows so that the package manager's `.cmd` shim is resolved

.changeset/fix-runwrangler-shell-injection.md
```diff
@@ -0,0 +1,21 @@
...
+specifically to flag this pattern, and that warning surfaced on every `opennextjs-cloudflare deploy`
+invoked through `wrangler-action` on Node 22+.
+
+Each flag/value pair is now passed as separate array entries, and `shell` is set to
+`false` on POSIX (where the package manager binary can be invoked directly). `shell: true`
+is retained on Windows so the package manager's `.cmd` shim still resolves via `cmd.exe`,
+but values are no longer interpolated into the shell command, so the injection vector
+is closed on every platform.
+
+Closes [#1182](https://github.com/opennextjs/opennextjs-cloudflare/issues/1182).
```
vicb: This is used for the changelog.
I don't think there is a need to give details as the changes in this PR will not affect users (the behavior stays the same).

As long as the changes are described in the PR, we should be good.

Thanks!

```suggestion
fix: harden `runWrangler` against shell injection and silence Node.js DEP0190.
```

kirkouimet:
Thanks for the review! Pushed [227371e](https://github.com/opennextjs/opennextjs-cloudflare/pull/1248/commits/227371e) addressing all four points. Re-verified locally: `pnpm code:checks` green, `pnpm -r test` 326/326 pass, end-to-end invocation against a real wrangler CLI parses each token correctly with no `DEP0190` emitted, and a security regression test confirms `--env 'staging$(echo INJECTED > /tmp/pwned.txt)'` no longer evaluates the subshell.

### `shell: true` with separated args

Tried it. Doesn't work for two reasons:

1. `shell: true` still emits `DEP0190` regardless of how the args were passed in. Node concatenates `[cmd, ...args].join(' ')` into a shell command string before spawning, and that's what the deprecation flags. So going back to `shell: true` un-fixes the original warning.
2. The shell then re-tokenizes the concatenated string, which breaks values containing whitespace or shell metacharacters. Concrete repro:
    ```js
    spawnSync('printf', ['%s\n', '--command', 'CREATE TABLE foo (x INT);'], {shell: true})
    // /bin/sh: -c: syntax error near unexpected token '('
    ```
    Our `d1 execute --command "CREATE TABLE ..."` would die on the parens.

So `shell: false` on POSIX is load-bearing for both the warning fix and to let the SQL `--command` value pass through intact.

### `quoteShellMeta`

Removed. Verified there are no other call sites and dropped the test mock entry too.

### `--preview` / `--no-preview`

Applied at all three call sites. The third one (the `kv bulk put` call at line 539) had the same `--preview ${bool}` pattern, so I converted it for consistency. Let me know if you'd rather I revert that one.

### One follow-up worth flagging

Per Node's [`child_process` docs](https://nodejs.org/api/child_process.html#shell-requirements), `shell: true` for `.cmd` resolution on Windows is "not recommended, see DEP0190". The Windows carve-out we kept (`shell: process.platform === "win32"`) means Windows users will still see the deprecation warning, even though the injection vector is now closed on every platform.

The clean fix is to switch this `spawnSync` to [`cross-spawn`](https://github.com/moxystudio/node-cross-spawn) (already a transitive dep, ~10KB, what npm/yarn/eslint/prettier use), which would let us drop the platform branch and have `shell: false` everywhere. I held off on that since (a) you didn't ask for it and (b) the repo's CI is ubuntu-only so I can't verify the Windows path locally. Happy to do a focused follow-up PR if you'd like.

vicb:
@kirkouimet thank you so much!

## shell: true with separated args

IIUC 2 says `shell: true` would not work, still it is used on windows. What's different there?

Maybe also linked to your last point?
I'd rather avoid a package that has not been updated in 2y.

Probably a point to discuss more before merging this PR

## quoteShellMeta

:+1:

## --preview / --no-preview

Consistency is the best, thanks!

packages/cloudflare/src/cli/commands/utils/helpers.ts
```diff
@@ -71,35 +71,3 @@ export async function getEnvFromPlatformProxy(options: GetPlatformProxyOptions,
...
-			return '""';
-		}
-		const needsEscaping = /[&|<>^()%!"]/;
-		const needsQuotes = /\s/.test(arg) || needsEscaping.test(arg);
-		let escaped = arg.replace(/"/g, '""');
-		if (/[&|<>^()%!]/.test(arg)) {
-			escaped = escaped.replace(/[&|<>^()%!]/g, "^$&");
-		}
-		return needsQuotes ? `"${escaped}"` : escaped;
-	}
```
james-elicx, line 97: we'll likely want to keep this given that it still uses shell: true on win32

packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -570,15 +573,17 @@ function populateD1TagCache(
...
 			D1_TAG_BINDING_NAME,
 			// Columns:
 			//   tag           - The cache tag.
 			//   revalidatedAt - Timestamp (ms) when the tag was last revalidated.
 			//   stale         - Timestamp (ms) when the cached entry becomes stale. Added in v1.19.
 			//   expire        - Timestamp (ms) when the cached entry expires. NULL means no expire. Added in v1.19.
-			`--command "CREATE TABLE IF NOT EXISTS revalidations (tag TEXT NOT NULL, revalidatedAt INTEGER NOT NULL, stale INTEGER, expire INTEGER default NULL, UNIQUE(tag) ON CONFLICT REPLACE);"`,
-			`--preview ${populateCacheOptions.shouldUsePreviewId}`,
+			"--command",
+			"CREATE TABLE IF NOT EXISTS revalidations (tag TEXT NOT NULL, revalidatedAt INTEGER NOT NULL, stale INTEGER, expire INTEGER default NULL, UNIQUE(tag) ON CONFLICT REPLACE);",
```
james-elicx, line 585: i think the sql statements might still need to be quoted for windows?

.changeset/fix-runwrangler-shell-injection.md
```diff
@@ -0,0 +1,21 @@
...
+escaping, so a value containing shell metacharacters (`;`, `&&`, `$(...)`, backticks,
+etc.) could break out of its intended argument boundary. Node.js 22 added `DEP0190`
+specifically to flag this pattern, and that warning surfaced on every `opennextjs-cloudflare deploy`
+invoked through `wrangler-action` on Node 22+.
+
+Each flag/value pair is now passed as separate array entries, and `shell` is set to
+`false` on POSIX (where the package manager binary can be invoked directly). `shell: true`
+is retained on Windows so the package manager's `.cmd` shim still resolves via `cmd.exe`,
+but values are no longer interpolated into the shell command, so the injection vector
+is closed on every platform.
```
james-elicx: isn't it still left open on windows?

kirkouimet:
Took another swing at this. Pushed [382e5d1](https://github.com/opennextjs/opennextjs-cloudflare/pull/1248/commits/382e5d1) which silences DEP0190 on Windows too, plus [9f39728](https://github.com/opennextjs/opennextjs-cloudflare/pull/1248/commits/9f39728) addressing the rest of the review feedback. End result: DEP0190 is silenced on every platform, the injection vector is closed everywhere, and call sites don't need to know about Windows.

### Windows DEP0190 (the asymmetry @vicb flagged)

Per Node.js docs ([Spawning .bat and .cmd files on Windows](https://nodejs.org/api/child_process.html#spawning-bat-and-cmd-files-on-windows)), the recommended pattern is to spawn `cmd.exe` directly with `/c` rather than relying on `shell: true`. The Windows path in `runWrangler` now does that:

```ts
spawnSync("cmd.exe", ["/d", "/s", "/c", `"${escapedCommandLine}"`], {
    shell: false,
    windowsVerbatimArguments: true,
    ...
});
```

Where `escapedCommandLine` is the joined, `quoteShellMeta`-escaped tokens. `windowsVerbatimArguments: true` tells Node not to add its own escaping on top of ours. This mirrors how `cross-spawn` handles non-`.exe` commands internally, without taking on `cross-spawn` as a dependency (your concern about its release cadence still stands).

POSIX path is unchanged from the previous round: `spawnSync(packager, args, { shell: false })`, no escaping needed because each value is a separate argv entry.

### `quoteShellMeta` (back, but moved)

@james-elicx was right that removing it was a Windows regression. It's restored in `helpers.ts`, but with one important change: it's now used **internally by `runWrangler`** rather than at every call site. The function escapes every token once, when constructing the `cmd.exe /c` command line. This means:

- `populate-cache.ts`, `deploy.ts`, and `upload.ts` revert to passing plain string arrays. No `process.platform === "win32" ? quoteShellMeta(x) : x` boilerplate.
- Future callers of `runWrangler` don't need to remember to quote.
- Net diff is actually 3 lines smaller than before this round, despite adding the full Windows code path.

### Other points

- **Comment in `run-wrangler.ts`**: rewrote to explain the cmd.exe approach with a link to the Node docs (since the previous "shell:true required on Windows" comment is no longer accurate).
- **Changeset**: kept @vicb's one-liner. The user-visible behavior remains "the warning goes away and injection is closed"; both are now true on every platform.
- **`--preview` / `--no-preview`**: applied at all three call sites in the previous round, kept.
- **`Closes #1182`**: still accurate.

### Verification (the honest caveat)

Verified locally on macOS: `pnpm code:checks` green, `pnpm -r test` 326/326 pass, end-to-end invocation against a real wrangler CLI works without DEP0190, security regression test confirms subshell injection is blocked, and a simulated `cmd.exe` command line construction produces correct escaping for paths-with-spaces and SQL-with-parens.

What I **cannot** verify locally: the Windows code path actually executing on real Windows. The repo's CI is ubuntu-only, and I don't have a Windows machine. The pattern matches cross-spawn's production implementation, the escaping is the codebase's existing well-tested function, and Node's docs explicitly recommend this approach, so I'm reasonably confident it works. But please validate on Windows before merging. Happy to follow any debugging if it turns out something needs tweaking.

kirkouimet:
Excellent catch from Devin Review.

### The bug

`quoteShellMeta` caret-escaped cmd.exe metacharacters AND THEN wrapped the result in `"..."`. cmd.exe treats `^` as literal inside double quotes, so the carets that were supposed to escape `(`, `)`, `;`, etc. became literal `^` characters in the argument value. The D1 `CREATE TABLE` SQL would have arrived at wrangler as `CREATE TABLE foo ^(x INT^)^;` rather than valid SQL.

### The fix

Pushed [229c45f](https://github.com/opennextjs/opennextjs-cloudflare/pull/1248/commits/229c45f) replacing the Windows branch with `cross-spawn`'s `escapeArgument` algorithm. The correct approach is:

1. Double backslash sequences that precede a quote (Windows argv parser quirk)
2. Wrap the value in `"..."`
3. Caret-escape every metacharacter, **including the wrapping quotes themselves**

The wrapping carets pair with the outer `cmd.exe /d /s /c "..."` parse: cmd.exe consumes them during command-line parsing, then re-parses the now-unquoted-but-still-quoted-internally string as a normal command, where the receiving program sees a normally-quoted argument. This is the production-tested algorithm used by `cross-spawn`, which is what npm/yarn/eslint/prettier all rely on for Windows spawn.

### Verification

- Output matches `cross-spawn`'s `escapeArgument` byte-for-byte across 8 representative inputs (SQL with parens, JSON with embedded quotes, paths with spaces, empty strings, etc.)
- Added `helpers.spec.ts` with 9 tests including an explicit regression test for the SQL case Devin flagged
- `pnpm code:checks` green, `pnpm -r test` 33 files / 335 tests pass (added 9 new)
- End-to-end POSIX test still works without DEP0190

Caveat from earlier still applies: the new Windows escaping is verified by simulation (output identical to cross-spawn's, which IS production-tested on Windows), but the actual `cmd.exe` execution path on real Windows hardware needs validation before merging.

packages/cloudflare/src/cli/commands/utils/helpers.ts
```diff
@@ -73,33 +73,53 @@ export async function getEnvFromPlatformProxy(options: GetPlatformProxyOptions,
 }
 
 /**
- * Escape shell metacharacters.
+ * Escape a single argument so it survives the `cmd.exe /d /s /c "<command line>"` invocation
```
vicb: The implementation seems to works for POSIX and windows, could you please update the comment to reflect that

packages/cloudflare/src/cli/commands/utils/helpers.ts
```diff
@@ -73,33 +73,53 @@ export async function getEnvFromPlatformProxy(options: GetPlatformProxyOptions,
...
  * @param arg
- * @returns escaped arg
+ * @returns escaped arg suitable for embedding in a `cmd /c "..."` command line
  */
 export function quoteShellMeta(arg: string) {
-	if (process.platform === "win32") {
-		if (arg.length === 0) {
-			return '""';
+	if (process.platform !== "win32") {
+		// POSIX fallback (preserved for any external callers): standard shell quoting.
```
vicb:
```suggestion
		// POSIX shell quoting.
```

packages/cloudflare/src/cli/commands/utils/run-wrangler.ts
```diff
@@ -96,39 +98,65 @@ export function runWrangler(
...
+				{
+					shell: false,
+					windowsVerbatimArguments: true,
+					stdio,
+					env,
+				}
+			)
+		: spawnSync(options.packager, packagerArgs, {
+				// Setting `shell` to `false` to avoid DEP0190.
+				// See https://nodejs.org/api/deprecations.html#dep0190-passing-args-to-nodechild-process-execfilespawn-with-shell-option
```
vicb: Move this up as it is used for both win and non win ands merge with l135/136

packages/cloudflare/src/cli/commands/utils/run-wrangler.ts
```diff
@@ -96,39 +98,65 @@ export function runWrangler(
...
+		...injectPassthroughFlagForArgs(options, [
+			...args,
+			...(wranglerOpts.environment ? ["--env", wranglerOpts.environment] : []),
+			...(wranglerOpts.configPath ? ["--config", wranglerOpts.configPath] : []),
+			...(wranglerOpts.target === "remote" ? ["--remote"] : []),
+			...(wranglerOpts.target === "local" ? ["--local"] : []),
+		]),
+	];
+
+	const stdio: ("inherit" | "ignore" | "pipe")[] =
```
vicb, line 118: Could you please re-add the former comment on this

vicb (changes requested):
The latest version looks much simpler hence better. Thanks for the updates.

We need to test on:
- WSL
- Windows

Before merging

kirkouimet:
Pushed [5a5a768](https://github.com/opennextjs/opennextjs-cloudflare/pull/1248/commits/5a5a768) addressing the four inline comments. All comment-only changes, no behavioral diff:

- **`helpers.ts:76`** — rewrote the `quoteShellMeta` docstring so the Windows and POSIX branches get equal weight rather than POSIX appearing as a footnote.
- **`helpers.ts:101`** — applied your suggestion verbatim (`// POSIX shell quoting.`).
- **`run-wrangler.ts:127`** — hoisted the `shell: false` / DEP0190 explanation up so it covers both branches, then prose describing the POSIX path and the Windows path in turn. Dropped the now-redundant inner comment.
- **`run-wrangler.ts:115`** — restored the stdio explanatory comment that was dropped when stdio got hoisted to a `const` in the earlier refactor.

`pnpm code:checks` green, all 335 tests pass.

marchequebec:
WSL2 + native Windows smoke test of the spawn logic at 5a5a768 (not the full `pnpm -r test` suite). We hit DEP0190 on every `opennextjs-cloudflare deploy` in Cloudflare Workers Builds (Linux, Node 24).

**WSL2 Ubuntu**, Node v22.20.0, `process.platform=linux`

- 1.20.6 pattern (`shell: true`, `` `--env ${payload}` `` with `staging$(echo INJECTED > file)`): file created (`INJECTED`).
- PR POSIX (`shell: false`, split `--env` / value): file absent; `npm exec wrangler --version` → `4.130.0`.

**Windows**, Node v24.19.0 (`C:\Program Files\nodejs\node.exe`), `process.platform=win32`

- `shell: true`: `wrangler 4.130.0` works; Node prints `DEP0190`.
- `shell: true` with `--env staging& echo INJECTED > file`: file created (`INJECTED`).
- PR `cmd.exe /d /s /c` + `quoteShellMeta`:
  - `node -e` works (`node-ok v24.19.0`); injection file absent.
  - `npm` / `npm.cmd exec wrangler --version` fails:

    ```
    Error: Cannot find module '...\npm\bin\npm-prefix.js'
    ```

  - `spawnSync("npm.cmd", …, { shell: false })` → `EINVAL` (expected on Windows).

POSIX path looks good for Linux CI. Windows `npm.cmd` prefix resolution breaks with the quoted `cmd.exe` line, so `opennextjs-cloudflare deploy` would fail there as written.

PR #1250, opened 2026-05-05
fix: define __name globally in the worker bundle

### LINKED ISSUE #1249: [Bug] ReferenceError: __name is not defined in inline script (Next 15)
## Summary

Every page in production logs `ReferenceError: __name is not defined`. The error fires from an inline `<script>` Next.js emits into the rendered HTML — line position points at `:10:11` and `:17:11` in the served document. Tracing the served HTML confirms `__name(k2,...)` is referenced but the helper is never defined globally.

## Versions

- `@opennextjs/cloudflare`: 1.19.4 (also reproduces on 1.19.6 — `dist/cli/build/bundle-server.js` is identical between versions)
- `next`: ^15.0.0
- `esbuild`: 0.25.4
- Node 20.x
- Deploy target: Cloudflare Workers (paid plan, `nodejs_compat`, `global_fetch_strictly_public`)

## Reproduction

1. Build any Next 15 app with `npx opennextjs-cloudflare build`.
2. Deploy: `npx wrangler deploy`.
3. Open any page in a browser, check the devtools console.
4. Observe: `ReferenceError: __name is not defined` at `<page-url>:10:11` (offset varies).
5. `curl <url> | grep __name` → returns one `__name(k2,...)` reference inside an inline `<script>` block.

## Root cause

`bundle-server.js` (`dist/cli/build/bundle-server.js` line ~62) calls esbuild's `build()` directly:

```js
const result = await build({
    entryPoints: [openNextServer],
    bundle: true,
    outfile: openNextServerBundle,
    format: "esm",
    target: "esnext",
    minifyWhitespace: projectOpts.minify && !debug,
    minifyIdentifiers: false,
    minifySyntax: projectOpts.minify && !debug,
    // ...
});
```

There's no `keepNames: true` and no `define: { __name: ... }`. When the Next 15 SSR runtime emits inline scripts that call esbuild's name-preservation helper `__name`, the helper is undefined at runtime in the worker bundle.

## Impact

- Non-fatal in flow (the framework catches it inside try/catch) but pollutes Sentry.
- Visible to anyone running devtools — investor / customer-facing demos look broken.
- Affects every page, every request.

## Suggested fix

Adding either of these to the esbuild options in `bundle-server.js` should resolve it:

**Option 1 — define an inline polyfill:**

```js
define: {
  __name: '((fn, name) => Object.defineProperty(fn, "name", { value: name, configurable: true }))',
},
```

**Option 2 — `keepNames: true`:**

esbuild's docs note that `keepNames: true` makes esbuild itself emit a polyfill helper whenever it inserts `__name` calls. Whether this also covers Next's already-emitted `__name` references is worth checking — option 1 is more defensive.

Either should be configurable via `defineCloudflareConfig()`, since users may want to opt out (e.g. to debug the runtime).

## Workaround for users (in the meantime)

A postbuild step can prepend the polyfill to `.open-next/server-functions/default/<app>/handler.mjs`:

```js
globalThis.__name = (fn, name) => Object.defineProperty(fn, 'name', { value: name, configurable: true });
```

Tracking this as maintenance debt until a proper upstream fix lands.

## Happy to PR

If the maintainers prefer Option 1 with a `defineCloudflareConfig` opt-out, I'm happy to send a PR.

richjhardy:
## Summary

Fixes #1249.

Next 15's emitted hydration script calls esbuild's name-preservation helper `__name`. The worker bundler in `bundle-server.ts` does not define this helper, so every page in production logs `ReferenceError: __name is not defined` to the browser console.

The error is non-fatal (the framework catches it inside try/catch) but it pollutes Sentry / devtools and is visible to anyone running devtools on a deployed app.

## Fix

Add an esbuild `define` block to the `build()` call:

```ts
define: {
  __name: '((fn, name) => Object.defineProperty(fn, "name", { value: name, configurable: true }))',
},
```

This inlines a tiny `__name` polyfill at every call site in the bundle, so the helper resolves at runtime regardless of whether the call comes from Next's emitted code or from esbuild's own transformations.

## Why `define` rather than `keepNames: true`

I considered passing `keepNames: true` to esbuild instead. The trade-off:

- `keepNames` makes esbuild emit its own runtime helper at the top of the bundle and inserts `__name` calls itself.
- But Next's already-emitted output contains `__name` calls that reference an unbound global. esbuild may not realise it needs to add the helper for those.

`define` is more defensive — it works for both Next's emitted calls and any esbuild-inserted ones.

## Test plan

- [x] Code compiles
- [x] Changeset added (`.changeset/define-name-helper.md`)
- [ ] Reviewer can verify no regression by running `pnpm --filter app-router preview` and checking the browser console — should be clean
- [ ] CI: lint, typecheck, unit tests

## Reproduction (for reviewers)

1. Build any Next 15 app with `npx opennextjs-cloudflare build`.
2. Deploy: `npx wrangler deploy`.
3. Open any page in a browser, open devtools console.
4. Without this PR: `ReferenceError: __name is not defined` appears.
5. With this PR: console is clean.

## Notes

I'm happy to iterate if the maintainers prefer `keepNames: true`, or if they'd like the choice to be configurable via `defineCloudflareConfig()`.

---

<a href="https://app.devin.ai/review/opennextjs/opennextjs-cloudflare/pull/1250" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/define-name-helper.md
+++ b/.changeset/define-name-helper.md
@@ -1,0 +1,11 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+fix: define `__name` globally in the worker bundle
+
+Next 15's emitted hydration script calls esbuild's name-preservation helper `__name`. Because the worker bundler did not define this helper, every page logged `ReferenceError: __name is not defined` to the console. The error is caught by the framework's try/catch and is non-fatal, but pollutes Sentry / devtools.
+
+Adds an esbuild `define` block to inline a tiny `__name` polyfill into the bundle, so the helper resolves at runtime regardless of whether the call comes from Next's emitted code or from esbuild's own transformations.
+
+Closes https://github.com/opennextjs/opennextjs-cloudflare/issues/1249
--- a/packages/cloudflare/src/cli/build/bundle-server.ts
+++ b/packages/cloudflare/src/cli/build/bundle-server.ts
@@ -78,6 +78,13 @@ export async function bundleServer(buildOpts: BuildOptions, projectOpts: Project
 		outfile: openNextServerBundle,
 		format: "esm",
 		target: "esnext",
+		// Define `__name` globally so calls emitted by the wrapped Next runtime
+		// (esbuild's keep-names helper) resolve at runtime in the worker.
+		// Without this, every page logs `ReferenceError: __name is not defined`
+		// from the Next 15 inline hydration script.
+		define: {
+			__name: '((fn, name) => Object.defineProperty(fn, "name", { value: name, configurable: true }))',
+		},
 		// Minify code as much as possible but stay safe by not renaming identifiers
 		minifyWhitespace: projectOpts.minify && !debug,
 		minifyIdentifiers: false,
--- a/packages/cloudflare/src/cli/build/bundle-server.ts
+++ b/packages/cloudflare/src/cli/build/bundle-server.ts
@@ -78,13 +78,6 @@ export async function bundleServer(buildOpts: BuildOptions, projectOpts: Project
 		outfile: openNextServerBundle,
 		format: "esm",
 		target: "esnext",
-		// Define `__name` globally so calls emitted by the wrapped Next runtime
-		// (esbuild's keep-names helper) resolve at runtime in the worker.
-		// Without this, every page logs `ReferenceError: __name is not defined`
-		// from the Next 15 inline hydration script.
-		define: {
-			__name: '((fn, name) => Object.defineProperty(fn, "name", { value: name, configurable: true }))',
-		},
 		// Minify code as much as possible but stay safe by not renaming identifiers
 		minifyWhitespace: projectOpts.minify && !debug,
 		minifyIdentifiers: false,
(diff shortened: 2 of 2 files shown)


richjhardy:
Good catch from the bot review — pushed a fix. The duplicate `define` key was silently being overwritten, which is exactly why my local test (via patch-package) was still showing the `__name` error after deploy.

Both `define` entries are now merged into the single existing block; the bot's diagnosis was spot-on.

richjhardy:
Update — the previous merge approach was wrong twice over. esbuild's `define` only accepts entity names or JS literals, not arrow function expressions. The build would have errored out with `Invalid define value (must be an entity name or JS literal)` even after the duplicate-key fix.

Switched to the existing `banner.js` so the polyfill is prepended to the bundle as ordinary source. Guarded with `globalThis.__name = globalThis.__name || ...` so it's a no-op if a future runtime supplies its own. This is verified working in my local equivalent patch.

Sorry for the noise — original analysis was based on the documented suggested fix in #1249, which I should have validated end-to-end before opening the PR.

richjhardy:
Closing — this PR is misdirected and I should have validated end-to-end before opening it. Apologies for the noise on the maintainers.

The `__name is not defined` error in #1249 fires from an inline `<script>` rendered into the HTML response (the next-themes theme-detection script), executed by the **browser**. Polyfilling `__name` in the **worker** bundle — which is what this PR does — has no effect on browser-side script execution.

The fix lives in the consuming app: a tiny inline `<script>` in the root layout's `<head>` that defines `globalThis.__name` ahead of any `<body>` content. I've applied that locally and verified it eliminates the error in the browser console.

If there's a related thing OpenNext could do — e.g. ensure third-party libraries' `<script>`-injection patterns don't emit `__name` calls without the helper — that would be a separate analysis and a separate PR. Not pretending to know whether that's even tractable.

Leaving #1249 open in case the maintainers want to keep tracking the broader pattern, or closing is fine too. Thanks for the patient bot reviews.

PR #1252, opened 2026-05-06
bump the number of retries for r2 cache uploads

vicb:
- Bump the max number of retries for R2 upload from 5 to 15
- Start at full speed - I couldn't reproduce any failure with the increase of the concurrency over time. It simplifies the code.

Note that the `BACKOFF_FACTOR` is now compute to respect a `MAX_RETRY_DELAY_MS`.

Also misc minor cleanup.

I think that https://github.com/opennextjs/opennextjs-cloudflare/issues/1173 and #1175 will no more be needed after this is merged.

Validated by uploading 100k assets to R2 (took about 20min from my local computer)

---

<a href="https://app.devin.ai/review/opennextjs/opennextjs-cloudflare/pull/1252" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/poor-peaches-jump.md
+++ b/.changeset/poor-peaches-jump.md
@@ -3,3 +3,5 @@
 ---
 
 bump the number of retries for r2 cache uploads
+
+Increase the retry count from 5 to 15 with a capped exponential backoff to improve resilience against transient R2 write failures during cache population.
--- a/.changeset/poor-peaches-jump.md
+++ b/.changeset/poor-peaches-jump.md
@@ -1,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+bump the number of retries for r2 cache uploads
--- a/packages/cloudflare/src/cli/commands/populate-cache.spec.ts
+++ b/packages/cloudflare/src/cli/commands/populate-cache.spec.ts
@@ -10,7 +10,13 @@ import { unstable_startWorker } from "wrangler";
 import { defineCloudflareConfig } from "../../api/config.js";
 import r2IncrementalCache from "../../api/overrides/incremental-cache/r2-incremental-cache.js";
 import { ensureR2Bucket } from "../utils/ensure-r2-bucket.js";
-import { getCacheAssets, populateCache, PopulateCacheOptions } from "./populate-cache.js";
+import {
+	getCacheAssets,
+	MAX_REQUEST_RETRIES,
+	MAX_RETRY_DELAY_MS,
+	populateCache,
+	PopulateCacheOptions,
+} from "./populate-cache.js";
 import { WorkerEnvVar } from "./utils/helpers.js";
 
 describe("getCacheAssets", () => {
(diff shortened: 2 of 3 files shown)


packages/cloudflare/src/cli/commands/populate-cache.spec.ts
```diff
@@ -398,61 +404,54 @@ describe("populateCache", () => {
...
+					)
+				);
 
-			expect(fetchMock).toHaveBeenCalledTimes(5);
-			expect(mockWorkerDispose).toHaveBeenCalled();
-		});
+				expect(fetchMock).toHaveBeenCalledTimes(MAX_REQUEST_RETRIES);
+				expect(mockWorkerDispose).toHaveBeenCalled();
+			},
+			/* timeout ms */ MAX_REQUEST_RETRIES * MAX_RETRY_DELAY_MS + 1_000
```
edmundhung: Why do we need to extend the test timeout time based on the max reties? We are using a fake timer here so it shouldn't wait for the actual delay.

edmundhung (approved):

packages/cloudflare/src/cli/commands/populate-cache.spec.ts
vicb: I guess it's because the `setTimeout` from `timers/promises`.
That's an unwanted behavior that was inherited from the current code.
You're welcome to take a look if you feel like.

_(It was not an issue before because the total test was less than the default timeout)_

packages/cloudflare/src/cli/commands/populate-cache.ts
```diff
@@ -56,6 +56,15 @@ import {
...
 } from "./utils/utils.js";
 
+// Maximum number of attempts to send the request
+export const MAX_REQUEST_RETRIES = 15;
+// Base delay for retries
+export const BASE_RETRY_DELAY_MS = 250;
+// Maximum delay for retries, used to calculate the backoff factor
+export const MAX_RETRY_DELAY_MS = 10_000;
+// Backoff factor for retries, calculated to ensure that the delay grows exponentially up to the maximum delay
+export const BACKOFF_FACTOR = (MAX_RETRY_DELAY_MS / BASE_RETRY_DELAY_MS) ** (1 / (MAX_REQUEST_RETRIES - 1));
```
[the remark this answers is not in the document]
vicb: It's ok to get an approximation here

conico974 (approved):

packages/cloudflare/src/cli/commands/populate-cache.spec.ts
vicb: OpenCode/Opus found the solution by mocking `node:timers/promises`

vicb:
Thanks for the review Edmund and Nico!

.changeset/poor-peaches-jump.md
```diff
@@ -0,0 +1,5 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+bump the number of retries for r2 cache uploads
```
vicb, line 5:
```suggestion
bump the number of retries for r2 cache uploads

Increase the retry count from 5 to 15 with a capped exponential backoff to improve resilience against transient R2 write failures during cache population.
```

Issue #1258, opened 2026-05-08
[BUG] Next 16.2.5 / 16.2.6 + 1.19.8 — every dynamic route 500s with `ComponentMod.handler is not a function` (and a leading-slash dynamic require of `/.next/server/middleware-manifest.json`)


johnoconnor0:
### Describe the bug

Building a Next.js project with `next@16.2.5` (and `16.2.6`) against `@opennextjs/cloudflare@1.19.8` produces a Worker that builds and uploads cleanly, but **every dynamic route returns HTTP 500** at runtime. Static assets (`/_next/...`, anything served via the `ASSETS` binding) serve fine — the failure is entirely in the request-handling pipeline.

`wrangler tail` surfaces two distinct failures depending on whether a workaround is applied:

1. **Without any patch** — `Error: Dynamic require of "/.next/server/middleware-manifest.json" is not supported`
2. **After patching `getMiddlewareManifest()` to short-circuit to `null`** (the project has no `middleware.ts`) — `TypeError: components.ComponentMod.handler is not a function`

Downgrading to `next@15.5.16` resolves both — every dynamic route handler returns the expected status with no patches.

The peerDep range in `@opennextjs/cloudflare@1.19.8` is `"next": ">=15.5.16 <16 || >=16.2.5"`, which advertises 16.2.5 as supported, so this is a regression against the documented matrix.

#### 1. `Dynamic require of "/.next/server/middleware-manifest.json"`

Same shape as #719 / #922, but reproducible on **pnpm 10.33** with no Yarn PnP. The leading-slash absolute path is the giveaway: `globalThis.__dirname ??= ""` in `cloudflare/init.js`, then Next-internal code does roughly `path.resolve(__dirname, ".next", ...)` → `"/.next/...`".

The bundle already contains a static stub:

```js
function loadManifest(path2, ...) {
  if (path2.endsWith("/server/middleware-manifest.json")) return { version: 3, middleware: {}, ... };
  ...
}
```

…but the failing call site bypasses that helper. It's a Next-internal method bundled verbatim:

```js
getMiddlewareManifest() {
  return this.minimalMode ? null : require(this.middlewareManifestPath);
}
```

esbuild can't statically resolve `this.middlewareManifestPath` so the dynamic-`require` shim throws.

#### 2. `TypeError: components.ComponentMod.handler is not a function`

[...]

PR #1242, opened 2026-04-30
bump wrangler & compat dates

vicb:
Also add a warning when the compat date is older than 6 months

image

---

<a href="https://app.devin.ai/review/opennextjs/opennextjs-cloudflare/pull/1242" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

PR #1262, opened 2026-05-11
Add Next.js 16.2.x compatibility patches (perf_hooks shim, fast-set-immediate shim, Turbopack handler delegation)

chanceaclark:
## What

Three build-time patches that make Next.js 16.2.x apps work on Cloudflare Workers. Without these, every app-router request crashes on cold start.

### 1. `perf_hooks` shim (`f61f7c4`)

Next 16.2.x bundles `gc-observer.js` (imports `PerformanceObserver` from `perf_hooks`) and `@vercel/otel` (uses `require("perf_hooks").performance`) into the instrumentation hook loading path. esbuild normalizes bare `perf_hooks` to `node:perf_hooks` on the node platform, and Workers does not expose this module even with `nodejs_compat`.

A new shim re-exports `globalThis.performance` and provides a no-op `PerformanceObserver`. Resolved via a new `shimNodeModules` `onResolve` plugin (not `alias`) so the shim is bundled instead of being externalized by `platform: "node"`.

### 2. `fast-set-immediate` shim (`b42ef6d`)

Next 16.2's app-render-scheduling imports `next/dist/server/node-environment-extensions/fast-set-immediate.external`, whose top-level `install()` runs:

```js
const nodeTimers = require('node:timers');
globalThis.setImmediate = nodeTimers.setImmediate = patchedSetImmediate;
```

Under `nodejs_compat` the `node:timers` namespace is a frozen ESM module, so `nodeTimers.setImmediate = ...` throws ``Cannot assign to read only property 'setImmediate' of object '[object Module]'`` on every request, before any user code runs.

The fast-immediate scheduler is an optimization, not a correctness requirement. A new shim exports the same names (`DANGEROUSLY_runPendingImmediatesAfterCurrentTask`, `expectNoPendingImmediates`, `unpatchedSetImmediate`) as no-ops and surfaces `globalThis.setImmediate` directly.

This commit also extends the bundle banner to expose `AsyncLocalStorage` on `globalThis` so Next's `async-local-storage.js` picks up the real implementation rather than falling back to `FakeAsyncLocalStorage` (which throws on `.run()`/`.exit()` and cascades into a confusing ``Cannot read properties of undefined (reading 'require')`` when Turbopack loads `app-page-turbo.runtime.prod.js`).

### 3. `patchPageExports` plugin (`39ab69b`)

Next 16.2.x Turbopack app-page templates set up exports on a *child* module via `esmExport(bindings, childId)`, then re-call `esmExport` for the entry module's own id. Each page chunk then does:

```js
R.m(<entryId>)
module.exports = R.m(<entryId>).exports
```

`R.m(<entryId>).exports` only carries the second call's vendored bindings — `handler` lives on the *child* module. At request time Next reads `components.ComponentMod.handler`, gets `undefined`, and crashes with ``ComponentMod.handler is not a function`` on every app route.

The fix:

1. Scan SSR `app-page` template chunks under `.next/server/chunks/ssr/` to build an entry→child module-id map by anchoring on `module.exports = [<entryId>` and `a.s([…,"handler",…], <childId>)`.
2. Add an esbuild `onLoad` plugin that matches `.next/server/app/**/page.js` and `.next/server/app/**/route.js`. When the entry id has a recorded child, replace the terminator with a merge expression that copies the entry's own-property descriptors first, then layers the child's descriptors for keys the entry doesn't define. The merged object exposes `handler` alongside the Turbopack bindings while preserving lazy getters.

The patch only touches `page.js` / `route.js`. It does not modify the Turbopack runtime, `esmExport`, or `instantiateModule`, so it cannot regress instrumentation hook loading or app-page-turbo external loading.

## Why

Without these three patches, Next.js 16.2.x on Cloudflare Workers fails on every request:

- Cold start crashes at `gc-observer.js` evaluation (perf_hooks)
- First request crashes at fast-set-immediate's `install()` (node:timers mutation)
- Every app-router request crashes at `ComponentMod.handler is not a function`

Together they make a real Next 16.2.6 app-router storefront return 200s across homepage, search, category, product, brand, and blog routes.

## Testing

Verified locally with a Next.js 16.2.6 app-router site running under `wrangler@4.90.0 dev`. Without the patches: 500 on every request. With the patches: 200 across the homepage, search, compare, category, product, brand, and blog routes.

`tsc --noEmit` and `eslint` are clean.

## Notes

- Related to #1258.
- One changeset (`patch`) covers all three fixes.
- Happy to split into three separate PRs if that's preferable for review.

### DIFF
--- a/.changeset/next-16-2-compat.md
+++ b/.changeset/next-16-2-compat.md
@@ -1,0 +1,11 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+Add Next.js 16.2.x compatibility patches
+
+- **`perf_hooks` shim**: Next 16.2.x's `gc-observer` and `@vercel/otel` import `PerformanceObserver` / `performance` from `node:perf_hooks`, which Cloudflare Workers does not expose. A new shim re-exports `globalThis.performance` and provides a no-op `PerformanceObserver`, resolved via a new `shimNodeModules` onResolve plugin.
+- **`fast-set-immediate` shim**: Next 16.2's app-render scheduler imports `next/dist/server/node-environment-extensions/fast-set-immediate.external`, whose `install()` mutates the frozen `node:timers` module on load and throws "Cannot assign to read only property 'setImmediate'". A shim replaces the module with no-op control functions; the scheduler optimization is not needed in Workers. The bundle banner also exposes `AsyncLocalStorage` on `globalThis` so Next's check picks up the real implementation rather than the throwing fallback.
+- **`patchPageExports` plugin**: Next 16.2.x Turbopack templates set `handler` on a child module via `esmExport(bindings, childId)` then re-call `esmExport` for the entry id. Each page chunk's `module.exports = R.m(<entryId>).exports` then lacks `handler`, causing "ComponentMod.handler is not a function" on every app route. A new build-time plugin scans the SSR template chunks to discover entry→child mappings and rewrites each page.js / route.js terminator to merge the child's own-property descriptors over the entry's, exposing `handler` without touching the Turbopack runtime.
+
+Related: #1258
--- a/packages/cloudflare/src/cli/build/bundle-server.ts
+++ b/packages/cloudflare/src/cli/build/bundle-server.ts
@@ -22,6 +22,7 @@ import { patchResolveCache, patchSetWorkingDirectory } from "./patches/plugins/o
 import { handleOptionalDependencies } from "./patches/plugins/optional-deps.js";
 import { patchPagesRouterContext } from "./patches/plugins/pages-router-context.js";
 import { patchDepdDeprecations } from "./patches/plugins/patch-depd-deprecations.js";
+import { patchPageExports } from "./patches/plugins/patch-page-exports.js";
 import { fixRequire } from "./patches/plugins/require.js";
 import { shimRequireHook } from "./patches/plugins/require-hook.js";
 import { patchRouteModules } from "./patches/plugins/route-module.js";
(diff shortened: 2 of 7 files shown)


vicb:
Thanks for the PR @chanceaclark 

> * Happy to split into three separate PRs if that's preferable for review.

That would really be helpful please - please include a repro of the failure (link to a GH repo with instruction) for each of them.

Some comments/thoughts about your points above:

1) `perf_hooks`

> esbuild normalizes bare perf_hooks to node:perf_hooks on the node platform, and Workers does not expose this module even with nodejs_compat

`perf_hooks` is implemented natively in workerd if the compatibility date is on or after [2026-03-17](https://github.com/cloudflare/workerd/blob/ac31559e8bec3787e72ae5ab1276ccc7e12899e7/src/workerd/io/compatibility-date.capnp#L1064)

2) Cannot assign to read only property 'setImmediate' of object '[object Module]'

There used to be [a bug in workerd](https://github.com/cloudflare/workerd/issues/5844). It has been fixed months ago. Here again a recent compatibility date might be enough to fix the issue.

So before splitting this PR into multiple issues, it would be nice to try and bump the compatibility_date to something recent i.e. 2026-04-15 and check if it is enough to fix at least 1 and 2.

We have recently added [a build time warning for older compat date](https://github.com/opennextjs/opennextjs-cloudflare/pull/1242). We should probably turn this into a build error.

This might also explain why I was not able to repro #1258.

chanceaclark:
1. and 2. were resolved by bumping the compatibility date like you said 👨‍🍳 💋

I could only reproduce 3. on older versions of `wrangler`, I assume maybe it's because of previous module resolution bundling happening 🤷‍♂️ Didn't dig too deep, but upgrading `wrangler` to the latest solved it.

Closing this PR.

PR #1270, opened 2026-05-18
fix: skip non-upload-triggered worker versions in skew-protection

### LINKED ISSUE #1230: Skew protection can map previous deployments to asset-less worker versions
## Skew protection can map previous deployments to asset-less worker versions

When a deploy pipeline creates new worker versions after `wrangler deploy` (e.g. Terraform's `cloudflare_worker_secret` resources), those versions become the newest in `listWorkerVersions`. `updateDeploymentMapping` then picks them to replace the previous `"current"` sentinel, which means old `?dpl=...` requests get redirected to a preview URL that doesn't serve static assets.

### Reproduction

Terraform config that creates `cloudflare_worker_secret` resources with `depends_on = [null_resource.opennext_deploy]`. After an initial deploy we get:

```
V1 wrangler upload     <- has assets
V2 terraform secret    <- no assets
V3 terraform secret
V4 terraform secret
V5 terraform secret
```

On the next deploy, `getDeploymentMapping` captures `V5.id` for the previous deploy. Requests with the previous deploy's `dpl` hit the skew protection redirect:

```
$ curl -sI "https://<V5-id>-<worker>.<preview-domain>.workers.dev/_next/static/css/<hash>.css"
HTTP/2 404
x-preview-user-error: true
```

The wrangler-triggered versions in the same cluster (V1 here) serve the asset fine (200). The redirect just lands on the wrong one.

### Root cause

In [`skew-protection.ts#L148`](https://github.com/opennextjs/opennextjs-cloudflare/blob/main/packages/cloudflare/src/cli/commands/skew-protection.ts#L148), `versions[0]!.id` replaces `"current"` with whichever worker version was created most recently, regardless of what triggered it. Secret API calls, binding changes, and similar metadata-only operations all create new versions without re-uploading the asset bundle.

### Proposed fix

Filter `listWorkerVersions` (or the replacement in `updateDeploymentMapping`) to only consider versions that carry the asset bundle. The `annotations.workers/triggered_by` field on each version distinguishes them:

- `upload` / `version_upload` -> from `wrangler deploy` / `wrangler versions upload`, has assets
- `secret` -> from secret API updates, no assets
- other metadata-only triggers likewise skip the asset bundle

Either drop non-upload triggers explicitly, or keep an allowlist of upload-type triggers. I'd lean allowlist since it fails safe against future trigger types.

Happy to open a PR if the approach makes sense.

alex-all3dp:
Worker versions created by metadata-only operations (e.g. Cloudflare API secret updates with `triggered_by=secret`) do not include the static assets bundle. Previously, such a version could be picked as the "latest" target when replacing the "current" sentinel in the deployment mapping, causing `/_next/static/*` requests to return 404 on past deployments.

`listWorkerVersions` now skips versions whose metadata.annotations["workers/triggered_by"] is not in `{upload, version_upload}`. Versions with no annotation are kept for backward compatibility.

Closes #1230

---

<a href="https://app.devin.ai/review/opennextjs/opennextjs-cloudflare/pull/1270" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/cloudflare/src/cli/commands/skew-protection.ts
+++ b/packages/cloudflare/src/cli/commands/skew-protection.ts
@@ -40,6 +40,9 @@ const MAX_NUMBER_OF_VERSIONS = 20;
 const MAX_VERSION_AGE_DAYS = 7;
 const MS_PER_DAY = 24 * 3600 * 1000;
 
+/** Worker-version trigger types that produce a full upload (assets + code). */
+const UPLOAD_TRIGGER_TYPES = new Set(["upload", "version_upload"]);
+
 /**
  * Compute the deployment mapping for a deployment.
  *
@@ -244,11 +247,18 @@ export async function listWorkerVersions(
 		})) {
 			const id = version.id;
 			const createdOn = version.metadata?.created_on;
+			const triggeredBy = (version.metadata as any)?.annotations?.["workers/triggered_by"];
 
 			if (id && createdOn) {
 				const createdOnMs = new Date(createdOn).getTime();
 				if (createdOnMs < afterTimeMs) {
 					break;
+				}
+				// Skip metadata-only versions (e.g. secret/service_token triggers)
+				// that lack the static assets bundle. Versions with no annotation
+				// are kept for backward compatibility.
+				if (triggeredBy !== undefined && !UPLOAD_TRIGGER_TYPES.has(triggeredBy)) {
+					continue;
 				}
 				versions.push({ id, createdOnMs });
 				if (versions.length >= maxNumberOfVersions) {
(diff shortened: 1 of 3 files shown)


packages/cloudflare/src/cli/commands/skew-protection.ts
```diff
@@ -40,6 +40,9 @@ const MAX_NUMBER_OF_VERSIONS = 20;
 const MAX_VERSION_AGE_DAYS = 7;
 const MS_PER_DAY = 24 * 3600 * 1000;
 
+/** Worker-version trigger types that produce a full upload (assets + code). */
+const UPLOAD_TRIGGER_TYPES = new Set(["upload", "version_upload"]);
```
james-elicx, line 44: I wonder if `rollback` and `promotion` are valid options here... https://github.com/cloudflare/workers-sdk/blob/b27eb18de664e416316d50116e568513d08123eb/packages/wrangler/src/versions/list.ts#L106-L119

james-elicx (approved):

PR #1272, opened 2026-05-19
fix(queue): treat 200 + non-REVALIDATED response as success

### LINKED ISSUE #662: [BUG]
### Describe the bug

When deploying a Next.js App Router app using open-next to Cloudflare Workers with R2 object storage as the cache backend, ISR pages using Partial Prerendering (◐) work initially but break after the revalidation period : 

```tsx
// app/challenge/[slug].tsx

export const revalidate = 60;

export async function generateStaticParams() {
  return posts.map((post) => ({ slug: post.slug }));
}

export default async function Page({ params }: { params: { slug: string } }) {
  const awaitedParams = await params;
  const { slug } = awaitedParams as { slug: string }; 
  const post = getPost(slug);

  if (!post) {
    return <h1>404</h1>;
  }

  return (
    <div className="p-8 space-y-4">
      <h1 className="text-3xl font-bold">{post.title}</h1>
      <p>{post.content}</p>

      <Suspense fallback={<p>Loading dynamic content...</p>}>
        <DynamicContent />
      </Suspense>
    </div>
  );
}

```

The build correctly shows ◐ and the page works at first.

However, after sometimes (like 24h ; revalidation period is 1h) , the route fails at runtime with:

```bash
FatalError: The revalidation for /challenge/[slug] cannot be done. This error should never happen.
```

After this happens:
- The Suspense fallback spins indefinitely
- Refreshing the page leads to a white screen
- No regeneration ever occurs again

### Steps to reproduce

1. Use App Router in Next.js (v14) with open-next
2. Create a route with:
    - Static content loaded 
    - A <Suspense> section 

3. Deploy with `opennextjs-cloudflare build && opennextjs-cloudflare deploy` to Cloudflare Workers

4. Visit the page → it works

5. Wait many hours

6. Refresh the page → it fails to revalidate and breaks with the error shown above

### Expected behavior

I expect that the route should be revalidated and regenerated after the revalidate interval

1.0.2

### Wrangler version

4.14.4

### next info output

```shell
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.4.0: Fri Apr 11 18:33:46 PDT 2025; root:xnu-11417.101.15~117/RELEASE_ARM64_T8112
  Available memory (MB): 16384
  Available CPU cores: 8
Binaries:
  Node: 20.15.1
  npm: 10.7.0
  Yarn: 1.22.22
  pnpm: 10.10.0
Relevant Packages:
  next: 15.4.0-canary.30 // There is a newer canary version (15.4.0-canary.31) available, please upgrade! 
  eslint-config-next: 15.4.0-canary.30
  react: 19.1.0
  react-dom: 19.1.0
  typescript: 5.2.2
```

### Additional context

_No response_

alex-all3dp:
## Problem

`DOQueueHandler.executeRevalidation` throws `FatalError("…cannot be done. This error should never happen.")` whenever the HEAD self-fetch returns `200` with an `x-nextjs-cache` value other than `REVALIDATED`. In real-world deployments serving meaningful traffic this happens routinely: when a user hits a stale page, the worker's in-isolate stale-while-revalidate regenerates the page in parallel with the queued revalidation. By the time the DO's HEAD self-fetch reaches the ISR handler, the page is already fresh in the incremental cache and Next.js returns `HIT` (or `STALE`) rather than `REVALIDATED`.

This produces noisy `FatalError` logs on healthy, correctly-serving pages. The desired outcome — fresh content in cache — has been achieved; the request just took a different path.

Reproduces under load with any throttling that delays the queued HEAD relative to the in-isolate SWR. Reported in [#662](https://github.com/opennextjs/opennextjs-cloudflare/issues/662); seen in production on opennextjs-cloudflare 1.19.9, Next.js 16.2.6 across deeply nested ISR routes (`dynamic = 'error'`, `revalidate` in the 1-week–1-year range).

## Fix

In `packages/cloudflare/src/api/durable-objects/queue.ts`, fall through to the existing success path when `status === 200` regardless of the `x-nextjs-cache` value:

- The sync table is updated (`INSERT OR REPLACE INTO sync …`).
- The route is removed from `routeInFailedState`.
- A `debug` log records the non-REVALIDATED status for observability.

The `FatalError` import is dropped (no other usage).

## Test changes

The existing test `"should not put it in failed state for an incorrect 200"` (under `failed revalidation`) is replaced by `"should treat 200 + non-REVALIDATED response as success (page already regenerated by another path)"` under `successful revalidation`, which now also asserts the sync-table write.

All 22 tests in `queue.spec.ts` pass; full suite (328 tests) passes; `pnpm code:checks` clean.

## Risk

Low. Behavior changes only for the previously-fatal branch. 200+REVALIDATED, 404, 500, and other non-200 responses are unchanged.

Closes #662

---

<a href="https://app.devin.ai/review/opennextjs/opennextjs-cloudflare/pull/1272" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


packages/cloudflare/src/api/durable-objects/queue.ts
```diff
@@ -132,11 +127,14 @@ export class DOQueueHandler extends DurableObject<CloudflareEnv> {
...
 			});
 			// Now we need to handle errors from the fetch
-			if (response.status === 200 && response.headers.get("x-nextjs-cache") !== "REVALIDATED") {
-				this.routeInFailedState.delete(msg.MessageDeduplicationId);
-				throw new FatalError(
-					`The revalidation for ${host}${url} cannot be done. This error should never happen.`
-				);
+			if (response.status === 200) {
+				const cacheStatus = response.headers.get("x-nextjs-cache");
+				if (cacheStatus !== "REVALIDATED") {
```
conico974, line 132: We shouldn't treat STALE as a success that's for sure. 
But even HIT should never happen (maybe in PPR, haven't tested that), we are supposed to remove the in isolate revalidation, if not it means there is a bug that we need to fix

conico974 (commented):
@alex-all3dp Is it for PPR only ?

packages/cloudflare/src/api/durable-objects/queue.ts
alex-all3dp: Thanks for the review @conico974 

Not PPR-only, we hit it on regular ISR routes (`dynamic = 'error'`  `revalidate` from 1 week to 1 year), no `cacheComponents`/PPR enabled. Our `open-next.config.ts` is

```javascript
const openNextConfig: OpenNextConfig = {
  ...defineCloudflareConfig({
    incrementalCache: withRegionalCache(r2IncrementalCache, {
      mode: 'long-lived',
    }),
    queue: queueCache(doQueue, {regionalCacheTtlSec: 30, waitForQueueAck: false}),
    tagCache: doShardedTagCache({
      baseShardSize: 4,
      regionalCache: true,
    }),
    cachePurge: purgeCache({type: 'durableObject'}),
  }),
  dangerous: {
    middlewareHeadersOverrideNextConfigHeaders: true,
  },
  cloudflare: {
    skewProtection: {
      enabled: true,
      maxNumberOfVersions: 10,
      maxVersionAgeDays: 20
    },
  },
}
```

You're right on both points though. If in-isolate SWR-triggered regen shouldn't exist alongside the queue's HEAD regen, then this PR would just paper over an underlying bug it seems. 

How do ypu prefer to handle this PR? Closing it would be reasonable. Alternatively narrow it to handle HIT only, drop STALE, add a guard rail ((e.g. only treat HIT as success when the cache freshness is verifiable) and mark the change as a defense for potential race conditions?

PR #1289, opened 2026-06-23
fix(cache-purge): spread SQLite bindings in BucketCachePurge alarm so tag purges run

### LINKED ISSUE #1288: fix(cache-purge): spread SQLite bindings in BucketCachePurge alarm (DELETE crashes with "Wrong number of parameter bindings")
Bug
BucketCachePurge never purges: its alarm() throws Wrong number of parameter bindings for SQL query on every flush, so on-demand revalidateTag purges silently never reach the Cloudflare cache. With bypassTagCacheOnCacheHit: true, pages then serve stale until ISR TTL.

Root cause
Cloudflare's SqlStorage.exec(query, ...bindings) is variadic — it counts the binding arguments. The DO passes bindings as a single array instead of spreading:

this.ctx.storage.sql.exec(
  `DELETE FROM cache_purge WHERE tag IN (${tags.map(() => "?").join(",")})`,
  tags.map((row) => row.tag),   // ← 1 array arg for N placeholders
);
With N placeholders but 1 argument, exec reports "Wrong number of parameter bindings". The INSERT survives only because it has a single placeholder (1 placeholder, 1 arg).

Fix
Spread the bindings:

// INSERT
this.ctx.storage.sql.exec(`INSERT OR REPLACE INTO cache_purge (tag) VALUES (?)`, tag);
// DELETE
this.ctx.storage.sql.exec(
  `DELETE FROM cache_purge WHERE tag IN (${tags.map(() => "?").join(",")})`,
  ...tags.map((row) => row.tag),
);
Also fixes the latent infinite-loop guard do { ... } while (tags.length >= 0) → while (tags.length > 0) (currently only saved by the if (tags.length === 0) return at the top of the loop).

Repro
Trigger revalidateTag for ≥2 tags (so the alarm's DELETE ... IN (?,?) has >1 placeholder) → the DO alarm throws and the purge never runs.

Affected: @opennextjs/cloudflare@1.19.11.

thatssoheil:
## What & why

`BucketCachePurge.alarm()` passed its tag bindings to `SqlStorage.exec` as a **single array**. `exec(query, ...bindings)` is variadic over its bindings, so a multi-tag `DELETE ... WHERE tag IN (?, ?, …)` received 1 binding for N placeholders and threw **"Wrong number of parameter bindings"** on every flush.

The consequence is silent: on-demand `revalidateTag` purges never reached the Cloudflare cache, and with `bypassTagCacheOnCacheHit` enabled, pages served stale until the ISR TTL expired. The `INSERT` survived only because it has a single placeholder (1 binding, 1 placeholder).

## Changes

- **Spread the `DELETE` bindings** (`...tags.map((row) => row.tag)`) — the actual crash.
- **Normalise the `INSERT`** binding to the same variadic form (`tag` instead of `[tag]`).
- **Tighten the drain-loop guard** `while (tags.length >= 0)` → `while (tags.length > 0)`. The `>= 0` form never exits via the condition; the loop only terminated via the early `return` at the top, leaving a latent infinite-loop guard.
- **Spec:** the existing `alarm` assertions encoded the buggy array form (`["tag1", "tag2"]`); corrected them to the spread form and added a regression guard (`.not.toHaveBeenCalledWith([...])`) referencing this bug.
- **Changeset:** `patch`.

## Repro

Trigger `revalidateTag` for ≥2 tags so the alarm's `DELETE ... IN (?, ?)` has >1 placeholder → the DO `alarm()` throws and the purge never runs. Affects `@opennextjs/cloudflare@1.19.11`.

Fixes #1288

---
<sub>Implemented with AI assistance (Claude), reviewed and verified by me before submitting.</sub>

### DIFF
--- a/.changeset/fix-bucket-cache-purge-bindings.md
+++ b/.changeset/fix-bucket-cache-purge-bindings.md
@@ -1,0 +1,17 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+fix: spread SQLite bindings in BucketCachePurge alarm so tag purges run
+
+`BucketCachePurge.alarm()` passed its tag bindings to `SqlStorage.exec` as a
+single array. `exec(query, ...bindings)` is variadic over its bindings, so for a
+multi-tag `DELETE ... WHERE tag IN (?, ?, …)` the binding count (1) disagreed
+with the placeholder count (N) and `exec` threw "Wrong number of parameter
+bindings" on every flush. On-demand `revalidateTag` purges therefore never
+reached the Cloudflare cache, and with `bypassTagCacheOnCacheHit` enabled pages
+served stale until the ISR TTL expired.
+
+Spread the bindings, normalise the `INSERT` to the same variadic form, and
+tighten the drain loop's guard from `while (tags.length >= 0)` (which never
+exits via the condition) to `while (tags.length > 0)`.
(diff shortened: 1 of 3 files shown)


conico974 (approved):
Thanks for the PR

Issue #1225, opened 2026-04-22
[BUG] cacheComponents: true permanent "Connection closed" error with blank static shell served


mahalrs:
### Describe the bug

When `cacheComponents: true` is enabled in next.config.ts (Next.js 16.2.4), the OpenNext Cloudflare only serves static shell (static pages and Suspense fallbacks) without streaming the dynamic content inside suspense. This results in all pages being permanently broken and the browser throws Error: Connection closed.

Related to https://github.com/opennextjs/opennextjs-cloudflare/issues/1115

### Steps to reproduce

1. Create a Next.js 16.2.4 app with cacheComponents: true
2. Have an async Server Component wrapped in `<Suspense>`:
```jsx
async function StreamedContent() {
  const user = await getUser() // async operation
  return <div>user.name</div>
}

export default function Page() {
  return (
    <div>
      <div>
          Some content...
      </div>
      <Suspense fallback={<div>Loading...</div>}>
          <StreamedContent />
      </Suspense>
    </div>
  )
}
```
3. Build and deploy with @opennextjs/cloudflare
4. Page shows blank static shell with "Some content..." and "Loading..." but the actual `<StreamedContent/>` never renders in the browser. Browser throws Error: Connection closed.

### Expected behavior

1. The cached static shell should be sent as the first chunk of a streaming response, rendering "Some content..." and "Loading...".
2. The Worker should then initiate dynamic rendering for the Suspense boundaries and render the `<StreamedContent/>` showing user name.
3. The browser receives the Suspense fallback HTML but never gets resolution data
4. The browser throws `Error: Connection closed` when the response stream ends
5. The page remains stuck on the loading fallback permanently

### @opennextjs/cloudflare version

1.19.1

### Wrangler version

4.83.0

### next info output

```shell
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.1.0: Mon Oct 20 19:32:41 PDT 2025; root:xnu-12377.41.6~2/RELEASE_ARM64_T6000
  Available memory (MB): 32768
  Available CPU cores: 10
Binaries:
  Node: 24.9.0
  npm: 11.6.0
  Yarn: 1.22.19
  pnpm: 10.17.0
Relevant Packages:
  next: 16.2.4 // Latest available version is detected (16.2.4).
  eslint-config-next: N/A
  react: 19.2.5
  react-dom: 19.2.5
  typescript: 5.9.3
Next.js Config:
[...]

PR #1293, opened 2026-06-25
fix: remove type: "bytes" from ReadableStream for flight data

e-chan1007:
# Summary

This PR fixes the "Connection closed" error caused by corrupted RSC flight data streams on Cloudflare Workers.

Next.js generates inline `<script>` tags for embedding flight data in the [`createInlinedDataReadableStream`](https://github.com/vercel/next.js/blob/d4b234030d24b8b4f4c29aa3aac27962d42739bc/packages/next/src/server/app-render/use-flight-response.tsx#L165-L178) function, which reconstructs a `ReadableStream` using the `type: "bytes"` (BYOB) option.

However, `workerd` splits the `ReadableStream` in BYOB mode into 4KB chunks and this can occasionally cause the data chunks to be read in the wrong order. When using nested or parallel `<Suspense>`, this behavior results in the flight data generated by multiple `<Suspense>` components getting mixed up. This corruption of the `<script>` tag payload triggers the "Connection closed" hydration error (as detailed in https://github.com/opennextjs/opennextjs-cloudflare/issues/1225#issuecomment-4335815843).

To resolve this issue, I implemented the patch for the `createInlinedDataReadableStream` function to remove the `type: "bytes"` option, using the default stream mode without chunk splitting. While the underlying bug seems to be within `workerd` itself, applying this patch serves as immediate workaround rather than waiting for an upstream fix in the runtime.

# Validation

- [X] `pnpm code:checks` passed
- [X] `pnpm test` passed

---

<a href="https://app.devin.ai/review/opennextjs/opennextjs-cloudflare/pull/1293" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/shy-laws-wonder.md
+++ b/.changeset/shy-laws-wonder.md
@@ -1,0 +1,10 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+fix: patch the createInlinedDataReadableStream function not to use the type: "bytes" option in ReadableStream constructor
+
+workerd seems not to correctly support the type: "bytes" option in ReadableStream constructor, that breaks inlined flight data streaming over 4kB.
+By removing the type: "bytes" option, the inlined flight data will not be split into multiple chunks, and the streaming will work correctly.
+
+(fixes #1225)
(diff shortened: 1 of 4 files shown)


xDragonZ:
any updates on this? tested the patch is working

Issue #617, opened 2025-04-30
[FEATURE]  Add support for node middleware


conico974:
Ideally we'd like to support the Node middleware.
Right now, it causes issues with some dynamic require of the manifest.

andrew-tcn:
+1

bookernath:
Node.js middleware went stable with Next 15.5:

https://nextjs.org/blog/next-15-5#nodejs-middleware-stable

evanlong-me:
## Unblock path that can actually ship to npm

npm still publishes `@opennextjs/cloudflare` from **this** repo (`1.20.2`, 2026-07-21). adapters-api#38 is promising but stacked/dirty and does not publish the live package yet — so it cannot unblock users today.

Merge-ready PR on current main with local build+runtime verification:

**https://github.com/opennextjs/opennextjs-cloudflare/pull/1320**
(rebase of @ScienHAC's #1309)

Verified on `examples/e2e/experimental` `proxy.ts` (`node:crypto.randomUUID()`):

- build no longer throws `Node.js middleware is not currently supported`
- `/` sets `x-middleware-test` + `x-random-node`
- `/redirect`, `/api/hello`, `/rewrite` behave correctly under `opennextjs-cloudflare preview`

@conico974 @james-elicx @vicb — if the standing decision is still “never in this repo, only adapters-api”, please say so explicitly with a cutover/ship timeline. Otherwise #1320 is ready for review and is the fastest way to stop forcing the `middleware.ts` deprecation workaround on Next 16.

PR #1309, opened 2026-07-13
feat: support Node.js middleware (proxy.ts)

### LINKED ISSUE #1277: [BUG] No support for `proxy.js` in Cloudflare Workers

ScienHAC:
Closes #1277
Related: #617

Added some changes from @evanlong-me in #1357

## What

Adds support for Node.js middleware, which is what Next.js 16 `proxy.ts` compiles to. Previously the build refused to continue with `Node.js middleware is not currently supported. Consider switching to Edge Middleware.`.

## Why the previous attempt (#1280) was wrong

I opened #1280 and it was rightly closed: it only removed the guard, so a `proxy.ts` app would build and then crash at runtime. This PR implements the missing capability instead, and the guard now only relaxes because the code behind it exists.

## Root cause

`@opennextjs/aws` builds external Node middleware for a Node server:

- `core/nodeMiddlewareHandler.ts` does `await import("./.next/server/middleware.js")` at request time
- esbuild marks `./.next/*` as external
- the OpenNext config manifests are read from disk on startup

None of that works in workerd: there is no filesystem and no runtime module loading, so the emitted `middleware/handler.mjs` cannot run on Cloudflare.

## Approach

After `createMiddleware()` produces the aws output, re-bundle `middleware/handler.mjs` on the Cloudflare side so it is fully self-contained (`packages/cloudflare/src/cli/build/open-next/bundle-node-middleware.ts`):

- keeps OpenNext's routing layer in charge — the entry point is still `adapters/middleware.js` + `core/nodeMiddlewareHandler.js`, so matchers, cookies, rewrites, redirects and `NextResponse.next()` are handled by OpenNext, not reimplemented here
- statically bundles the compiled middleware from the `copyTracedFiles` output via an esbuild plugin, replacing the dynamic `import()`
- inlines the config manifests with the existing `openNextEdgePlugins`
- defines `process.env.NEXT_RUNTIME = "edge"` so `setup-node-env.external.js` (which patches read-only workerd globals) is skipped
- aliases `@opentelemetry/api` to `next/dist/compiled/@opentelemetry/api`, because Next's tracer `require`s it without a try/catch on that branch
- inlines chunks for both bundlers: `patchWebpackRuntime` for webpack, and the existing Turbopack runtime patch (now reused as `patchTurbopackRuntimeCode`) for Turbopack, which is the default in Next 16 `create-next-app`
- reuses `setWranglerExternal()` so `.wasm`/`.bin` assets are bundled by wrangler rather than esbuild

No existing behaviour changes when there is no Node middleware.

## Safety

The build still fails loudly rather than deferring a crash to production: it throws if the compiled middleware is missing from the traced output, and if the Turbopack runtime chunk cannot be found. `nodejs_compat` is still required, and a warning is logged that the support is experimental.

## Tests

- `examples/e2e/experimental` now uses `src/proxy.ts` instead of `src/middleware.ts`, and the four previously skipped tests in `e2e/nodeMiddleware.test.ts` are enabled.
- All existing unit tests pass (342/342).
- Verified end to end on a real Cloudflare Worker (see screenshots) with a `proxy.ts` exercising `node:crypto` (`randomUUID`, `createHash`, `createHmac`), `Buffer`, `node:path`, a direct short-circuit response, redirect, rewrite, cookie read/write, query and header access, `await` inside the middleware, request-header override via `NextResponse.next({ request })`, response headers, and a `config.matcher` — 18/18 live assertions pass, on a Turbopack build.

## Known limits

- Support is experimental and requires `nodejs_compat`.
- The middleware bundle increases worker size, since it is now bundled instead of loaded at runtime.
- Wasm bindings inside middleware are untested.

## Screenshots

---

<a href="https://app.devin.ai/review/opennextjs/opennextjs-cloudflare/pull/1309" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### 1. Build
*what main does today (the build refuses to continue), then what this PR does*

Build fails on main branch
Build succeeds on this PR

### 2. Live production
*18/18 assertions pass; "runtime":"nodejs" confirms it runs on the Node.js runtime*

Live production 18/18 assertions pass
Node.js runtime confirmation

### 3. Tests
*unit tests, code checks, and the four Node middleware e2e tests this PR enables*

Passed tests and code checks


ScienHAC:
@conico974 any update ??

conico974:
I'll have a look this week when I get time.

evanlong-me:
Rebased this onto current `main` (1.20.2) and re-verified build + local runtime end-to-end (including `node:crypto` in `proxy.ts`).

Opened a merge-ready PR so it can ship from the **publishing** repo without waiting on the adapters-api stack:

→ https://github.com/opennextjs/opennextjs-cloudflare/pull/1320

Full credit remains yours — 1320 is just a rebase + verification so maintainers have a clean branch to land. Happy to close 1320 in favor of updating this PR if you prefer to push the rebase here instead.

packages/cloudflare/src/cli/build/open-next/bundle-node-middleware.ts
```diff
@@ -0,0 +1,245 @@
...
+		treeShaking: true,
+		conditions: ["module"],
+		mainFields: ["module", "main"],
+		external: ["node:*", "./open-next.config.mjs"],
+		define: {
+			// The base of the middleware compiled by Next.js is runtime agnostic.
+			// "edge" skips the setup of the Node.js environment (`setup-node-env.external.js`
+			// patches globals that are read-only in workerd). Node.js builtin modules used by
+			// the middleware are provided by workerd via `nodejs_compat`.
+			"process.env.NEXT_RUNTIME": '"edge"',
```
conico974, line 183: What is that ?

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
```diff
@@ -253,6 +253,40 @@ function discoverExternalSubpaths(mappings: Map<string, string>, tracedFiles: st
...
+ *
+ * The runtime resolves the chunks it needs at runtime, which workerd does not support.
+ * The chunks are inlined so that they are statically bundled.
+ *
+ * @param code The code of the Turbopack runtime.
+ * @param filePath Path to the Turbopack runtime, in the OpenNext output.
+ * @param tracedFiles The files traced by Next.js, copied next to the runtime.
+ * @returns The patched code.
+ */
+export function patchTurbopackRuntimeCode({
```
conico974, line 267: Why this change ? Can we please stop making useless change

packages/cloudflare/src/cli/build/bundle-server.ts
```diff
@@ -62,7 +62,7 @@ export async function bundleServer(buildOpts: BuildOptions, projectOpts: Project
 
 	console.log(`\x1b[35m⚙️ Bundling the OpenNext server...\n\x1b[0m`);
 
-	await patchWebpackRuntime(buildOpts);
+	await patchWebpackRuntime(path.join(dotNextPath, "server"));
```
conico974, line 65: Why ?

packages/cloudflare/src/cli/build/open-next/bundle-node-middleware.ts
```diff
@@ -0,0 +1,245 @@
...
+ * `@opennextjs/aws` bundles the external middleware for a Node.js server:
+ * the OpenNext config is read from the filesystem at runtime and the middleware compiled
+ * by Next.js is loaded with `await import("./.next/server/middleware.js")`.
+ *
+ * workerd can not access the filesystem nor load modules at runtime, so the handler
+ * built by `@opennextjs/aws` is replaced with a fully self-contained bundle:
+ *
+ * - the config manifests are inlined by `openNextEdgePlugins` (as for the edge middleware)
+ * - the middleware compiled by Next.js is statically bundled from the traced files that
+ *   `@opennextjs/aws` copies to `middleware/<package path>/.next/server/middleware.js`
```
conico974, line 18: Add in a comment that this is not supported by original maintainer, and will **never** be.

packages/cloudflare/src/cli/build/open-next/bundle-node-middleware.ts
```diff
@@ -0,0 +1,245 @@
...
+}
+
+/**
+ * Resolves the middleware compiled by Next.js to the copy created by `copyTracedFiles`.
+ */
+export function setCompiledMiddlewarePlugin(compiledMiddlewarePath: string): Plugin {
+	return {
+		name: "compiled-middleware",
+		setup(build) {
+			build.onResolve({ filter: getCrossPlatformPathRegex("./.next/server/middleware.js") }, () => ({
```
conico974, line 83: Where is this reached ? We should never in theory reach this...

packages/cloudflare/src/cli/build/open-next/bundle-node-middleware.ts
```diff
@@ -0,0 +1,245 @@
...
+		},
+		alias: {
+			path: "node:path",
+			stream: "node:stream",
+			fs: "node:fs",
+			// `next/dist/server/lib/trace/tracer.js` requires `@opentelemetry/api`, an optional
+			// dependency that most apps do not install. On the Node.js runtime Next.js falls back
+			// to its own compiled copy when the require throws, but not on the edge runtime.
+			// Use the compiled copy, which is always present.
+			"@opentelemetry/api": "next/dist/compiled/@opentelemetry/api",
```
conico974: This also means that a real dependency won't work. I don't care that much since it will not be supported really but still

packages/cloudflare/src/cli/build/patches/ast/webpack-runtime.ts
```diff
@@ -73,22 +72,16 @@ fix: |
  * Fixes the webpack-runtime.js and webpack-api-runtime.js files by inlining
  * the webpack dynamic requires.
  */
-export async function patchWebpackRuntime(buildOpts: BuildOptions) {
-	const { outputDir } = buildOpts;
-
-	const dotNextServerDir = join(
```
conico974, line 79: Where did this go ?

packages/cloudflare/src/cli/build/open-next/bundle-node-middleware.ts
```diff
@@ -0,0 +1,245 @@
...
+ * `require` calls are converted into ESM imports via a virtual module. The virtual module
+ * re-exports the default export so that the interop helpers in the middleware compiled by
+ * Next.js receive the full module (`export * from ...` alone would drop the default export).
+ */
+export function nodeBuiltinsPlugin(): Plugin {
+	const namespace = "node-builtins";
+	return {
+		name: namespace,
+		setup(build) {
+			build.onResolve({ filter: /.*/ }, ({ path: specifier, kind }) => {
```
conico974: This is literally running on everything... Can we please not do that

packages/cloudflare/src/cli/build/open-next/bundle-node-middleware.ts
```diff
@@ -0,0 +1,245 @@
...
+				if (!isBuiltin(specifier)) {
+					return undefined;
+				}
+				const prefixed = specifier.startsWith("node:") ? specifier : `node:${specifier}`;
+				return kind === "require-call" ? { path: prefixed, namespace } : { path: prefixed, external: true };
+			});
+			build.onLoad({ filter: /.*/, namespace }, ({ path: builtin }) => ({
```
conico974: Same as onResolve

packages/cloudflare/src/cli/build/open-next/bundle-node-middleware.ts
```diff
@@ -0,0 +1,245 @@
...
+		banner: {
+			js: `
+import { Buffer } from "node:buffer";
+globalThis.Buffer = Buffer;
+
+import { AsyncLocalStorage } from "node:async_hooks";
+globalThis.AsyncLocalStorage = AsyncLocalStorage;
+
+// Next.js sets \`__import_unsupported\` on \`globalThis\` with \`configurable: false\`.
+// The middleware and the server both run this code in the same Worker, the second
```
conico974: This is an assumption that is not necessarily true

packages/cloudflare/src/cli/build/open-next/bundle-node-middleware.ts
ScienHAC: The base of the middleware compiled by Next.js is runtime-agnostic. With `NEXT_RUNTIME="edge"` it skips `setup-node-env.external.js`, which patches globals that are read-only in workerd (e.g. it would try to reassign `globalThis`). The Node.js builtins the middleware actually uses are still provided by workerd through `nodejs_compat`. Open to a clearer approach if you have one in mind.

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
ScienHAC: This only extracts the existing chunk-inlining logic into `patchTurbopackRuntimeCode` so the middleware bundle can reuse it instead of duplicating it — the `CodePatcher` export and its behaviour are unchanged. If you would rather I not touch this file, I can duplicate the small part I need on the middleware side instead. Let me know which you prefer.

packages/cloudflare/src/cli/build/bundle-server.ts
ScienHAC: Same reason as the turbopack change: `patchWebpackRuntime` now takes the `.next/server` dir so it can run on the middleware output too, not only the main server. Happy to revert to the old signature and duplicate instead if you prefer the shared helper untouched.

packages/cloudflare/src/cli/build/open-next/bundle-node-middleware.ts
ScienHAC: Agreed — in practice it is not reached because the compiled middleware is statically bundled, so `./.next/server/middleware.js` is resolved before this. It is a safety net so the build fails clearly rather than emitting a broken dynamic import if that ever changes. I can drop it if you would rather not carry the dead path.

packages/cloudflare/src/cli/build/open-next/bundle-node-middleware.ts
ScienHAC: Done — added a note at the top of the file (commit f3bdab8) stating that Node.js middleware on workerd is experimental, not supported by the OpenNext maintainers, and relies on re-bundling an internal Next.js output that can change between versions.

packages/cloudflare/src/cli/build/open-next/bundle-node-middleware.ts
ScienHAC: Fixed in 9c11f4f — the alias is now applied only when the app has not installed `@opentelemetry/api`. If a real one is installed, it is used; otherwise it falls back to the copy compiled in Next.js (which the edge runtime does not do on its own).

packages/cloudflare/src/cli/build/patches/ast/webpack-runtime.ts
ScienHAC: `patchWebpackRuntime` was changed to take the `.next/server` directory directly instead of `BuildOptions`, so it can be called for the middleware output as well as the server. The body did not move — only the argument it derives `dotNextServerDir` from. Callers are updated in `bundle-server.ts`.

packages/cloudflare/src/cli/build/open-next/bundle-node-middleware.ts
ScienHAC: Fixed in 9c11f4f — the `onResolve` filter is now a regex built from `builtinModules` (plus the `node:` prefix), so esbuild only calls back for actual Node.js builtins instead of every module.

packages/cloudflare/src/cli/build/open-next/bundle-node-middleware.ts
ScienHAC: Fixed in 9c11f4f — the `onLoad` filter is now `/^node:/` and scoped to the plugin namespace, so it only runs for the builtins this plugin resolves.

packages/cloudflare/src/cli/build/open-next/bundle-node-middleware.ts
ScienHAC: You are right. The guard is defensive regardless: it only skips the redefine when `globalThis.__import_unsupported` is already set, and otherwise runs `Object.defineProperty` as usual — so it is correct whether or not the middleware and server share a Worker. I reworded the comment in 9c11f4f to not state the shared-Worker case as a certainty.

ScienHAC:
Thanks for the detailed review — I have pushed fixes for all of it:

- **`@opentelemetry/api` alias** → now conditional; only used when the app has not installed the real package (9c11f4f)
- **broad esbuild filters** → the node-builtins plugin now filters on actual builtins only, not every module (9c11f4f)
- **`__import_unsupported` comment** → reworded; the guard was already defensive (9c11f4f)
- **`.wasm` traced files / `Object.defineProperty` return** → from the earlier bot review (32ef5fc)
- **experimental / unsupported note** → added at the top of the file as requested (f3bdab8)

The refactors you flagged (`patchTurbopackRuntimeCode`, `patchWebpackRuntime` signature) only exist to reuse the chunk-inlining logic for the middleware instead of duplicating it — I left inline replies offering to revert those to keep the shared helpers untouched if you prefer.

I also see you are adding Node middleware in adapters-api#38. Since the published `@opennextjs/cloudflare` still builds from this repo and Next 16 apps currently fail on release, I opened this as a stopgap until that lands. Happy to close in favour of #38 whenever it publishes — whichever you prefer.

packages/cloudflare/src/cli/build/patches/ast/webpack-runtime.ts
ScienHAC: Approach: `patchWebpackRuntime` now takes the `.next/server` directory instead of `BuildOptions`, so the same function works for the server output and the middleware output. The logic did not move — only the argument it derives the dir from.

```ts
// before: export async function patchWebpackRuntime(buildOpts: BuildOptions)
export async function patchWebpackRuntime(dotNextServerDir: string) { ... }
```

Now both callers reuse it:
```ts
// bundle-server.ts (server)
await patchWebpackRuntime(path.join(dotNextPath, "server"));
// bundle-node-middleware.ts (middleware)
await patchWebpackRuntime(dotNextServerDir);
```

If you would rather keep this signature as-is and have the middleware duplicate the small part it needs, I am happy to do that — I will not change it until you confirm which you prefer.

packages/cloudflare/src/cli/build/bundle-server.ts
ScienHAC: This is just the updated call site for the `patchWebpackRuntime` signature change above — passing `.next/server` explicitly so the same helper serves both the server and the middleware. No behaviour change for the server. I will leave it as-is unless you would prefer the old signature back (I will not revert until you confirm).

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
ScienHAC: Approach: I only extracted the existing body into `patchTurbopackRuntimeCode({ code, filePath, tracedFiles })` so the middleware can call the same code instead of duplicating it. The `CodePatcher` (`patchTurbopackRuntime`) still exists and behaves identically — it now delegates:

```ts
export const patchTurbopackRuntime: CodePatcher = {
  name: "inline-turbopack-chunks",
  patches: [{ ...,
    patchCode: async ({ code, tracedFiles, filePath }) =>
      patchTurbopackRuntimeCode({ code, filePath, tracedFiles }),
  }],
};
```
The middleware then reuses it directly for its own Turbopack runtime. If you prefer I not touch this file, I can duplicate the piece on the middleware side instead — I will not change it until you confirm.

packages/cloudflare/src/cli/build/open-next/bundle-node-middleware.ts
ScienHAC: You are right that it should not be reached in normal builds — the compiled middleware is statically bundled, so `./.next/server/middleware.js` is resolved by `setCompiledMiddlewarePlugin` before this. This resolver is only a safety net so the build fails clearly instead of emitting a broken dynamic `import()` if that ever stops matching:

```ts
build.onResolve({ filter: getCrossPlatformPathRegex("./.next/server/middleware.js") }, () => ({
  path: compiledMiddlewarePath,
}));
```
Happy to drop it entirely if you would rather not carry the unused path — I will not remove it until you confirm.

ScienHAC:
I want to be honest about my motivation: I'm not attached to this being "my" PR. I just need `proxy.ts` / Node.js middleware to work on Cloudflare for my own app, and right now Next 16 apps break on every release. So whoever lands it — this PR or your work in `adapters-api#38` — I'm genuinely grateful either way.

If there's anything else you'd like changed here, tell me and I'll do it. The two refactors you questioned only exist to reuse the chunk-inlining logic instead of duplicating it — if you'd prefer I revert them and keep the shared helpers untouched, just say the word and I'll push that. And if `#38` is the intended home and this PR is better off closed, I'm happy to close it.

ScienHAC:
A bit more detail on the two refactors you questioned, since they are the only changes I have not already updated — I wanted to explain the logic rather than just leave them:

**1. `patchTurbopackRuntimeCode` (turbopack.ts)** — I only extracted the existing body into a function so the middleware can call the same code. The `patchTurbopackRuntime` `CodePatcher` still exists and behaves identically; it now delegates to it:
```ts
patchCode: async ({ code, tracedFiles, filePath }) =>
  patchTurbopackRuntimeCode({ code, filePath, tracedFiles }),
```

**2. `patchWebpackRuntime` (webpack-runtime.ts + bundle-server.ts)** — changed to take the `.next/server` directory instead of `BuildOptions`, so the same helper works for the server and the middleware. The logic did not move, only the argument:
```ts
// bundle-server.ts (server, unchanged behaviour)
await patchWebpackRuntime(path.join(dotNextPath, "server"));
// bundle-node-middleware.ts (middleware, reuses the same helper)
await patchWebpackRuntime(dotNextServerDir);
```

The intent was to reuse the chunk-inlining logic instead of duplicating it across the server and the middleware. If you would rather I keep those shared helpers untouched, I can revert both and duplicate the small part the middleware needs — I have not changed them and will not until you confirm which you prefer.

packages/cloudflare/src/cli/build/open-next/bundle-node-middleware.ts
conico974: I don't really like the solution TBH, this could have other side effects. But as I already said, this is just to unblock this, and I don't intend on providing any support so we can keep it

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
conico974: Sorry missed that

packages/cloudflare/src/cli/build/open-next/bundle-node-middleware.ts
conico974: Drop it please. If we reach it it means something else is broken, and we should do something about it

packages/cloudflare/src/cli/build/build.ts
```diff
@@ -81,10 +82,11 @@ export async function build(
...
 	}
 
-	// Make sure no Node.js middleware is used
-	if (useNodeMiddleware(options)) {
-		logger.error("Node.js middleware is not currently supported. Consider switching to Edge Middleware.");
-		process.exit(1);
+	const hasNodeMiddleware = useNodeMiddleware(options);
+	if (hasNodeMiddleware) {
+		logger.warn(
+			"Node.js middleware support is experimental, make sure that `nodejs_compat` is enabled in your wrangler configuration."
```
conico974:
```suggestion
			"Node.js middleware support is experimental in cloudflare, and not officially maintained by OpenNext maintainers. Use at your own risk.

**Review by conico974 — COMMENTED** [review]
Just 2 little changes, but after that it should be good

File: packages/cloudflare/src/cli/build/open-next/bundle-node-middleware.ts (hunk 5 above)
Comment (ScienHAC) [reply packages/cloudflare/src/cli/build/open-next/bundle-node-middleware.ts:83]: I tried dropping it and the build fails:

```
✘ [ERROR] Could not resolve "./.next/server/middleware.js"
  node_modules/@opennextjs/aws/dist/core/nodeMiddlewareHandler.js:13:38
```

It turns out this is not an error path — it is the normal resolution. `nodeMiddlewareHandler.js` does `await import("./.next/server/middleware.js")`, and esbuild resolves that relative to the handler's own location (`node_modules/@opennextjs/aws/dist/core/`), where the traced middleware is not. This plugin redirects it to the copy `copyTracedFiles` puts under `middleware/<pkg>/.next/server/middleware.js`.

So without it there is no way for esbuild to find the compiled middleware. I can keep it (as you were fine doing for the sibling comment), or if you would prefer the handler resolve it some other way I am happy to follow — but it cannot simply be removed without breaking the build. Let me know which you prefer.

Comment (ScienHAC) [reply packages/cloudflare/src/cli/build/open-next/bundle-node-middleware.ts:83]: To be concrete about why it is reached — the import is hardcoded in the aws handler (`@opennextjs/aws` 4.0.2), `core/nodeMiddlewareHandler.js`:

```js
//@ts-expect-error - This file should be bundled with esbuild
_module = await (await import("./.next/server/middleware.js")).default;
```

and `openNextExternalMiddlewarePlugin` only resolves `./middleware.mjs`, not `./.next/server/middleware.js`. So this specifier is always emitted and nothing on the aws side resolves it — the `@ts-expect-error "should be bundled with esbuild"` note is essentially delegating that to the adapter. That is what this plugin does.

If the cleaner fix is for the aws handler to import a path the adapter can resolve without a redirect, that would be an aws-side change. For this PR I think keeping the resolver is the minimal way to keep it building — but happy to do whatever you prefer.

ScienHAC:
Both addressed:

1. Warning message — updated to your wording. ✅

2. Dropping the `compiled-middleware` resolver — I tried, and the build fails:
```
✘ Could not resolve "./.next/server/middleware.js"
  node_modules/@opennextjs/aws/dist/core/nodeMiddlewareHandler.js:13
```
It is not a dead path: the aws handler hardcodes that import (`@ts-expect-error - This file should be bundled with esbuild`) and `openNextExternalMiddlewarePlugin` only resolves `./middleware.mjs`, so the adapter has to resolve it. Kept it so the build stays green — happy to switch to an aws-side import if you prefer (details in the inline thread).

Tests: 342/342 unit pass; 4/4 Node-middleware e2e pass on a real workerd (headers, JSON, redirect, rewrite).

evanlong-me:
@conico974 following up on the one open point from your Aug 8 review ("Drop it please. If we reach it it means something else is broken").

I traced where the `compiled-middleware` resolver is reached, since dropping it breaks the build. The full chain, all verifiable in the published `@opennextjs/aws@4.1.0` dist:

1. `adapters/middleware.js` → `core/routing/middleware.js:12` does `import("./middleware.mjs")`
2. `openNextExternalMiddlewarePlugin` maps `./middleware.mjs` → `core/nodeMiddlewareHandler.js`
3. `nodeMiddlewareHandler.js:13` does `await import("./.next/server/middleware.js")`

esbuild resolves that string-literal dynamic import at bundle time, relative to `node_modules/@opennextjs/aws/dist/core/` — where the file does not exist. aws's own `buildExternalNodeMiddleware` avoids this with `external: ["./.next/*"]`, leaving it as a runtime filesystem import. That works on Lambda but is impossible on workerd, so for a self-contained bundle the import **must** be resolved statically — the resolver is reached on every build here by design, not because something else is broken. This matches the exact error @ScienHAC posted (`Could not resolve "./.next/server/middleware.js"` from `nodeMiddlewareHandler.js:13`).

So "we should never reach this" holds for the aws build (where `./.next/*` is external), but not for this one.

If the concern is the filter being broad, it could be scoped to only fire for that one importer:

```ts
build.onResolve(
  { filter: getCrossPlatformPathRegex("./.next/server/middleware.js") },
  (args) =>
    getCrossPlatformPathRegex("nodeMiddlewareHandler.js").test(args.importer)
      ? { path: compiledMiddlewarePath }
      : undefined,
);
```

Functionally the current version already does the right thing, though.

With the warning wording fixed in 0762b49, I believe both items from your last review are addressed. Could you take another look — and approve the workflow runs? There are currently no checks reported on the branch. Would be great to get this landed.

vicb (dismissed):
Thanks for the PR and sorry for being late at this party!

We will merge the PR, coudl you please make sure it is rebased on the main branch. I hope to be able to review later in the day.

Thanks a lot for your work!

packages/cloudflare/src/cli/build/open-next/bundle-node-middleware.ts
```diff
@@ -0,0 +1,257 @@
...
+
+/**
+ * Makes the Node.js builtins used by the bundled code (i.e. `require("crypto")` or
+ * `import from "node:crypto"`) resolve to the modules workerd provides via `nodejs_compat`.
+ *
+ * `require` calls are converted into ESM imports via a virtual module. The virtual module
+ * re-exports the default export so that the interop helpers in the middleware compiled by
+ * Next.js receive the full module (`export * from ...` alone would drop the default export).
+ */
+export function nodeBuiltinsPlugin(): Plugin {
```
vicb, line 113: Could you expand on why this is needed now?
It should be handled by the wrangler build step - for example it is working outside of middlewares

packages/cloudflare/src/cli/build/open-next/bundle-node-middleware.ts
vicb: Agreed with @conico974 that this looks fishy as the whole purpose of this PR is to enable Node middlewares.

Could `setCompiledMiddlewarePlugin` could be updated to remove the code path using ast-grep (as we do with other files)?

packages/cloudflare/src/cli/build/open-next/bundle-node-middleware.ts
```diff
@@ -0,0 +1,257 @@
...
+			// "edge" skips the setup of the Node.js environment (`setup-node-env.external.js`
+			// patches globals that are read-only in workerd). Node.js builtin modules used by
+			// the middleware are provided by workerd via `nodejs_compat`.
+			"process.env.NEXT_RUNTIME": '"edge"',
+			"process.env.NODE_ENV": '"production"',
+		},
+		alias: {
+			path: "node:path",
+			stream: "node:stream",
+			fs: "node:fs",
```
vicb: Why are those aliases needed?

packages/cloudflare/src/cli/build/open-next/bundle-node-middleware.ts
```diff
@@ -0,0 +1,257 @@
...
+				isInCloudflare: true,
+			}),
+		] as Plugin[],
+		banner: {
+			js: `
+import { Buffer } from "node:buffer";
+globalThis.Buffer = Buffer;
+
+import { AsyncLocalStorage } from "node:async_hooks";
+globalThis.AsyncLocalStorage = AsyncLocalStorage;
```
vicb, line 250: Is this still required?
It should be the default behavior in workerd now ?

evanlong-me:
@ScienHAC @vicb — the branch is rebased onto current `main` (2f863ad) and today's CI break is fixed. Ready for you to update this PR:

```
git remote add evanlong https://github.com/evanlong-me/opennextjs-cloudflare.git
git fetch evanlong pr-1309-rebase
git switch feat/proxy-node-middleware && git reset --hard evanlong/pr-1309-rebase
git push --force-with-lease origin feat/proxy-node-middleware
```

What's in it (your commits preserved, rebased):

1. **Today's red CI**: `0762b49` applied the warning wording via the web UI and the string lost its closing quote — `build.ts:88` "Unterminated string literal" fails `tsc` during the `pnpm install` postinstall, so none of the checks ever got to run. One-character fix, folded into that commit.

2. @vicb's review points from today:
   - *aliases* — not needed: `nodeBuiltinsPlugin` already resolves bare builtins to external `node:*`, so the `path`/`stream`/`fs` aliases resolved the same imports twice. Removed (59c40f2).
   - *AsyncLocalStorage banner* — still required. Without it the worker crashes at startup: `ReferenceError: AsyncLocalStorage is not defined` at `adapters/middleware.js:8` (`globalThis.__openNextAls = new AsyncLocalStorage()`), on wrangler 4.86 / compatibility_date 2026-04-15 / `nodejs_compat`. aws does the same for the edge bundle (`createEdgeBundle.js:79`) and for its node middleware build (`buildNodeMiddleware.js:76`).
   - *nodeBuiltinsPlugin vs the wrangler build step* — I tried exactly that (`platform: "node"`, no plugin, `__require`→`require` post-processing like the server bundle): works on Turbopack, but breaks on webpack, where `import crypto from "node:crypto"` compiles to a `.default` access that ends up `undefined` (`TypeError: Cannot read properties of undefined (reading 'randomUUID')` at request time). The virtual module bridges the default-export interop for both bundlers, so it stays.
   - *NEXT_RUNTIME `"nodejs"` + ast-grep removal of the code path* — tried: even with the `if (process.env.NEXT_RUNTIME !== 'edge')` block removed from the traced `setup-node-env.external.js` via ast-grep, another NEXT_RUNTIME-gated path then bundles `next/dist/compiled/next-server/app-page-turbo.runtime.prod.js`, which patches read-only workerd globals (`setImmediate`) and 500s at request time. `"edge"` is not claiming this is edge middleware — it keeps Next's bundled internals on their web-platform branches, which is the surface workerd implements; patching every gated path one by one would be very fragile across Next versions.

3. Verified end to end on real workerd (`wrangler dev`, `nodejs_compat`) with a Next 16.2.11 `proxy.ts` app, on **both** bundlers (`next build` and `next build --webpack`): `node:crypto` (`randomUUID`/`createHash`/`createHmac`), `Buffer`, a direct `NextResponse.json`, 307 redirect, rewrite, and response headers via `NextResponse.next()` — all pass. `pnpm code:checks` clean, 342/342 unit tests pass.

(Local note: the e2e example fails `next build` while prerendering `/_global-error` on macOS — that's vercel/next.js#86178, unrelated to this PR; I built with `--experimental-build-mode=compile` to skip prerendering.)

vicb:
> git remote add evanlong https://github.com/evanlong-me/opennextjs-cloudflare.git

I know you already did that in #1320 but could you please open a PR and we'll get that merged.

What would be the best way:

- start from this PR
- merge the main branch into your branch (easier to review than doing a rebase)
- add commits for each of the fixes

That way we can clearly attribute the work to both @ScienHAC and you.

> nodeBuiltinsPlugin vs the wrangler build step — I tried exactly that (platform: "node", no plugin, __require→require post-processing like the server bundle): works on Turbopack, but breaks on webpack, where import crypto from "node:crypto" compiles to a .default access that ends up undefined (TypeError: Cannot read properties of undefined (reading 'randomUUID') at request time). The virtual module bridges the default-export interop for both bundlers, so it stays.

I see that's because we do a full build here.

It would still be good to add a link to the [ESBuild plugin used for the same reason in `wrangler`](https://github.com/cloudflare/workers-sdk/blob/c457bfc6b5a575586354f5b0ad7a1100eff915fe/packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts#L102-L133) and make sure they are in sync as the `wrangler` one has been battle tested for years.

One last thing that would be nice is to update `playground-16` app and e2e instead of the experimental one. This last one will probably be removed at some point.

Thanks for your help

ScienHAC:
Thanks @vicb — really appreciate it, glad it's landing! 🙏

Small note: I just cleaned up the branch, so it's at `5e9e3fb` now (same code).

@evanlong-me since you've already got it rebased on latest `main`, I'm happy for you to push the final version onto this branch (with `Co-authored-by` so we're both credited) — including the `nodeBuiltinsPlugin` → wrangler `hybrid-nodejs-compat` link and moving the example to `playground16`. Thanks for carrying it over the line! 🙌

evanlong-me:
@ScienHAC thanks for the quick turnaround on the fixes — and for the generous co-author offer! Your commits are kept as-is in the new PR, so the credit stays with you.

Opened #1357 as requested, @vicb: it starts from this branch (both of @ScienHAC's latest commits included), merges current main, and adds the remaining fixes as separate commits — the wrangler plugin link and the move of the e2e coverage to playground16 included. Details in the PR body.

vicb:
@ScienHAC Thank you so much again for initiating this PR.

Kudos also to @evanlong-me to add more  commits.

Thanks to @conico974 for the review 🙏

You both deserve all the credits and will please a lot of our users.

I only added a commit to fix `nodeBuiltinsPlugin`, sync it with `wrangler` and document the differences.

Please :+1: this message if you both are happy with the state of this branch. I'll wait for that to merge the PR.

vicb (approved):
Thanks @ScienHAC and @evanlong-me !

PR #1316, opened 2026-07-23
fix: ignore Vercel OG traces for split functions

### LINKED ISSUE #1314: `functions` route-splitting crashes in patchVercelOgLibrary (ENOENT) when a split-out route's trace includes @vercel/og
### Versions
- `@opennextjs/cloudflare`: 1.19.10 (bundled `@opennextjs/aws` 4.0.2)
- `next`: 16.2.6
- Node 22.23.1, reproduced on both Linux (Ubuntu WSL2) and Windows 11

### Summary

Declaring a named function in `open-next.config.ts` via the AWS-style `functions` option correctly excludes the claimed routes from `server-functions/default` (`createServerBundle` builds `default` from `remainingRoutes`) — but the build then crashes in `patchVercelOgLibrary` with ENOENT if any *split-out* route's `.nft.json` trace includes `@vercel/og`:

```
Error: ENOENT: no such file or directory, open
'.../.open-next/server-functions/default/.next/server/app/vs/smugmug/page.js'
    at parseFile (@opennextjs/aws/dist/build/patch/astCodePatcher.js:51:22)
    at patchVercelOgLibrary (@opennextjs/cloudflare/dist/cli/build/patches/ast/patch-vercel-og-library.js:55:25)
    at bundleServer (@opennextjs/cloudflare/dist/cli/build/bundle-server.js:56:19)
```

### Root cause

`patch-vercel-og-library.js` globs **all** `.next/server/**/*.nft.json` from the app build output and, for every trace that references `@vercel/og/index.node.js`, derives the corresponding route file path under `server-functions/default` and parses it:

```js
const routeFilePath = traceInfoPath.replace(appBuildOutputPath, packagePath).replace(".nft.json", "");
const ast = parseFile(routeFilePath); // ENOENT for routes claimed by a named function
```

With `config.functions` splitting, routes claimed by a named function are (correctly) not copied into `server-functions/default`, so the read fails.

### Repro

1. Next app with at least one route whose trace includes `@vercel/og` (e.g. a colocated `opengraph-image.tsx`).
2. `open-next.config.ts`:
   ```ts
   import { defineCloudflareConfig } from '@opennextjs/cloudflare'
   const config = defineCloudflareConfig({})
   export default {
     ...config,
     functions: {
       marketing: { routes: ['app/vs/smugmug/page', /* … */], patterns: ['/vs/*'] },
     },
   }
   ```
3. `npx opennextjs-cloudflare build` → ENOENT above.

### Suggested fix

Skip traces whose route file isn't present in the default function's output:

```js
const routeFilePath = traceInfoPath.replace(appBuildOutputPath, packagePath).replace(".nft.json", "");
if (!existsSync(routeFilePath)) continue;
const ast = parseFile(routeFilePath);
```

Verified locally: with that one-line guard the build completes and the split behaves as expected (claimed routes excluded from `default`, `default` bundle correspondingly smaller).

### Related (separate issue if useful)

On **Windows**, `functions` splitting silently no-ops even before reaching this crash: the route keys derived in the CF adapter's `createServerBundle.js` (`app/${relativePath.replace(/\.js$/, "")}`) contain backslashes from `path.join`, while `validateConfig` enforces forward-slash `app/...` route strings in config — so `foundRoutes` never matches and `remainingRoutes` retains every route. Happy to file that separately if wanted.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

james-elicx:
## Summary

- skip Vercel OG traces whose route is not present in the default server function
- keep OG usage detection and asset patching scoped to routes bundled into that function
- add regression coverage for mixed and split-only OG routes

## Root cause

`patchVercelOgLibrary` scanned NFT traces from the complete Next.js build, but parsed the corresponding route under `server-functions/default`. Named-function routes remain in the build traces while being intentionally absent from the default package, causing ENOENT during parsing.

## Testing

- `pnpm --filter cloudflare exec vitest --run src/cli/build/patches/ast/patch-vercel-og-library.spec.ts`
- `pnpm code:checks`
- `pnpm test` (34 test files, 343 tests)

Closes #1314

### DIFF
--- a/.changeset/tidy-pandas-build.md
+++ b/.changeset/tidy-pandas-build.md
@@ -1,0 +1,7 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+fix: ignore Vercel OG traces for routes assigned to split functions.
+
+Only patch Vercel OG files that are present in the default server function, preventing builds from failing when another function owns an OG route.
--- a/packages/cloudflare/src/cli/build/patches/ast/patch-vercel-og-library.spec.ts
+++ b/packages/cloudflare/src/cli/build/patches/ast/patch-vercel-og-library.spec.ts
@@ -1,14 +1,15 @@
-import { mkdirSync, readdirSync, readFileSync, writeFileSync } from "node:fs";
+import { existsSync, mkdirSync, readdirSync, readFileSync, unlinkSync, writeFileSync } from "node:fs";
 import path from "node:path";
 
 import { BuildOptions } from "@opennextjs/aws/build/helper.js";
 import mockFs from "mock-fs";
-import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
 
 import { patchVercelOgLibrary } from "./patch-vercel-og-library.js";
 
 const nodeModulesVercelOgDir = "node_modules/.pnpm/next@14.2.11/node_modules/next/dist/compiled/@vercel/og";
 const nextServerOgNftPath = "examples/api/.next/server/app/og/route.js.nft.json";
+const splitFunctionOgNftPath = "examples/api/.next/server/app/split/page.js.nft.json";
 const openNextFunctionDir = "examples/api/.open-next/server-functions/default/examples/api";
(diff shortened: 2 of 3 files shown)


james-elicx:
Feels like is probably just masking a separate issue rather than a proper fix.

PR #1332, opened 2026-08-06
fix: retrieveCompiledConfig should respect buildOutputPath

yuanzhixiang:
## What

`retrieveCompiledConfig` resolves the compiled config from a hardcoded path:

https://github.com/opennextjs/opennextjs-cloudflare/blob/main/packages/cloudflare/src/cli/commands/utils/utils.ts#L98

`<cwd>/.open-next/.build/` does not follow the `buildOutputPath` config, so with `buildOutputPath` set, every command that goes through this function — `deploy`, `preview`, `upload`, `populateCache` — exits immediately after a successful build:

```
┌──────────────────────────────┐
│ OpenNext — Cloudflare deploy │
└──────────────────────────────┘

ERROR Could not find compiled Open Next config, did you run the build command?
```

`build` is unaffected because it compiles the config from source via `compileConfig`, which is why this only surfaces at deploy time.

## Approach

The path can't simply be prefixed with `buildOutputPath`: that value lives in the very config being loaded, so it isn't known at this point. Instead, when the compiled file is missing, this falls back to recompiling from the source config — the same `compileConfig` path `build` takes. `compileOpenNextConfig` emits to `os.tmpdir()`, so nothing is written into the project.

Kept as a fallback rather than replacing the fast path, so projects without a custom `buildOutputPath` behave exactly as before.

If you'd rather have the command layer resolve `buildOutputPath` first and keep a single load path, I'm happy to rework it that way — just let me know.

## Related

`@opennextjs/aws` has the same class of bug in `getBundlerRuntime`, which breaks the *build* under a custom `buildOutputPath`. Fixed in opennextjs/opennextjs-aws#1208. Both are needed for `buildOutputPath` to work end to end — with only one of them applied you get a green build and a broken deploy, or no build at all.

## Testing

Verified against a monorepo Next.js 16.3.0 app (Turbopack) with `buildOutputPath: ".cache"`, using both patches:

```
opennextjs-cloudflare build          → Worker saved in `.cache/.open-next/worker.js`
opennextjs-cloudflare deploy --dry-run
  → Populating Workers static assets... Successfully populated
  → Read 143 files from the assets directory .../.cache/.open-next/assets
  → --dry-run: exiting now.
```

`pnpm lint:check`, `pnpm ts:check`, `pnpm test` (342 tests) and `prettier --check` all pass locally.

---

<a href="https://app.devin.ai/review/opennextjs/opennextjs-cloudflare/pull/1332" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/retrieve-compiled-config-build-output-path.md
+++ b/.changeset/retrieve-compiled-config-build-output-path.md
@@ -1,0 +1,17 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+Fix `deploy`/`preview`/`upload`/`populateCache` failing when `buildOutputPath` is set
+
+`retrieveCompiledConfig` looked for the compiled config under a hardcoded
+`<cwd>/.open-next/.build/`, which does not follow the `buildOutputPath` config.
+Every command that goes through it therefore exited with `Could not find
+compiled Open Next config, did you run the build command?` right after a
+successful build. `build` itself was unaffected because it compiles the config
+from source, so the failure only showed up at deploy time.
+
+The compiled path cannot simply be prefixed with `buildOutputPath` — that value
+lives in the very config being loaded. When the file is missing, the config is
+now recompiled from source instead (the same path `build` takes;
+`compileOpenNextConfig` emits to a temp dir, so nothing lands in the project).
--- a/.changeset/retrieve-compiled-config-build-output-path.md
+++ b/.changeset/retrieve-compiled-config-build-output-path.md
@@ -2,16 +2,20 @@
 "@opennextjs/cloudflare": patch
 ---
 
-Fix `deploy`/`preview`/`upload`/`populateCache` failing when `buildOutputPath` is set
+fix: make `retrieveCompiledConfig` respect `buildOutputPath`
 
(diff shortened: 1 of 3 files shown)


yuanzhixiang:
Thanks — both review findings were valid, addressed in 28aed56.

**Missing-build case no longer swallowed.** The fallback conflated "`buildOutputPath` moved the compiled config" with "the build was never run". Now:

- `findOpenNextConfig` is checked before `compileConfig`, so a missing source config can never reach the create-a-config branch — these commands must not write to the project.
- After recompiling, the resolved output directory is checked for the built worker (`getOutputWorkerPath`). A source config on its own doesn't mean the app was built.

Both failure paths keep the original `Could not find compiled Open Next config, did you run the build command?` error, so running deploy/preview/upload/populateCache without building behaves as before.

**Changeset format.** Reformatted to the `<TYPE>: <TITLE>` first line required by CONTRIBUTING.md / AGENTS.md.

Also added unit tests for `retrieveCompiledConfig` covering the recompile path, both missing-build paths, and an assertion that no config file is ever created. `pnpm code:checks` and `pnpm test` (346 tests) pass.

Issue #1327, opened 2026-08-04
[BUG] Segment prefetch requests intermittently return 503 (worker outcome `ok`, no exceptions)


hellocomunamx:
## Environment

| Package | Version observed |
|---|---|
| `next` | 16.2.4 and 16.2.12 (reproduced on both) |
| `@opennextjs/cloudflare` | 1.19.4 and 1.20.2 (reproduced on both) |
| `wrangler` | 4.106.0 and 4.118.0 |
| Runtime | Cloudflare Workers, `compatibility_date 2024-12-30`, `nodejs_compat`, Workers Assets |
| App | App Router, fully dynamic pages (Supabase SSR, cookies), no PPR |

## Observed behavior

On authenticated page loads, one specific Next.js router prefetch request intermittently returns **HTTP 503** to the browser:

```
GET /dashboard/employees?company=<uuid>&_rsc=<key>   → 503
```

- The response `Vary` header of the route includes `rsc, next-router-state-tree, next-router-prefetch, next-router-segment-prefetch`, so this is the segment-prefetch machinery.
- Real navigations to the same URL always return 200. Only the prefetch fires the 503, on both Next 16.2.4 + adapter 1.19.4 and Next 16.2.12 + adapter 1.20.2 (upgraded specifically to test this).
- `wrangler tail` shows the worker EXECUTES those requests with `"outcome": "ok"`, `"exceptions": []`, cpuTime 24–33 ms (evidence captured across 350+ events, zero exceptions total). So the 503 status is a deliberate response produced inside the request handling, not a runtime failure or resource limit.
- Replaying the same URL with `RSC: 1` and `Next-Router-Prefetch: 1` headers (without the exact `Next-Router-State-Tree`/segment headers the router sends) returns 200 every time, so the failing shape is specifically the router-generated segment prefetch.
- The bursts are worst in the minutes after a deployment but a subset keeps reproducing at steady state (same `_rsc` cache key each time for a given build).

## Impact

No user-facing breakage (the router falls back to a normal navigation), but:

- error-rate monitoring at the edge reports persistent 5xx,
- during post-deploy bursts we observed full RSC navigation fetches also getting 503, which made `router.push()` after server actions silently fail (we had to add a client-side recovery timeout).

## Repro sketch

1. App Router app with fully dynamic, cookie-authenticated pages (we use `@supabase/ssr`).
2. Deploy with `opennextjs-cloudflare build && opennextjs-cloudflare deploy`.
[...]

PR #1348, opened 2026-08-19
fix: stop Next.js 16.3 from endlessly retrying page prefetches

### LINKED ISSUE #1334: Next 16.3.0 + OpenNext triggers an unbounded RSC prefetch loop
## Bug
With Next.js 16.3.0 and `@opennextjs/cloudflare` 1.20.2, an OpenNext preview triggers an unbounded RSC prefetch loop after page load. A fresh build and isolated Chrome profile produced **1,910 `Next-Router-Prefetch: 1` requests in 30 seconds** (~64 requests/sec), accelerating and repeating the same `_rsc` payload. This can amplify Workers traffic rapidly.

## Control
The same application with Next.js 16.2.12 + OpenNext 1.20.2 produced **0 prefetch requests in 30 seconds**. OpenNext 1.20.1 with Next.js 16.3.0 also reproduced the loop, so this is not an OpenNext-version-only regression.

Reproduction: build with `next build --turbopack`, build/serve with OpenNext Cloudflare, open the app using a fresh browser profile, and observe repeated RSC prefetch requests.

ahmadaccino:
## Summary

Fixes #1334.

- serve cached App Router segment payloads even when Next.js prefetch inlining is enabled
- apply a focused, fail-fast patch to the generated OpenNext middleware only when cache interception is enabled
- update `playground16` to Next.js 16.3.1 and add unit and Worker regression coverage plus a patch changeset

## Root cause

`@opennextjs/aws@4.1.0` only selected `segmentData` when `experimental.prefetchInlining` was falsy. Next.js 16.3 enables that setting by default and normalizes it to an object, so cache interception answered `Next-Router-Segment-Prefetch` requests with the full-page RSC payload instead of the requested segment.

Next.js still requests the `/_tree` response and any outlined segment bundles when inlining is enabled. Because the client never received the segment response or its `x-nextjs-postponed: 2` header, it kept requesting the route tree indefinitely. The reproduction generated 3,482 repeated `/_tree` requests in five seconds.

The patch removes only the incorrect prefetch-inlining gate and keeps the existing `segmentData` membership check. It is scoped to cache-interception builds and fails loudly if the pinned upstream implementation changes. The corresponding upstream report is opennextjs/opennextjs-aws#1212.

## Validation

- `pnpm code:checks`
- `pnpm test` (35 test files, 345 tests)
- `SKIP_NEXT_APP_BUILD=true pnpm --filter playground16 exec playwright test e2e/prefetch.cloudflare.spec.ts -c e2e/playwright.config.ts` (1 passed)
- `pnpm --filter playground16 exec playwright test e2e/prefetch.cloudflare.spec.ts -c e2e/playwright.turbopack.config.ts` (1 passed)

---

<a href="https://app.devin.ai/review/opennextjs/opennextjs-cloudflare/pull/1348" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/curvy-lions-prefetch.md
+++ b/.changeset/curvy-lions-prefetch.md
@@ -1,0 +1,7 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+fix: serve cached segment prefetches when Next.js prefetch inlining is enabled
+
+Prevent Next.js 16.3 clients from repeatedly requesting the route tree when cache interception is enabled.
--- a/examples/playground16/app/prefetch/page.tsx
+++ b/examples/playground16/app/prefetch/page.tsx
@@ -1,0 +1,5 @@
+import Link from "next/link";
+
+export default function PrefetchPage() {
+	return <Link href="/prefetch/target">Target</Link>;
+}
--- a/examples/playground16/app/prefetch/target/page.tsx
+++ b/examples/playground16/app/prefetch/target/page.tsx
@@ -1,0 +1,3 @@
+export default function PrefetchTargetPage() {
+	return <p>Target page</p>;
+}
--- a/examples/playground16/e2e/prefetch.cloudflare.spec.ts
+++ b/examples/playground16/e2e/prefetch.cloudflare.spec.ts
@@ -1,0 +1,43 @@
+import { expect, test } from "@playwright/test";
+
+test("cache interception serves segment prefetches without retrying indefinitely", async ({ page }) => {
+	const segmentPrefetches: string[] = [];
+	page.on("request", (request) => {
+		if (new URL(request.url()).pathname !== "/prefetch/target") {
+			return;
+		}
+
+		const segment = request.headers()["next-router-segment-prefetch"];
+		if (segment) {
+			segmentPrefetches.push(segment);
+		}
+	});
+
+	const routeTreeResponsePromise = page.waitForResponse((response) => {
+		const request = response.request();
+		return (
+			new URL(request.url()).pathname === "/prefetch/target" &&
+			request.headers()["next-router-segment-prefetch"] === "/_tree"
+		);
+	});
+
+	await page.goto("/prefetch");
+
+	const routeTreeResponse = await routeTreeResponsePromise;
+	expect(routeTreeResponse.status()).toBe(200);
+	expect(routeTreeResponse.headers()).toMatchObject({
+		"content-type": "text/x-component",
+		"x-nextjs-postponed": "2",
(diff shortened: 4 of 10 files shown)


spira-unplugged:
Confirming this against production and against an older adapter line. Two notes that may help scoping, plus a request.

I hit this on a live site: deploying next 16.2.6 -> 16.3.1 took a small app from 30-40k Workers requests/day to 329M and 175M on two consecutive days, with R2 Class B reads tracking it roughly 1:1 (details in #1334). Your root-cause description matches what I observed, so thank you for chasing it down.

### The ast-grep rule matches `@opennextjs/aws@3.10.2` verbatim, so this is not 4.1.0-only

The doc comment on `patchCacheInterceptor` says "OpenNext AWS 4.1.0 treats prefetch inlining as if it eliminates segment responses". `@opennextjs/cloudflare@1.19.2` pins `@opennextjs/aws@3.10.2`, and `getBodyForAppRouter` there emits the same expression:

```js
// node_modules/@opennextjs/aws/dist/core/routing/cacheInterceptor.js
const isSegmentResponse = Boolean(segmentHeader) &&
    segmentHeader in (cachedValue.segmentData || {}) &&
    !NextConfig.experimental?.prefetchInlining;
```

That is structurally identical to `segmentPrefetchRule`'s pattern, so `patchCacheInterceptorSource` should rewrite it cleanly on 1.19.x as well as 1.20.x. Worth reflecting in the doc comment and the changeset - as written, someone on 1.19.x reading this may conclude they are unaffected, and they are not.

### `patchCacheInterceptorSource` throws when nothing matches

Fail-fast is a reasonable default here, but it turns "pattern absent" into a hard build failure for every cache-interception user, including anyone on an aws version that renames or reshapes the expression without being buggy. Would it be worth throwing only when the known-bad `!NextConfig.experimental?.prefetchInlining` substring is present in a shape the rule cannot rewrite, and warning otherwise? Either behaviour is defensible - raising it because the failure would land on people who are not otherwise affected.

### Ask

Is there a maintainer who could review this? As far as I can see there is no PR against `@opennextjs/aws` for opennextjs/opennextjs-aws#1212 yet, so this patch is currently the only route to a fix for anyone running cache interception on Next 16.3. #1334 and #1327 have no maintainer response either, and the failure mode is unbounded client-side retries against a metered runtime.

Happy to run a canary build on a preview deployment of the app that hit this and report the numbers back.

conico974 (changes requested):
Just fix it upstream, we don't want to patch OpenNext code itself

PR #1353, opened 2026-08-23
fix(cloudflare): declare esbuild as runtime dependency

### LINKED ISSUE #1339: esbuild is imported at build time but not declared as a dependency — works only by hoisting accident, breaks under npm ci
**Version:** @opennextjs/cloudflare 1.20.2 (also inspected 1.20.x line)

`@opennextjs/cloudflare` imports `esbuild` during `opennextjs-cloudflare build`, but declares it only in its own **devDependencies** (`^0.27.0`) — consumers don't get it. In practice the import resolves against `@opennextjs/aws`'s nested `esbuild` (pinned 0.25.4) **only when npm happens to hoist it to the consumer's top level**.

**How it breaks:** any consumer whose dependency graph claims that hoist slot loses the build. In our case, adding `@vitejs/plugin-react` hoisted `vite`, whose **optional** `esbuild` peer (`^0.27.0 || ^0.28.0`) reserved the top-level slot without filling it (optional peers are omitted by npm 11), demoting aws's copy → `npx opennextjs-cloudflare build` fails with `ERR_MODULE_NOT_FOUND: Cannot find package 'esbuild'`.

**The nasty variant:** a fresh `npm install` can build a tree that works while writing the lockfile entry for esbuild as `"extraneous": true` — **`npm ci` then refuses to install it**, so local builds pass and CI/build machines fail from the same committed lockfile.

**Workaround:** declare `esbuild` in the consumer's own devDependencies, pinned inside your declared range (we used `^0.27.0`).

**Suggested fix:** add `esbuild` to `@opennextjs/cloudflare`'s `dependencies` (or re-export/bundle it), so the build-time import is satisfied by contract rather than by hoisting luck.

uttkarsh-26:
Fixes #1339

**Problem**
`@opennextjs/cloudflare` imports `esbuild` at build time (`src/cli/build/bundle-server.ts`, `src/cli/build/open-next/*.ts`) but declared it only in `devDependencies`. Consumers only resolved it via hoisting from `@opennextjs/aws` (0.25.4), which breaks when the hoist slot is claimed (e.g. `@vitejs/plugin-react` optional peer) and leaves `npm ci` with `extraneous: true`.

**Approach**
- Move `esbuild: catalog:` from `devDependencies` to `dependencies` in `packages/cloudflare/package.json` (catalog already `^0.27.0`).
- Add regression test `build-deps.spec.ts` that asserts `dependencies.esbuild` is defined (TDD: RED before fix, GREEN after).
- Add changeset patch.

**Tests**
- RED: `build-deps.spec.ts` fails with `expected undefined to be defined` before fix (1 failed, 342 passed).
- GREEN: 343/343 passed, 35 files after fix.
- `pnpm -C packages/cloudflare lint:check` and `ts:check` pass (import sorted).

**Risk**
Low. No runtime behavior change; only ensures published package carries its build-time dependency.

---

<a href="https://app.devin.ai/review/opennextjs/opennextjs-cloudflare/pull/1353" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/esbuild-dependency-82f9a43d.md
+++ b/.changeset/esbuild-dependency-82f9a43d.md
@@ -1,0 +1,7 @@
+---
+"@opennextjs/cloudflare": patch
+---
+
+fix: declare esbuild as runtime dependency
+
+esbuild is imported at build time by the Cloudflare adapter but was only in devDependencies, so consumers relied on hoisting from @opennextjs/aws. Add it to dependencies so builds work under npm ci and with hoist conflicts.
--- a/packages/cloudflare/package.json
+++ b/packages/cloudflare/package.json
@@ -59,6 +59,7 @@
 		"cloudflare": "^4.4.1",
 		"comment-json": "^4.5.1",
 		"enquirer": "^2.4.1",
+		"esbuild": "catalog:",
 		"glob": "catalog:",
 		"ts-tqdm": "^0.8.6",
 		"yargs": "catalog:"
@@ -73,7 +74,6 @@
 		"@types/rclone.js": "^0.6.1",
 		"@types/yargs": "catalog:",
 		"diff": "^8.0.2",
-		"esbuild": "catalog:",
 		"eslint": "catalog:",
 		"eslint-plugin-import": "catalog:",
 		"eslint-plugin-simple-import-sort": "catalog:",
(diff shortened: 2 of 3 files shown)


uttkarsh-26:
This PR fixes a real install-time failure: esbuild is imported at build time by the Cloudflare adapter but was declared only in devDependencies, so consumers relied on it being hoisted by @opennextjs/aws and builds break under npm ci or when another package claims the hoist. The change moves esbuild to runtime dependencies and includes a changeset plus a regression test. A maintainer review would be appreciated.

PR #1359, opened 2026-08-25
chore: bump Next.js to 15.5.24 and 16.3.3

### LINKED ISSUE #1342: [BUG] wasm loader patch silently fails on next@16.3.0 (peer range says it's supported) → every Prisma query throws WebAssembly.compileStreaming is not a function
### Describe the bug

@opennextjs/cloudflare: 1.20.2
next: 16.3.0 (16.2.11 works)

**Summary**

`@opennextjs/cloudflare 1.20.2` patches Turbopack's wasm loader so it works in workerd. The patch looks for a `function_declaration` named `loadWebAssemblyModule` (`replaceLoadWebAssemblyModuleRule` in `dist/cli/build/patches/plugins/turbopack.js`). **Next 16.3.0 renamed/changed that loader to a `compileModule` helper that calls `WebAssembly.compileStreaming`**, so the AST patch silently no-ops — the bundle keeps the `compileStreaming` code path, which workerd does not implement. Every Prisma query then throws `TypeError: WebAssembly.compileStreaming is not a function` (surfaces as failed signup/login, "An error occurred while registering", etc.). The failure is silent at build time; nothing warns that the patch didn't match.

**Why this is a bug in this package**

- The peerDependencies range "next": ">=15.5.21 <16 || >=16.2.11" permits 16.3.0, so npm install happily resolves it — but the deployment is broken at runtime.
- The AST patch failing to match is a hard failure mode that should either match the new shape or fail the build loudly.

**Evidence**

Build with each version, then inspect the final bundle:

```
npx wrangler deploy --dry-run --outdir=/tmp/dry
grep -o 'compileStreaming' /tmp/dry/worker.js | wc -l
grep -o 'loadWasmChunk'    /tmp/dry/worker.js | wc -l
```
```
- next@16.3.0 → compileStreaming: 3, loadWasmChunk: 0 (broken)
- next@16.2.11 → compileStreaming: 0, loadWasmChunk: 6 (correct)
```

In the 16.3.0 bundle the loader is emitted as a `compileModule `function ending in `await WebAssembly.compileStreaming(response);` in 16.2.11 it's the patchable `function loadWebAssemblyModule(chunkPath, _edgeModule)` that the plugin rewrites to `loadWasmChunk(chunkPath)`.

**Suggested fixes**

1. Update the AST patch to also match the Next 16.3.x `compileModule` wasm loader shape, or
2. Constrain `peerDependencies` to versions the patch actually supports (e.g. `<16.3.0`), and/or
3. When the patch matches zero `loadWebAssemblyModule/wasm functions`, log a hard warning — the silent fallthrough to `WebAssembly.compileStreaming` is the trap here.

### Steps to reproduce

1. Create a Next.js 16.3.0 App Router app (Turbopack) and install `@opennextjs/cloudflare@1.20.2` + `wrangler`:
`npx create-next-app@16.3.0 repro --app --turbopack`
`npm i @opennextjs/cloudflare@1.20.2`
(peerDependencies allows this: `next >=16.2.11`.)

2. Add a minimal Prisma query to the app (e.g. `new PrismaClient({ adapter })` with `@prisma/adapter-pg`, generated with `runtime = "cloudflare"`).

3. Build and dry-run the bundle:
```
npx opennextjs-cloudflare build
npx wrangler deploy --dry-run --outdir=/tmp/dry
```

4. Check the emitted worker:

```
grep -o 'compileStreaming' /tmp/dry/worker.js | wc -l   # 16.3.0 → 3
grep -o 'loadWasmChunk'    /tmp/dry/worker.js | wc -l   # 16.3.0 → 0
```

5. Deploy and run any Prisma query (e.g. call a server action that reads/writes a row).

**Observed:** 

On `next@16.3.0` the bundle retains Turbopack's `compileModule` loader that calls `WebAssembly.compileStreaming(response)` — the r`eplaceLoadWebAssemblyModuleRule` AST patch (matches `function_declaration` named `loadWebAssemblyModule`) matched nothing. In workerd, WebAssembly.compileStreaming is not a function, so every Prisma query throws `TypeError: WebAssembly.compileStreaming` is not a function (signup/login fails with "An error occurred while registering"). No warning at build time.

For comparison, the same app on `next@16.2.11 `produces `compileStreaming: 0`, `loadWasmChunk: 6` and works.

### Expected behavior

- The wasm loader patch should work on all `next` versions the `peerDependencies` range claims to support (`>=16.2.11`), including 16.3.x — either by matching the new `compileModule` shape, or
- The `peerDependencies` range should exclude versions the patch doesn't cover (e.g. `<16.3.0`), and
- The build should fail loudly (non-zero exit / hard warning) when the patch matches no wasm loader functions, instead of silently shipping a worker that throws `WebAssembly.compileStreaming is not a function` at runtime.

1.20.2

### Wrangler version

4.114.0

### next info output

```shell
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.5.0: Tue Jun  9 22:26:46 PDT 2026; root:xnu-12377.121.10~1/RELEASE_ARM64_T8103
  Available memory (MB): 8192
  Available CPU cores: 8
Binaries:
  Node: 24.18.0
  npm: 11.16.0
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  next: 16.2.11 // There is a newer version (16.3.1) available, upgrade recommended!
  eslint-config-next: N/A
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 6.0.3
Next.js Config:
  output: N/A
```

### Additional context

Also useful: the repro repo at https://github.com/joelrb/uploadthing-opennext-repro

**Environment**
- next 16.3.0, App Router, Turbopack
- @opennextjs/cloudflare 1.20.2
- wrangler 4.114.0
- Cloudflare Workers (workerd), nodejs_compat

### LINKED ISSUE #1351: [BUG] Turbopack WASM patcher misses Next 16.3's new emission — `WebAssembly.compileStreaming is not a function` at runtime (breaks Prisma on workerd)
## Summary

The Turbopack WASM patcher added in 1.19.6 (`patchTurbopackRuntime` / `loadWasmChunkFn` in `cli/build/patches/plugins/turbopack.js`) rewrites the **named** `loadWebAssembly` / `loadWebAssemblyModule` function declarations in `[turbopack]_runtime.js`. Next **16.3** no longer emits those: WASM loading now lives in **anonymous per-chunk helpers** registered as `compileModule`, which stream the `.wasm` from disk and call `WebAssembly.compileStreaming()`. The patcher never matches, the call survives into the deployed worker, and every WASM-backed import (e.g. Prisma's query compiler with `runtime = "workerd"`) throws at runtime:

```
TypeError: WebAssembly.compileStreaming is not a function
```

Because workerd forbids runtime WASM compilation entirely, there is no polyfill escape hatch — the chunk must be rewritten to a static import so wrangler uploads it as a CompiledWasm module (exactly what `loadWasmChunkFn` already does for the old emission).

This is nasty in production because pages that don't touch WASM still render fine; only the WASM-dependent paths 500. In our case it silently broke all Prisma/D1 queries — including login — while the site otherwise looked healthy.

## Environment

- `@opennextjs/cloudflare`: 1.20.2 (also reproduced on 1.20.1)
- `next`: 16.3.1 (Turbopack build; 16.2.9 output was patched correctly)
- `prisma` / `@prisma/client`: 7.9.1, generator `runtime = "workerd"`, `@prisma/adapter-d1`
- `wrangler`: 4.124.0
- Reproduces in `opennextjs-cloudflare preview` (local workerd) and deployed Workers

## What Next 16.3 emits

In the built server chunks (minified, formatted here), each WASM chunk gets a helper like:

```js
// registered as ["compileModule", 0, fn] in the Turbopack module system
async function compileModule(chunkPath) {
  const response = (function (p) {
    const { createReadStream } = require("fs");
    const { Readable } = require("stream");
    const stream = createReadStream(require("path").resolve(root, p));
    return new Response(Readable.toWeb(stream), {
      headers: { "content-type": "application/wasm" },
    });
  })(chunkPath);
  return await WebAssembly.compileStreaming(response);
}
```

alongside a path-registration module:

```js
moduleId, function (ctx) {
  ctx.q("server/chunks/ssr/<chunk-name>_bg.wasm");
}
```

There is no `function loadWebAssemblyModule(...)` / `function loadWebAssembly(...)` declaration in `[turbopack]_runtime.js` for the ast-grep rules (`replaceLoadWebAssemblyModuleRule` / `replaceLoadWebAssemblyRule`) to match, so `patchCode` is a no-op for these sites. The `.wasm` files themselves are still traced and uploaded — they're just never reachable via a static import, so the runtime falls back to the fs + `compileStreaming` path.

## Repro

1. Next 16.3.x app with any WASM-importing dependency — easiest is Prisma with the workerd client (`generator client { provider = "prisma-client"; runtime = "workerd" }`) and `@prisma/adapter-d1`.
2. `next build` (Turbopack) + `opennextjs-cloudflare build`.
3. `opennextjs-cloudflare preview`, hit any route that runs a Prisma query.
4. 500 with `TypeError: WebAssembly.compileStreaming is not a function`.

Building the same app with `next@16.2.x` produces the old emission and works (patcher matches).

## Workaround we're shipping

A post-build script (run after `opennextjs-cloudflare build`) that finds the emitted helpers and rewrites them to delegate to a generated switch of static literal `import("<chunk>.wasm")` calls (same idea as `loadWasmChunkFn`, keyed by the `.next`-relative chunk path the helpers receive). With this in place, Prisma-on-D1 works again under 16.3 — verified with a full login flow in local workerd and in production:

```js
/**
 * Post-build patch: make Turbopack's WASM loading work on Cloudflare Workers.
 *
 * Next 16.3's Turbopack emits per-chunk `compileModule` helpers that stream a
 * .wasm file from disk and call `WebAssembly.compileStreaming()`. workerd
 * forbids runtime WASM compilation ("Wasm code generation disallowed by
 * embedder"), so every Prisma query in production throws
 * "WebAssembly.compileStreaming is not a function".
 *
 * @opennextjs/cloudflare ships a patcher for this class of problem
 * (loadWasmChunkFn, since 1.19.6), but it only rewrites the NAMED
 * `loadWebAssembly{,Module}` functions that Next <=16.2 emitted inside
 * `[turbopack]_runtime.js`. Next 16.3 moved the logic into anonymous helpers
 * inside the compiled chunks, which the upstream patcher never matches.
 *
 * This script rewrites those helpers to resolve the chunk through a generated
 * switch of static `import("<path>.wasm")` calls — string literals, so
 * wrangler discovers each .wasm at bundle time and uploads it as a
 * CompiledWasm module (the only WASM mechanism workerd allows). Remove once
 * opennextjs-cloudflare supports Next 16.3's emission.
 *
 * Run after `opennextjs-cloudflare build` (see package.json "build").
 */
import fs from "node:fs";
import path from "node:path";

const root = path.resolve(path.dirname(new URL(import.meta.url).pathname), "..");
const serverDir = path.join(root, ".open-next", "server-functions", "default");
const handlerPath = path.join(serverDir, "handler.mjs");

if (!fs.existsSync(handlerPath)) {
  console.error("patch-turbopack-wasm: handler.mjs not found — did the build run?");
  process.exit(1);
}

/** Every traced .wasm chunk, keyed by its `.next`-relative runtime path. */
function collectWasmChunks() {
  const chunks = [];
  const base = path.join(serverDir, ".next", "server", "chunks");
  const walk = (dir) => {
    if (!fs.existsSync(dir)) return;
    for (const entry of fs.readdirSync(dir, { withFileTypes: true })) {
      const p = path.join(dir, entry.name);
      if (entry.isDirectory()) walk(p);
      else if (entry.name.endsWith(".wasm")) {
        chunks.push({
          runtimePath: path.relative(path.join(serverDir, ".next"), p),
          importPath: `./${path.relative(serverDir, p)}`,
        });
      }
    }
  };
  walk(base);
  return chunks;
}

let code = fs.readFileSync(handlerPath, "utf8");
const marker = "WebAssembly.compileStreaming(";
let patchedCount = 0;
let searchFrom = 0;

while (true) {
  const at = code.indexOf(marker, searchFrom);
  if (at === -1) break;

  // Anchor to the enclosing `async function NAME(PARAM){` emitted directly
  // before the helper body; the whole helper is a few hundred bytes.
  const windowStart = Math.max(0, at - 600);
  const window = code.slice(windowStart, at);
  const fnMatch = [...window.matchAll(/async function (\w+)\((\w+)\)\{/g)].pop();
  if (!fnMatch) {
    searchFrom = at + marker.length;
    continue;
  }
  const fnStart = windowStart + fnMatch.index;
  const [, fnName, fnParam] = fnMatch;

  // The helper ends at the `)}` closing the compileStreaming call.
  const end = code.indexOf(")}", at);
  if (end === -1) break;

  const replacement = `async function ${fnName}(${fnParam}){return await globalThis.__cfLoadWasmChunk(${fnParam})}`;
  code = code.slice(0, fnStart) + replacement + code.slice(end + 2);
  patchedCount += 1;
  searchFrom = fnStart + replacement.length;
}

if (patchedCount === 0) {
  console.log("patch-turbopack-wasm: no compileStreaming helpers found — nothing to patch.");
  process.exit(0);
}

const chunks = collectWasmChunks();
const cases = chunks
  .map(
    (c) =>
      `    case ${JSON.stringify(c.runtimePath)}: return (await import(${JSON.stringify(c.importPath)})).default;`,
  )
  .join("\n");

code += `
// Injected by scripts/patch-turbopack-wasm.mjs — see that file for why.
globalThis.__cfLoadWasmChunk = async function (chunkPath) {
  switch (chunkPath) {
${cases}
    default:
      throw new Error("Unknown wasm chunk: " + chunkPath);
  }
};
`;

fs.writeFileSync(handlerPath, code);
console.log(
  `patch-turbopack-wasm: rewrote ${patchedCount} compileStreaming helper(s); ${chunks.length} wasm chunk(s) importable:`,
);
for (const c of chunks) console.log(`  ${c.runtimePath}`);
```

## Suggested fix

Extend the Turbopack patch to also match 16.3's emission — e.g. an ast-grep rule targeting `await WebAssembly.compileStreaming($X)` inside the compiled chunks/handler (not just `[turbopack]_runtime.js`), rewriting those call sites to route through the generated `loadWasmChunk` switch. A guard in the build that fails loudly when `WebAssembly.compileStreaming` survives into the final bundle would also have surfaced this at build time instead of in production.

### LINKED ISSUE #1362: Node proxy bundle retains unsupported dynamic instrumentation require
After the Node.js middleware/proxy support from #1309, an OpenNext build for a Next.js 16.3.3 app with no instrumentation hook still leaves getInstrumentationModule in .open-next/middleware/handler.mjs. That function calls the bundle helper __require with a runtime-computed path to server/instrumentation.js.

The build succeeds, but a deployed Worker fails every proxy request in Workerd with:

    Dynamic require of "/bundle/.next/server/instrumentation.js" is not supported

Environment:
- @opennextjs/cloudflare package built from commit 070e6ec34047e7dd420ef6acc639c1d836a86db6 (reports 1.20.3)
- Next.js 16.3.3
- Node.js middleware/proxy enabled
- no instrumentation.js or instrumentation.ts in the application

Expected: when no instrumentation hook exists, the Node proxy bundle should statically replace or eliminate the optional instrumentation loader, matching the instrumentation patch already applied to other OpenNext bundles.

A downstream deterministic repair that replaces only this audited loader with a null-returning implementation restores local Workerd and deployed request handling, but it should be removable once the generated proxy artifact is fixed upstream.

vicb:
fixes #1362
fixes #1342 
fixes #1351

---

<a href="https://app.devin.ai/review/opennextjs/opennextjs-cloudflare/pull/1359" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


conico974 (approved):

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
```diff
@@ -299,6 +395,14 @@ export const patchTurbopackRuntime: CodePatcher = {
 			patchCode: async ({ code, tracedFiles, filePath }) =>
 				patchTurbopackRuntimeCode({ code, filePath, tracedFiles }),
 		},
+		{
+			versions: ">=15.0.0",
```
conico974: Why 15 ? Shouldn't it be 16.2 according to the comment

conico974 (approved):
Just a little nit/question

packages/cloudflare/src/cli/build/patches/plugins/turbopack.ts
vicb:
```suggestion
			// Next.js 16.3.0 moved the wasm loaders out of the Turbopack runtime and started
			// emitting them on demand in the chunks that need them. Earlier versions are handled
			// by the runtime patch above.
			versions: ">=16.3.0",
```

vicb: Good catch!
Thanks!

PR #1383, opened 2026-09-09
feat: support multiple zones during automatic cache purge

### LINKED ISSUE #1347: [FEATURE] Support multiple zones during automatic cache purge

We currently run a single Next.js application behind multiple domains, and are looking at migrating to Cloudflare workers.

Each domain is a separate zone in Cloudflare. Automatic cache purging currently only supports a single Cloudflare zone via environment variables, so it seems we cannot configure one worker to automatically purge multiple domains.

An ability to configure multiple zones for automatic cache purge.

Separate CF workers deployments for each zone. This increases management complexity and results in duplicated DO caches (several of the domains use the same underlying data)

Latest version

### Additional context

_No response_

### Before submitting

- [x] I have checked that there isn't already a similar feature request
- [x] This is a single feature (not multiple features in one request)

harry-gocity:
Closes https://github.com/opennextjs/opennextjs-cloudflare/issues/1347.

Automatic cache purge only supports a single Cloudflare zone through `CACHE_PURGE_ZONE_ID`. When one worker serves multiple domains (each a separate Cloudflare zone), the cache is only purged for one zone. The other zones serve stale content.

This PR adds support so that `CACHE_PURGE_ZONE_ID` can accept a comma-separated list of zone IDs.

Purge requests run in parallel across all configured zones. The `CACHE_PURGE_API_TOKEN` must have the Cache Purge permission on every configured zone.

When using the durable object path (`NEXT_CACHE_DO_PURGE`), rate-limited zones are tracked per-retry so that zones which already succeeded are not purged again.

Existing single-zone setups that only set `CACHE_PURGE_ZONE_ID` continue to work without changes.

### DIFF
--- a/packages/cloudflare/src/api/overrides/internal.ts
+++ b/packages/cloudflare/src/api/overrides/internal.ts
@@ -55,31 +55,65 @@ export async function purgeCacheByTags(tags: string[]) {
 	}
 }
 
-export async function internalPurgeCacheByTags(env: CloudflareEnv, tags: string[]) {
-	if (!env.CACHE_PURGE_ZONE_ID || !env.CACHE_PURGE_API_TOKEN) {
+export type PurgeCacheResult =
+	| "missing-credentials"
+	| "rate-limit-exceeded"
+	| "purge-failed"
+	| "purge-success";
+
+export async function internalPurgeCacheByTags(
+	env: CloudflareEnv,
+	tags: string[]
+): Promise<PurgeCacheResult> {
+	const zoneIds = new Set<string>();
+
+	if (env.CACHE_PURGE_ZONE_ID) {
+		zoneIds.add(env.CACHE_PURGE_ZONE_ID);
+	}
+
+	if (env.CACHE_PURGE_ZONE_IDS) {
+		for (const raw of env.CACHE_PURGE_ZONE_IDS.split(",")) {
+			const id = raw.trim();
+			if (id) zoneIds.add(id);
+		}
+	}
+
+	if (zoneIds.size === 0 || !env.CACHE_PURGE_API_TOKEN) {
 		// THIS IS A NO-OP
-		error("No cache zone ID or API token provided. Skipping cache purge.");
+		error("No cache zone ID(s) or API token provided. Skipping cache purge.");
 		return "missing-credentials";
 	}
 
+	const results = await Promise.all(zoneIds.values().map((zoneId) => purgeZone(env, zoneId, tags)));
+
+	// If any zone hit the rate limit, report that so the caller can retry.
+	if (results.includes("rate-limit-exceeded")) {
+		return "rate-limit-exceeded";
+	}
+
+	if (results.includes("purge-failed")) {
+		return "purge-failed";
+	}
+
+	return "purge-success";
+}
+
+async function purgeZone(env: CloudflareEnv, zoneId: string, tags: string[]): Promise<PurgeCacheResult> {
 	let response: Response | undefined;
 	try {
-		response = await fetch(
-			`https://api.cloudflare.com/client/v4/zones/${env.CACHE_PURGE_ZONE_ID}/purge_cache`,
-			{
-				headers: {
-					Authorization: `Bearer ${env.CACHE_PURGE_API_TOKEN}`,
-					"Content-Type": "application/json",
-				},
-				method: "POST",
-				body: JSON.stringify({
-					tags,
-				}),
-			}
-		);
+		response = await fetch(`https://api.cloudflare.com/client/v4/zones/${zoneId}/purge_cache`, {
+			headers: {
+				Authorization: `Bearer ${env.CACHE_PURGE_API_TOKEN}`,
+				"Content-Type": "application/json",
+			},
+			method: "POST",
+			body: JSON.stringify({
+				tags,
+			}),
+		});
 		if (response.status === 429) {
 			// Rate limit exceeded
-			error("purgeCacheByTags: Rate limit exceeded. Skipping cache purge.");
+			error(`purgeCacheByTags: Rate limit exceeded for zone ${zoneId}. Skipping cache purge.`);
 			return "rate-limit-exceeded";
 		}
 		const bodyResponse = (await response.json()) as {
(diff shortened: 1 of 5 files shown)


packages/cloudflare/src/api/overrides/internal.ts
```diff
@@ -55,31 +55,65 @@ export async function purgeCacheByTags(tags: string[]) {
...
+	if (env.CACHE_PURGE_ZONE_ID) {
+		zoneIds.add(env.CACHE_PURGE_ZONE_ID);
+	}
+
+	if (env.CACHE_PURGE_ZONE_IDS) {
+		for (const raw of env.CACHE_PURGE_ZONE_IDS.split(",")) {
+			const id = raw.trim();
+			if (id) zoneIds.add(id);
+		}
+	}
```
vicb, line 73: What about keeping only `CACHE_PURGE_ZONE_ID` but allow coma separated values there?

vicb (changes requested):

packages/cloudflare/src/api/overrides/internal.ts
harry-gocity: Happy to do that, I guess it's not technically a breaking change 🤔 .

harry-gocity: Removed `CACHE_PURGE_ZONE_IDS` and updated changesets. cheers!

harry-gocity: Hey @vicb. Do you need any other changes on this PR?

packages/cloudflare/src/api/overrides/internal.ts
```diff
@@ -88,15 +116,15 @@ export async function internalPurgeCacheByTags(env: CloudflareEnv, tags: string[
...
 				bodyResponse.errors.map((error) => `${error.code}: ${error.message}`)
 			);
 			return "purge-failed";
 		}
-		debugCache("purgeCacheByTags", "Cache purged successfully for tags:", tags);
+		debugCache("purgeCacheByTags", `Cache purged successfully for zone ${zoneId}, tags:`, tags);
 		return "purge-success";
 	} catch (error) {
-		console.error("Error purging cache by tags:", error);
+		console.error(`Error purging cache by tags for zone ${zoneId}:`, error);
```
vicb: While you are updating this file, can you change all `console.error` to `error` ?
Thanks

vicb (changes requested):
Sorry I missed your update.

The changes look good.

Could you please address the comment? (the error was here before the PR but would be nice to fix anyway).

Thanks!

packages/cloudflare/src/api/overrides/internal.ts
harry-gocity: done!

vicb:
Devin has a sensible comment:

packages/cloudflare/src/api/overrides/internal.ts:R81-85

**Successful zones repeat on retries**

When one zone returns 429, `internalPurgeCacheByTags` discards which zones succeeded. [BucketCachePurge.alarm](packages/cloudflare/src/api/durable-objects/bucket-cache-purge.ts:59-69) retries the entire batch, so successful zones receive duplicate purges and can prolong rate limiting.

The durable object stores one batch of tags, not progress for each zone. A rate-limited aggregate result makes the alarm throw before deleting that batch, so the built-in alarm retry invokes the same zone list again. Successes from the first attempt are therefore lost. Repeated requests consume purge quota and can exhaust the alarm's retry attempts while the original limited zone remains unpurged.

**Example:** With zones A and B, A succeeds while B returns 429. The alarm retries both zones instead of only B. If A then returns 429, the batch remains pending even if B succeeds.

**Recommended fix:** Preserve per-zone outcomes across durable-object retries. Store pending zone/tag work separately, or return per-zone results and persist only the unsuccessful zone/tag pairs before throwing for rate-limited work.

harry-gocity:
@vicb fair enough on the retries, I can take a look at this. Would you like it on this PR or stacked separately?

vicb:
An additive commit is good enough but feel free to use stacked PRs if you prefer

harry-gocity:
Pushed some changes here to retry-rate limited zones across alarms.

`internalPurgeCacheByTags()` now takes `zoneIds` rather than parsing env and returns `{ status, rateLimitedZones }` instead of just a status string. `rateLimitedZones` contains the zone IDs that returned 429.

A new `pending_zones` table tracks rate-limited zones. On rate-limit, the alarm stores only the rate-limited zone IDs in `pending_zones`, then throws to trigger the existing retry mechanism. On retry, the alarm reads `pending_zones` and passes only those zones to `internalPurgeCacheByTags`, so zones that already succeeded are not purged again.

Added some test cases too, had to make quite a few changes to mocks in `bucket-cache-purge.spec.ts` since `alarm()` now makes multiple `sql.exec` calls in a single invocation, across both tables. Tried to capture the state transitions with `toHaveBeenNthCalledWith` but it does feel like the suite would benefit from some tests with a stronger integration and real row state.

harry-gocity:
There's a gap in my current approach, and I'm about to go on holiday until next week. 

- First alarm fires, and tags A and B are in `cache_purge`. Zones X and Y are configured, zone X succeeds, zone Y returns 429. The DO stores Y in `pending_zones` and the alarm throws.
- Before the retry fires, a new request adds tag C to `cache_purge`.
- Retry fires, and tags A, B and C are read from `cache_purge`. Pending_zones contains Y, so only zone Y is purged.
Tags A, B and C are deleted from `cache_purge`. Tag C was never purged against zone X.

Will take a look when I'm back, open to any suggestions on preferred implementation approaches.

packages/cloudflare/src/api/durable-objects/bucket-cache-purge.workerd.spec.ts
```diff
@@ -0,0 +1,342 @@
...
+
+			// zone-x pairs are removed, zone-y pairs remain.
+			const pending = await getPendingPurges(stub);
+			expect(pending).toEqual([{ tag: "tag1", zone_id: "zone-y" }]);
+
+			// Tags stay in cache_purge (retry not complete).
+			expect(await getTags(stub)).toEqual(["tag1"]);
+		});
+
+		it("should not re-purge tags added during retry against already-succeeded zones", async () => {
```
harry-gocity, line 225: Covers https://github.com/opennextjs/opennextjs-cloudflare/pull/1383#issuecomment-5665686226.

harry-gocity:
Hey @vicb I reverted that initial approach and took a new direction to avoid the issues with new tags only being purged against retried zones if there was previous rate limiting (https://github.com/opennextjs/opennextjs-cloudflare/pull/1383#issuecomment-5665686226)

Along the way I found testing was quite hard to update and reason with in general, the mocked `sql.exec` chains relied heavily on call order and internal query string assertions.

I found Cloudflare has a Vitest plugin which I've used instead in tests for `bucket-cache-purge.spec.ts`. There's a separate Vitest project targeting `.workerd.spec.ts` suites, which runs the tests inside workerd with real SQLite and more accurate alarm scheduling and lifecycle events. This removes a ton of jest-based mocking and means the test query the tag tables to assert, so things are less brittle. 

Some parts e.g. `internalPurgeCacheByTags` are still mocked because they call the Cloudflare API and depends on modules unavailable in workerd. I didn't want to go to far and undo a lot of the fetch mocks too.

I've not used this plugin before so I used the [docs](https://developers.cloudflare.com/durable-objects/examples/testing-with-durable-objects) and also examples from https://github.com/cloudflare/workers-sdk/tree/main/fixtures/vitest-plugin-examples. There's also a new tsconfig for test files only so that `cloudflare:test` and `cldouflare:workers` types are available without polluting the source.

I can't create stacked PRs for this because GH doesn't support stacked PRs when branches aren't all in the same repo, and this is a PR from my fork. So I've kept the commits isolated instead.

vicb:
Thanks for working on this.

The current PR adds a lot of code to maintain for a narrow use case.

I'll spend some time in the coming days to think about ways to keep the simple behavior we have in this repo and at the same time allow users to extend the behavior (i.e. purge multiple zones).

I think it would be a reasonable tradeoff

